PR #9, opened 2018-08-22
[WIP] plaintext benchmark

jangko:
- [ ] finishing bench bot implementation
- [ ] add participant: rust actix
- [ ] add participant: go-lang fasthttp
- [ ] add participant: c++ ulib
- [ ] polishing benchmark report

### DIFF
--- a/tests/benchmark/asyncdispatch2/plaintext.dockerfile
+++ b/tests/benchmark/asyncdispatch2/plaintext.dockerfile
@@ -1,0 +1,6 @@
+FROM statusteam/nim-base
+RUN nimble install -y https://github.com/status-im/nim-asyncdispatch2
+WORKDIR /server
+COPY server.nim server.nim
+RUN nim c -d:release server.nim
+CMD ["./server"]
--- a/tests/benchmark/asyncdispatch2/server.nim
+++ b/tests/benchmark/asyncdispatch2/server.nim
@@ -1,0 +1,44 @@
+import strutils, posix
+
+import asyncdispatch2
+
+const DOUBLECRLF = @['\r'.byte, '\L'.byte, '\r'.byte, '\L'.byte]
+
+proc handleConnection(ss: StreamServer, st: StreamTransport) {.async.} =
+  var resp = "HTTP/1.1 200 OK\r\L" &
+             "Content-Length: 15\r\L\r\L" &
+             "Hello, World!\r\L"
+  var readBuff: array[256, byte]
+  while not st.atEof:
+    try:
+      let read = await st.readUntil(addr readBuff[0], len(readBuff), DOUBLECRLF)
+    except:
+      break
+    let w = await st.write(addr resp[0], resp.len)
+    assert w == resp.len
+  st.close()
+
+proc handleBreak(udata: pointer) =
+  var cdata = cast[ptr CompletionData](udata)
+  var svr = cast[StreamServer](cdata.udata)
+  echo "\nCTRL+C pressed, stopping server..."
+  svr.stop()
+  svr.close()
+
+proc serve(onAddress: string) =
+  let
+    ta = initTAddress(onAddress)
+    svr = createStreamServer(ta, handleConnection, {ReuseAddr}, backlog = 128)
+  when not defined(windows):
+    discard addSignal(SIGINT, handleBreak, udata = cast[pointer](svr))
+  svr.start()
+  echo "Server started at ", ta
+  try:
+    waitFor svr.join()
+  except:
+    echo "Error happens: ", getCurrentExceptionMsg()
+  finally:
+    echo "bye!"
+
+when isMainModule:
+  serve("0.0.0.0:8080")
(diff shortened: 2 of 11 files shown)


cheatfate:
I'm sorry @jangko, there no reason to make benchmarks on multi-threaded vs single-threaded apps. Could you please try to limit number of processes/threads used by chosen framework?

jangko:
> Could you please try to limit number of processes/threads used by chosen framework?

sure, i will try to make the benchmark as fair as possible for each participant

cheatfate:
From my tests on VM with just only 2 processors available, `mofuw` is not so performant, and also it produces more errors, not successful responses:
```

**Comment by cheatfate** [comment]
While on the same VM asyncdispatch2 benchmark produces such output:
```

jangko:
mofuw need `/plaintext` uri to avoid `Non-2xx or 3xx responses`

cheatfate:
@jangko, from what i see `asyncdispatch` and `asyncdispatch2` benchmarks - do not support `pipeline` messages. So why you are testing it?

jangko:
Most of performant techempower benchmark participant are designed to handle pipeline messages. On the other hand, this benchmark does not take it into account. 
While testing those frameworks, I realize their performance can vary significantly with/without pipeline mode.  I think it would be important to keep this information.
The final result of this benchmark will include both pipeline and no-pipeline mode for comparison, or it will become bench-bot switchable feature.
What we can do now is make them all run in single thread mode. Then we can decide what we will need to do with this pipeline.

cheatfate:
But you can adjust benchmark source to support pipeline for both `asyncdispatch` and `asyncdispatch2`.

jangko:
> But you can adjust benchmark source to support pipeline for both asyncdispatch and asyncdispatch2.

agree

jangko:
ad2 is fast but then suffer massive slowdown, hmm. interesting

.travis.yml
```diff
@@ -37,3 +37,5 @@ install:
 script:
   - nimble install -y
   - nimble test
+  - nimble benchmark
```
arnetheduck: what's the purpose of running benchmark on CI setup? neither travis nor appveyor provide stable hardware so numbers will be meaningless, and benchmarks tend to take a while so it will slow down every PR / build roundtrip..

jangko:
> and benchmarks tend to take a while so it will slow down every PR / build roundtrip..

That's right, it took significant amount of time. I already removed it from CI.

---

## Summary
* __mofuw__, mofuw use asyncdispatch, expected performance should not more than asynchdispatch itself.
* __asyncdispatch__, although it is slower than asyncdispatch2, it can handle high concurrency quite well.
* __asyncdispatch2__, at high concurrency it has tendency become slower significantly,
  but surpringsingly it is the only framework in this test that can handle non pipeline request faster than other  frameworks although using almost identical code with asyncdispatch when handle request/response.
* __actix-raw__, very fast when multi threaded, not so when single threaded.
* __fasthttp__, very fast when multi threaded, not so when single threaded.
* __libreactor__, still very fast although in single thread mode.
## Conclusion
* asyncdispatch2 could be a good candidate to replace asycndispatch
* it still has room for improvement especially when handle high count connections.

Sorry I cannot work faster because of some circumstances, but I think this one is ready for review.

dm1try:
looks like `asyncdispatch2` benchmark has broken implementation at least on Mac OS: it generates ~10x responses for the same request([provided results](https://github.com/status-im/nim-asyncdispatch2/pull/9/files#diff-2d7cf66f8863e56b2cd962da2556eb1eR23) have a similar correlation)

zah:
@jangko, I've pushed to your branch a commit adding a command-line option for deciding whether threads should be used. To support it, the test programs need a minor modification - they must check whether the environment variable `USE_THREADS` is set. You can see an example here:

https://github.com/status-im/nim-asyncdispatch2/pull/9/commits/4fa3b6e3c7096cc72c33264c4174abe8334c064b#diff-a700604e55a2b00d28959045bdda5b09R26

The asyncdispatch2 program that you've prepared is violating the rules of the competition, which are given here:
https://www.techempower.com/benchmarks/#section=code

In particular, this rule:

> This test is not intended to exercise the allocation of memory or instantiation of objects. Therefore it is acceptable but not required to re-use a single buffer for the response text (Hello, World). However, the response must be fully composed from the response text and response headers within the scope of each request and it is not acceptable to store the entire payload of the response, or an unnaturally large subset of the response, headers inclusive, as a pre-rendered buffer. "Buffer" here refers to a byte array, byte buffer, character array, character buffer, string, or string-like data structure. The spirit of the test is to require the construction of the HTTP response as is typically done by a framework or platform via concatenation of strings or similar. For example, pre-rendering a buffer with HTTP/1.1 200 OK<cr>Content-length: 15<cr>Server: Example<cr> would not be acceptable.

So, you must break up a bit the strings being written as a response. I think you can avoid some of the allocations and concatenations as well, @cheatfate may provide some hints for what is the most efficient way to build the response piece by piece.

PR #13, opened 2018-11-29
Clear ReadError from transp.state when setting transp.error to nil

kdeme:
This PR solves a segfault I've encountered when working with nim-eth-p2p.

The segfault will occur in following setting:
1. Error occurs on `recvfrom` in `readDatagramLoop` and `setReadError` is ran.
2. A `raise transp.getError()` is done in `getMessage` or `peekMessage` in e.g. the transport `function`.
3. On next call of this `getMessage`, it will still have the `ReadError` in the state and again try to get the error, which is nil now.

While I've encountered the issue with use of `DatagramTransport`, I think a similar issue can occur for `StreamTransport`, so I changed also the `getError()` there.

I'm guessing I can remove `getError()` but I've not done it as I see this gets exported?
The two `setReadError` templates could also be moved to common.nim but perhaps not done for above reason? (not to export them?)

~~One thing that still has me puzzled is the actual error I am getting on `recvMsg`.
It is `EAGAIN` error, but the `recvfrom` call is set to blocking and I believe there is no receive timeout set (at least it reads out as 0 when I tested).~~ 
Woops, the `setSocketBlocking` proc naming had me confused, makes sense that is not blocking... and thus EAGAIN sounds normal (and should perhaps be handled even differently?).

### DIFF
--- a/asyncdispatch2/transports/common.nim
+++ b/asyncdispatch2/transports/common.nim
@@ -469,6 +469,12 @@ template getError*(t: untyped): ref Exception =
   (t).error = nil
   err
 
+template getReadError*(t: untyped): ref Exception =
+  var err = (t).error
+  (t).error = nil
+  (t).state.excl(ReadError)
+  err
+
 template getTransportOsError*(err: OSErrorCode): ref TransportOsError =
   var msg = "(" & $int(err) & ") " & osErrorMsg(err)
   var tre = newException(TransportOsError, msg)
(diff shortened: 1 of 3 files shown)


cheatfate:
`EAGAIN` is bad error, it means that you trying to read message while system queue is empty. This must not be happens in normal condition. You must read message only when its available. So if you encountered this error, then it looks like you tried to read message twice or more times, while only one message can be read.

cheatfate:
Also it looks like you using `DatagramTransport` in wrong manner, because if you get exception on `peekMessage` or `readMessage` you just need to return from your handler.

kdeme:
Sorry, perhaps I did not explain properly.
The Step 3 is again triggered from the handler.

The error occurs here: https://github.com/status-im/nim-eth-p2p/blob/master/eth_p2p/discovery.nim#L251

That is the correct way of using it right?

But looking more in detail in the code (and in the epoll / selector code as I am running Linux), I think I understand what you mean. I don't get it why the `EAGAIN` occurs.

cheatfate:
Could you please explain bug in detail and fill an issue about it, because this PR is wrong place for discussing bug.

PR #22, opened 2019-03-12
all() fix with old behavior and tests for it.

cheatfate:


chronos/asyncfutures2.nim
```diff
@@ -413,64 +415,68 @@ proc `or`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
...
-  when T is void:
-    var
-      retFuture = newFuture[void]("asyncdispatch.all")
-      completedFutures = 0
+  ## If the awaited futures *are* ``Future[void]``, this proc returns
+  ## ``Future[void]``.
+  ##
+  ## Note, that if one of the futures in ``futs`` will fail, result of ``all()``
+  ## will also be failed with error from failed future.
+  let totalFutures = len(futs)
```
yglukhov: `assert(totalFutures != 0)`?

stefantalpalaru: Please use doAssert() instead of assert().

chronos/asyncfutures2.nim
```diff
@@ -413,64 +415,76 @@ proc `or`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
   fut2.callback = cb
   return retFuture
 
-# ZAH: The return type here could be a tuple
-# This will enable waiting a heterogenous collection of futures.
+# proc all*[T](futs: varargs[Future[T]]): Future[void] =
```
arnetheduck: generally, git keeps a copy of the old code already..

cheatfate: Commented version is not for git, its for me and my memory. Because when in month i will not remember that there was also nice version, even if Git keep track on it.

cheatfate:
Something broken here i will create a new one.

PR #24, opened 2019-03-21
Introduce monotonic timer functions.

cheatfate:
Old fast timers are available through compiler switch.
Add tests for both timers.

### DIFF
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -9,26 +9,45 @@ skipDirs      = @["tests"]
 
 requires "nim > 0.18.0"
 
-task test, "Run all tests":
-  for tfile in @[
-      "testsync",
-      "testsoon",
-      "testtime",
-      "testfut",
-      "testsignal",
-      "testaddress",
-      "testdatagram",
-      "teststream",
-      "testserver",
-      "testbugs",
-    ]:
-    for cmd in @[
-        "nim c -r -d:useSysAssert -d:useGcAssert tests/" & tfile,
-        "nim c -r tests/" & tfile,
-        #"nim c -r --gc:markAndSweep tests/" & tfile,
-        "nim c -r -d:release tests/" & tfile
-      ]:
-      echo "\n" & cmd
-      exec cmd
-      rmFile("tests/" & tfile.toExe())
+import ospaths
 
+task test, "Run all tests":
+
+  var testFiles = @[
+    "testsync",
+    "testsoon",
+    "testtime",
+    "testfut",
+    "testsignal",
+    "testaddress",
+    "testdatagram",
+    "teststream",
+    "testserver",
+    "testbugs",
+  ]
+
+  var testCommands = @[
+    "nim c -r -d:useSysAssert -d:useGcAssert",
+    "nim c -r",
+    "nim c -r -d:release"
+  ]
+
+  var timerCommands = @[
+    " -d:asyncTimer=fast",
+    " -d:asyncTimer=mono"
+  ]
+
+  for tfile in testFiles:
+    if tfile == "testtime":
+      for cmd in testCommands:
+        for def in timerCommands:
+          var commandLine = (cmd & def & " tests") / tfile
+          echo "\n" & commandLine
+          exec commandLine
+          rmFile("tests" / tfile.toExe())
+    else:
+      for cmd in testCommands:
+        var commandLine = (cmd & " tests") / tfile
+        echo "\n" & commandLine
+        exec commandLine
+        rmFile("tests" / tfile.toExe())
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -1,5 +1,5 @@
 packageName   = "chronos"
-version       = "2.2.1"
+version       = "2.2.2"
 author        = "Status Research & Development GmbH"
 description   = "Chronos"
 license       = "Apache License 2.0 or MIT"
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -33,7 +33,7 @@ task test, "Run all tests":
   ]
 
   var timerCommands = @[
-    " -d:asyncTimer=fast",
+    " -d:asyncTimer=system",
     " -d:asyncTimer=mono"
   ]
 
--- a/chronos/timer.nim
+++ b/chronos/timer.nim
@@ -429,23 +429,6 @@ proc fromNow*(t: typedesc[Moment], a: Duration): Moment {.inline.} =
   ## Returns moment in time which is equal to current moment + Duration.
   result = Moment.now() + a
 
-func getAsyncTimestamp*(a: Duration): auto {.inline.} =
-  ## Return rounded up value of duration with milliseconds resolution.
-  ##
-  ## This procedure used internally in `asyncloop.nim`.
-  let res = a.value div 1_000_000'i64
-  let mid = a.value mod 1_000_000'i64
-  when defined(windows):
-    if res > int64(high(int32)):
-      raise newException(OverflowError, "Timestamp value is too big " & $res)
-    result = cast[DWORD](res)
-    result += DWORD(min(1'i32, cast[int32](mid)))
-  else:
-    if res > int64(high(int)):
-      raise newException(OverflowError, "Timestamp value is too big " & $res)
-    result = cast[int](res)
-    result += min(1, cast[int](mid))
-
 when defined(posix):
   from posix import Time, Suseconds, Timeval, Timespec
 
--- a/chronos/timer.nim
+++ b/chronos/timer.nim
@@ -28,14 +28,15 @@ when defined(windows):
   when asyncTimer == "system":
     from winlean import DWORD, getSystemTimeAsFileTime, FILETIME
 
-    proc fastEpochTime*(): uint64 {.inline.} =
+    proc fastEpochTime*(): uint64 {.
+         inline, deprecated: "Use Moment.now()".} =
       ## Timer resolution is millisecond.
       var t: FILETIME
       getSystemTimeAsFileTime(t)
       result = ((cast[uint64](t.dwHighDateTime) shl 32) or
                  cast[uint64](t.dwLowDateTime)) div 10_000
 
-    proc fastEpochTimeNano*(): uint64 {.inline.} =
+    proc fastEpochTimeNano(): uint64 {.inline.} =
       ## Timer resolution is nanosecond.
       var t: FILETIME
       getSystemTimeAsFileTime(t)
@@ -51,13 +52,14 @@ when defined(windows):
     var queryFrequencyM: uint64
     var queryFrequencyN: uint64
 
-    proc fastEpochTimeNano*(): uint64 {.inline.} =
+    proc fastEpochTimeNano(): uint64 {.inline.} =
       ## Procedure's resolution is nanosecond.
       var res: uint64
       QueryPerformanceCounter(res)
       result = res * queryFrequencyN
 
-    proc fastEpochTime*(): uint64 {.inline.} =
+    proc fastEpochTime*(): uint64 {.
+         inline, deprecated: "Use Moment.now()".} =
       ## Procedure's resolution is millisecond.
       var res: uint64
       QueryPerformanceCounter(res)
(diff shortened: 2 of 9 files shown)


chronos/timer.nim
```diff
@@ -8,45 +8,141 @@
...
+  else:
+    proc QueryPerformanceCounter(res: var uint64) {.
+      importc: "QueryPerformanceCounter", stdcall, dynlib: "kernel32".}
+    proc QueryPerformanceFrequency(res: var uint64) {.
+      importc: "QueryPerformanceFrequency", stdcall, dynlib: "kernel32".}
+
+    var queryFrequencyM: uint64
+    var queryFrequencyN: uint64
+
+    proc fastEpochTimeNano*(): uint64 {.inline.} =
```
arnetheduck: so let's introduce a type `Moment` for the monotonic "current time". we can also introduce an api like:
```
proc now(T: type Moment): Moment = ...
```
which makes it explicit what kind of time we're dealing with, at the call site:
```
let moment = Moment.now()
```

chronos/timer.nim
```diff
@@ -8,45 +8,141 @@
 #    Apache License, version 2.0, (LICENSE-APACHEv2)
 #                MIT license (LICENSE-MIT)
 
-
 ## This module implements cross-platform system timer with
 ## milliseconds resolution.
+##
+## Timer supports two types of clocks:
+## ``fast`` uses the most fast OS primitive to obtain wall clock time.
+## ``mono`` uses monotonic wall clack time (default)
```
arnetheduck: monotonic is no longer wall clock :)

chronos/timer.nim
```diff
@@ -8,45 +8,141 @@
 #    Apache License, version 2.0, (LICENSE-APACHEv2)
 #                MIT license (LICENSE-MIT)
 
-
 ## This module implements cross-platform system timer with
 ## milliseconds resolution.
+##
+## Timer supports two types of clocks:
+## ``fast`` uses the most fast OS primitive to obtain wall clock time.
```
arnetheduck: let's call this system time perhaps, and highlight the risk of it going back in time..

chronos/timer.nim
cheatfate: This will be next step.

chronos.nimble
```diff
@@ -9,26 +9,45 @@ skipDirs      = @["tests"]
...
-        "nim c -r -d:useSysAssert -d:useGcAssert tests/" & tfile,
-        "nim c -r tests/" & tfile,
-        #"nim c -r --gc:markAndSweep tests/" & tfile,
-        "nim c -r -d:release tests/" & tfile
-      ]:
-      echo "\n" & cmd
-      exec cmd
-      rmFile("tests/" & tfile.toExe())
 
+  var testFiles = @[
```
arnetheduck: by the way, in the other projects we have a single `all_tests` which just imports each test... it's pretty nice because it doesn't have to compile each test separately (but you can still do it if you want to, on command line for example..)

chronos/timer.nim
arnetheduck: sure. when we have that, do we still want to expose the public name `fastEpochTimeNano`?

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+
+proc init*(t: typedesc[Moment], value: int64, precision: Duration): Moment =
+  ## Initialize Moment with absolute time value ``value`` with precision
+  ## ``precision``.
+  result.value = value * precision.value
+
+proc fromNow*(t: typedesc[Duration], a: Duration): Moment {.inline.} =
+  result = Moment.now() + a
+
+func getTimestamp*(a: Duration): int {.inline.} =
```
arnetheduck: this looks a bit strange - ie, there's no standard that says something called a timestamp should be in `milliseconds` specifically (it's pretty common with micros these days, or nanos). I think it might be better not to have this function here, and instead use `nanoseconds` and do the rounding at the call site.

the easier non-branching way to do strict round up is `(a.value + 1000000-1) div 1000000`

further, there's no guarantee the resulting millisecond value will fit, needs 41 bits:
```
>>> math.log(2**63 / 1000000, 2)
43.06843143067571
```

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
 
-  proc fastEpochTime*(): uint64 {.inline.} =
-    var t: Timespec
-    discard clock_gettime(CLOCK_REALTIME, t)
-    result = (uint64(t.tv_sec) * 1_000 + uint64(t.tv_nsec) div 1_000_000)
+    proc fastEpochTimeNano(): uint64 {.inline.} =
+      ## Procedure's resolution is nanosecond.
+      var t: Timespec
+      discard clock_gettime(CLOCK_MONOTONIC, t)
+      result = cast[uint64](t.tv_sec) * 1_000_000_000'u64 +
```
arnetheduck: why cast and not just `uint64(t.tv_sec)`?

chronos/timer.nim
arnetheduck: it looks like the template where getTimestamp is used could be rewritten to use durations throughout.. then in the windows section, it could convert to dword accordingly.. 

also, instead of rounding up, an alternative is to use `min(1, timeout)`

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+  ## as integer.
+  result = (a div unit).value
+
+func `$`*(a: Duration): string {.inline.} =
+  ## Returns string representation of Duration ``a`` as nanoseconds value.
+  result = $(a.value)
+
+func `$`*(a: Moment): string {.inline.} =
+  ## Returns string representation of Moment ``a`` as nanoseconds value.
+  result = $(a.value)
```
arnetheduck: hm, it's questionable that this should be printed ever.. given that it's pretty much just a random value

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+  seconds(v)
+
+func toInt*(a: Duration, unit: Duration): int64 {.inline.} =
+  ## Round Duration ``a`` to specific Duration ``unit`` and returns value
+  ## as integer.
+  result = (a div unit).value
+
+func `$`*(a: Duration): string {.inline.} =
+  ## Returns string representation of Duration ``a`` as nanoseconds value.
+  result = $(a.value)
```
arnetheduck: wdyt of some nice human printing here, like: https://github.com/status-im/nim-beacon-chain/pull/197/files#diff-d0557607caf72c6824f50304cedaef9dR98

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+
+func microseconds*(v: Duration): Duration {.inline.} =
+  ## Round Duration ``v`` to microseconds.
+  result = (v div Microsecond) * Microsecond
+
+func milliseconds*(v: Duration): Duration {.inline.} =
+  ## Round Duration ``v`` to milliseconds.
+  result = (v div Millisecond) * Millisecond
+
+func seconds*(v: Duration): Duration {.inline.} =
```
arnetheduck: think it might be useful to return `int64` here and elsewhere, so that it's possible to interface with other systems.. 
two overloads, so we can use them in each direction:
 `func seconds(v: Duration): int64 = ..` and `func seconds(v: int64): Duration` 

consider for example: `echo seconds(10).milliseconds()` - I want this to print 10000, and give me an integer I can serialize for example

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
 
-  proc fastEpochTime*(): uint64
+  proc fastEpochTime*(): uint64 {.deprecated: "Use Moment.now()".}
     ## Returns system's timer in milliseconds.
 
 else:
   error("Sorry, your operation system is not yet supported!")
+
+type
+  Moment* = object
```
arnetheduck: let's document these a little:
```
A Moment in time. Its value has no direct meaning, but can be compared with other Moments. Moments are captured using a monotonically non-decreasing clock.
```

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+    value: int64
+
+  Duration* = object
```
arnetheduck: And here..
```
A Duration is the interval between to points in time.
```

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+  result = (a.value >= b.value)
+
+func `>=`*(a, b: Moment): bool {.inline.} =
+  ## Returns ``true`` if ``a`` bigger or equal ``b``.
+  result = (a.value >= b.value)
+
+func `*`*(a, b: Duration): Duration {.inline.} =
+  result.value = a.value * b.value
+
+func `div`*(a, b: Duration): Duration {.inline.} =
```
arnetheduck: this should probably return a scalar value (int64) that is the ratio between the two durations.. a division would look like: `div(a: Duration, b: SomeInteger): int64`

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+
+func `>=`*(a, b: Duration): bool {.inline.} =
+  ## Returns ``true`` if ``a`` bigger or equal ``b``.
+  result = (a.value >= b.value)
+
+func `>=`*(a, b: Moment): bool {.inline.} =
+  ## Returns ``true`` if ``a`` bigger or equal ``b``.
+  result = (a.value >= b.value)
+
+func `*`*(a, b: Duration): Duration {.inline.} =
```
arnetheduck: this should be `*(a: Duration, b: SomeInteger): Duration`

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+  result.value = a.value * b.value
+
+func `div`*(a, b: Duration): Duration {.inline.} =
+  result.value = a.value div b.value
+
+const
+  Nanosecond* = Duration(value: 1'i64)
+  Microsecond* = Duration(value: 1_000'i64) * Nanosecond
```
arnetheduck: with the above definition for `*`, this simply becomes `Nanosecond * 1000`

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+  Millisecond* = Duration(value: 1_000'i64) * Microsecond
+  Second* = Duration(value: 1_000'i64) * Millisecond
+  Minute* = Duration(value: 60'i64) * Second
+  Hour* = Duration(value: 60'i64) * Minute
+  Day* = Duration(value: 24'i64) * Hour
+  Week* = Duration(value: 7'i64) * Day
+
+  ZeroDuration* = Duration(value: 0'i64)
+  InfiniteDuration* = Duration(value: -1'i64)
```
arnetheduck: I think it might be better to have this as `high(int64)` - that will make it sort naturally and we can saturate operations that would otherwise overflow to this value (just like we saturate the lower bound to 0)

chronos/timer.nim
cheatfate: Because `cast[T]` do not perform any checking, so its faster.

chronos/timer.nim
```diff
@@ -8,45 +8,397 @@
...
+
+func `$`*(a: Duration): string {.inline.} =
+  ## Returns string representation of Duration ``a`` as nanoseconds value.
+  result = $(a.value)
+
+func `$`*(a: Moment): string {.inline.} =
+  ## Returns string representation of Moment ``a`` as nanoseconds value.
+  result = $(a.value)
+
+func isOver*(a: Duration): bool {.inline.} =
```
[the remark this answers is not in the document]
cheatfate: propose variants please, i just do not want to use `finished`.

chronos/timer.nim
cheatfate: Everything should have printed version.

chronos/timer.nim
cheatfate: high(int64) is still valid Duration, while `-1` is marked as Infinite, its normally to define it out of range.

arnetheduck: yes, so it's a question of which is more useful - simple sorting or that last value (out of 2^63), when talking about nanoseconds. I think sorting, in this particular case, because we're using durations for timeouts often, so we want to sort many lists and it seems useful to have the inifinate ones at the end... what are good use cases for that last value?

ie consider the windows DWORD infinate - they use 0xffffffff which is in-range technically, and sorts correctly.

chronos/timer.nim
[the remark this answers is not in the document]
arnetheduck: I'm don't understand sure what `Over` means, in this case.. but `isZero` doesn't sound to bad? specially since we have `ZeroDuration`

chronos/timer.nim
cheatfate: `echo toInt(seconds(10), Millisecond)`

chronos/timer.nim
cheatfate: I just do not want to see `rust` biggest problem (unable to print everything) in Nim.

PR #25, opened 2019-03-24
Breaking Datagram API change.

cheatfate:
```
DatagramCallback* = proc(transp: DatagramTransport,
                                         remote: TransportAddress): Future[void] {.gcsafe.}
                                         message: seq[byte],
                                         remote: TransportAddress,
                                         error: OSErrorCode): Future[void]
```

message - UDP message bytes.
If len(message) == 0 you need to check `OSErrorCode` to obtain error code.

### DIFF
--- a/chronos/transports/datagram.nim
+++ b/chronos/transports/datagram.nim
@@ -11,6 +11,8 @@ import net, nativesockets, os, deques
 import ../asyncloop, ../handles
 import common
 
+export OSErrorCode
+
 when defined(windows):
   import winlean
 else:
@@ -28,7 +30,9 @@ type
     writer: Future[void]        # Writer vector completion Future
 
   DatagramCallback* = proc(transp: DatagramTransport,
-                           remote: TransportAddress): Future[void] {.gcsafe.}
+                           message: seq[byte],
+                           remote: TransportAddress,
+                           error: OSErrorCode): Future[void] {.gcsafe.}
 
   DatagramTransport* = ref object of RootRef
     fd*: AsyncFD                    # File descriptor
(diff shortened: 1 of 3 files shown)


chronos/transports/datagram.nim
```diff
@@ -28,7 +30,9 @@ type
     writer: Future[void]        # Writer vector completion Future
 
   DatagramCallback* = proc(transp: DatagramTransport,
-                           remote: TransportAddress): Future[void] {.gcsafe.}
+                           message: seq[byte],
```
arnetheduck: `openArray[byte]` tends to give more flexibility for the implementation.. a common option later on is to allow the original caller to set up a byte buffer of their choice..

arnetheduck: this is also the kind of place where `Result` is appropriate - giving either an error or a buffer but never both - makes for cleaner code everywhere

cheatfate: @arnetheduck we can't use `openarray[T]` because it can't be captured inside of iterator, so it can't work in async procedures.

chronos/transports/datagram.nim
```diff
@@ -28,7 +30,9 @@ type
...
+                           remote: TransportAddress,
+                           error: OSErrorCode): Future[void] {.gcsafe.}
```
yglukhov: Seems like `error` which can happen here is either not realistic or unrelated to remote-specific message receiving. Thus I would consider removing the `error` from this function signature. For remote-unrelated events I would introduce another callback like `onNetworkChanged`.

cheatfate: Sorry, but `chronos` is cross-platform framework, and if Linux has small number of errors, BSD/MacOS and Windows has much more, so i wish to leave error code here.

PR #33, opened 2019-04-04
Initial tracking mechanism.

cheatfate:
1. Add simple tracking of Datagram and Stream transports.
2. Fix leaks in transports.
2. Fix leaks in tests.
3. Add leaks tests to Datagram and Stream transport tests.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -182,9 +182,15 @@ type
     finishAt*: Moment
     function*: AsyncCallback
 
+  TrackerBase* = ref object of RootRef
+    id*: string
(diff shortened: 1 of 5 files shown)


cheatfate:
@yglukhov i'm not sure, but i think it is possible.

arnetheduck:
I guess this is ok for open/close, but for tracking sends etc will be expensive, with that by-string lookup.. one alternative might be to have events on the dispatcher which are proc callbacks ("var opened: proc (eventname: string)"), so they can be set to anything, including a proc with a capture that holds a reference to the counters

PR #34, opened 2019-04-08
Add SourceLocation.

cheatfate:
All the Future[T] operations using SourceLocation now.
fromProc is now cstring.

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -37,7 +37,6 @@ type
     errorStackTrace*: StackTrace
     stackTrace: StackTrace ## For debugging purposes only.
     id: int
-    fromProc: cstring
 
   # ZAH: we have discussed some possible optimizations where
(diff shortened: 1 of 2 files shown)


chronos/asyncfutures2.nim
```diff
@@ -24,14 +30,14 @@ type
...
+    location: array[2, ptr SrcLoc]
     callbacks: Deque[AsyncCallback]
-
     finished: bool
     error*: ref Exception ## Stored exception
     errorStackTrace*: StackTrace
     stackTrace: StackTrace ## For debugging purposes only.
     id: int
-    fromProc: string
+    fromProc: cstring
```
zah: If the proc name is also helpful, you can embed it in the `SourceLocation` struct. The only required change would be to introduce an extra `static string` parameter to `getSrcLocation`. 

The need for such an extra parameter may disappear in the future - It has been suggested few times in the Nim issue tracker that `instantiationInfo` itself can include this information.

PR #35, opened 2019-04-10
Fix a build issue affecting nim-eth/rlpx.nim

zah:


arnetheduck:
I have no idea what this does - but it looks like the kind of issue that would be valuable to start building up a test base around - even to serve as example of intended behavior

arnetheduck:
@zah is this still relevant?

zah:
This is a good change in general, but not strictly needed right now. @cheatfate once asked me about a strange build problem he was seeing and this fix was addressing it.

arnetheduck:
so rebase and merge? ie if it's at least harmless and sometimes helps?

PR #39, opened 2019-05-28
Ignore ECONNRESET and EPIPE errors on sending.

cheatfate:
With this PR behavior of handling `ECONNRESET` and `EPIPE` errors will be changed. This errors will not raise `TransportOsError` exception, but it will return `0` as return value from `write` procedures.

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -1642,7 +1642,7 @@ proc readLine*(transp: StreamTransport, limit = 0,
       await fut
 
 proc read*(transp: StreamTransport, n = -1): Future[seq[byte]] {.async.} =
-  ## Read all bytes (n == -1) or exactly `n` bytes from transport ``transp``.
+  ## Read all bytes (n <= 0) or exactly `n` bytes from transport ``transp``.
   ##
   ## This procedure allocates buffer seq[byte] and return it as result.
   checkClosed(transp)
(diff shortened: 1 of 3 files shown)


chronos/transports/stream.nim
```diff
@@ -1594,7 +1657,7 @@ proc read*(transp: StreamTransport, n = -1): Future[seq[byte]] {.async.} =
     if transp.offset > 0:
       let s = len(result)
       let o = s + transp.offset
-      if n < 0:
+      if n <= 0:
```
kdeme: Why the need for this change? This will give both for `n = -1` and `n = 0` read all bytes right?
I can understand that n=0 does not make much sense else, but perhaps update the documentation also then.

chronos/transports/stream.nim
```diff
@@ -1475,7 +1538,7 @@ proc readUntil*(transp: StreamTransport, pbytes: pointer, nbytes: int,
   ## Read data from the transport ``transp`` until separator ``sep`` is found.
   ##
   ## On success, the data and separator will be removed from the internal
-  ## buffer (consumed). Returned data will NOT include the separator at the end.
+  ## buffer (consumed). Returned data will include the separator at the end.
```
kdeme: This is only a documentation update and was already the case, correct?

cheatfate: Yep, this was documentation problem only.

chronos/transports/stream.nim
cheatfate: Fixed documentation.

PR #43, opened 2019-07-15
Fixed asynclock race

yglukhov:


chronos/asyncsync.nim
```diff
@@ -109,12 +108,15 @@ proc release*(lock: AsyncLock) =
   ## allow exactly one of them to proceed.
   var w: Future[void]
   if lock.locked:
-    lock.locked = false
+    var unlock = true
     while len(lock.waiters) > 0:
       w = lock.waiters.popFirst()
       if not(w.finished()):
+        unlock = false
```
arnetheduck: is everything that's being called here exception-free, preferably guaranteed by compiler?

yglukhov: Conceptually it is exception-free. Formally it's not.

cheatfate:
Fixed it in https://github.com/status-im/nim-chronos/commit/c4a99447bd1939d736778e053a2afee527446f76

PR #44, opened 2019-07-17
[WIP] Asynchronous tunnels.

cheatfate:


stefantalpalaru:
Why "tunnels" and not "channels"?

cheatfate:
Because `AsyncChannel` is too long, and `Channel` is something reserved when you enable `--threads:on` (you don't need to import channels.nim), so i called it `Tunnel`. You don't like?

stefantalpalaru:
"channel" is an established name for the concept: https://en.wikipedia.org/wiki/Channel_(programming)

"tunnel" is also an established name in networking, for a different concept: https://en.wikipedia.org/wiki/Tunneling_protocol

cheatfate:
@stefantalpalaru i have added procedure `sendNoWait()` so writer threads can perform sending without event loop.

chronos/tunnels.nim
```diff
@@ -0,0 +1,228 @@
...
+  if rtun.count >= cap:
+    if cap == 0: cap = 1
+    var n = cast[ptr UncheckedArray[byte]](allocShared0(cap * 2 * nbytes))
+    var z = 0
+    var i = rtun.rd
+    var c = rtun.count
+
+    while c > 0:
+      dec(c)
+      copyMem(addr(n[z * nbytes]), addr(rtun.data[i * nbytes]), nbytes)
```
zah: Can you add some comments explaining the logic here. Why are we copying the input data to multiple locations in the allocated buffer?

cheatfate: @zah we maintain extendable queue in shared memory, here we copying data items from old queue to new queue (extended) because there no more room for new elements in old queue.

arnetheduck:
I'd probably avoid `tunnel` too given its other meaning (wrapping one thing in another) - this will be confusing. `queue` comes a bit closer, though still not quite.

cheatfate:
I'm closing this because of https://github.com/status-im/nim-chronos/pull/45.

PR #45, opened 2019-07-25
[WIP]: AsyncChannels[T].

cheatfate:
* Add `AsyncChannel[T]` and tests
* Add `AsyncThreadEvent` and tests.
* Refactor OS API declarations to `osapi.nim`.
* Add `awaitForSingleObject()` and `awaitForSelectEvent()`.
* Add `skip()` to some tests.
* Bump version to 2.2.9.

`AsyncChannel[T]` now support multiple writer single reader mode. It is also possible to use write/read to/from channel with both synchronous and asynchronous API.

### DIFF
--- a/chronos/achannels.nim
+++ b/chronos/achannels.nim
@@ -1,0 +1,254 @@
+#
+#    Chronos multi-threading asynchronous channels
+#             (c) Copyright 2018-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+import strutils
+import asyncloop, handles, transport, asyncsync
+
+const hasThreadSupport* = compileOption("threads")
+
+when hasThreadSupport:
+  import locks
+
+type
+  RawAsyncChannelImpl {.pure, final.} = object
+    rd, wr, count, mask: int
+    maxItems: int
+    refCount: int
+    data: ptr UncheckedArray[byte]
+    when hasThreadSupport:
+      lock: Lock
+    eventNotEmpty: AsyncThreadEvent
+    eventNotFull: AsyncThreadEvent
+
+  RawAsyncChannel = ptr RawAsyncChannelImpl
+  AsyncChannel*[Msg] = RawAsyncChannel
+  AsyncChannelError* = object of CatchableError
+
+proc initLocks(rchan: RawAsyncChannel) {.inline.} =
+  ## Initialize and create OS locks.
+  when hasThreadSupport:
+    initLock(rchan.lock)
+  else:
+    discard
+  rchan.eventNotEmpty = newAsyncThreadEvent()
+  if rchan.maxItems > 0:
+    rchan.eventNotFull = newAsyncThreadEvent()
+
+proc deinitLocks(rchan: RawAsyncChannel) {.inline.} =
+  ## Deinitialize and close OS locks.
+  when hasThreadSupport:
+    deinitLock(rchan.lock)
+  else:
+    discard
+
+  close(rchan.eventNotEmpty)
+  if rchan.maxItems > 0:
+    close(rchan.eventNotFull)
+
+proc acquireLock(rchan: RawAsyncChannel) {.inline.} =
+  ## Acquire lock in multi-threaded mode and do nothing in
+  ## single-threaded mode.
+  when hasThreadSupport:
+    acquire(rchan.lock)
+  else:
+    discard
+
+proc releaseLock(rchan: RawAsyncChannel) {.inline.} =
+  ## Release lock in multi-threaded mode and do nothing in
+  ## single-threaded mode.
+  when hasThreadSupport:
+    release(rchan.lock)
+  else:
+    discard
+
+proc newAsyncChannel*[Msg](maxItems: int = -1): AsyncChannel[Msg] =
+  ## Create new AsyncChannel[Msg] with internal queue size ``maxItems``.
+  ##
+  ## If ``maxItems <= 0`` (default value), then queue size is unlimited.
+  let loop = getGlobalDispatcher()
+  result = cast[AsyncChannel[Msg]](allocShared(sizeof(RawAsyncChannelImpl)))
+  result.mask = -1
+  if maxItems <= 0:
+    result.maxItems = -1
+  else:
+    result.maxItems = maxItems
+  result.count = 0
+  result.wr = 0
+  result.rd = 0
+  result.refCount = 0
+  result.data = cast[ptr UncheckedArray[byte]](0)
+  result.initLocks()
+
+proc raiseChannelClosed() {.inline.} =
+  var err = newException(AsyncChannelError, "Channel closed or not opened")
+  raise err
+
+proc raiseChannelFailed() {.inline.} =
+  var err = newException(AsyncChannelError, "Channel synchronization failed")
+  raise err
+
+proc open*[Msg](chan: AsyncChannel[Msg]) =
+  ## Open channel ``chan``.
+  let loop = getGlobalDispatcher()
+  chan.acquireLock()
+  inc(chan.refCount)
+  chan.releaseLock()
+
+proc close*[Msg](chan: AsyncChannel[Msg]) =
+  ## Close channel ``chan``.
+  chan.acquireLock()
+  if chan.refCount == 0:
+    if not(isNil(chan.data)):
+      deallocShared(cast[pointer](chan.data))
+    chan.deinitLocks()
+    deallocShared(cast[pointer](chan))
+  else:
+    dec(chan.refCount)
+  chan.releaseLock()
+
+proc `$`*[Msg](chan: AsyncChannel[Msg]): string =
+  ## Dump channel ``chan`` debugging information as string.
+  chan.acquireLock()
+  result = "channel 0x" & toHex(cast[uint](chan)) & " ("
+  result.add("eventNotEmpty = 0x" & toHex(cast[uint](chan.eventNotEmpty)))
+  result.add(", eventNotFull = 0x" & toHex(cast[uint](chan.eventNotFull)))
+  result.add(", rd = " & $chan.rd)
+  result.add(", wr = " & $chan.wr)
+  result.add(", count = " & $chan.count)
+  result.add(", mask = 0x" & toHex(chan.mask))
+  result.add(", data = 0x" & toHex(cast[uint](chan.data)))
+  result.add(", maxItems = " & $chan.maxItems)
+  result.add(", refCount = " & $chan.refCount)
+  result.add(")")
+  chan.releaseLock()
+
+proc rawSend(rchan: RawAsyncChannel, pbytes: pointer, nbytes: int) =
+  var cap = rchan.mask + 1
+  if rchan.count >= cap:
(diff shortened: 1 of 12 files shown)


cheatfate:
Now the only thing which is left is marshalling/unmarshalling of GC objects through channels.
Cc @zah.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+#             (c) Copyright 2018-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+import strutils
+import asyncloop, handles, transport, asyncsync
+
+const hasThreadSupport* = compileOption("threads")
```
stefantalpalaru: Is this supposed to work at all without `--threads:on`? That would mean no locks whatsoever.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+    release(rchan.lock)
+  else:
+    discard
+
+proc newAsyncChannel*[Msg](maxItems: int = -1): AsyncChannel[Msg] =
+  ## Create new AsyncChannel[Msg] with internal queue size ``maxItems``.
+  ##
+  ## If ``maxItems <= 0`` (default value), then queue size is unlimited.
+  let loop = getGlobalDispatcher()
+  result = cast[AsyncChannel[Msg]](allocShared(sizeof(RawAsyncChannelImpl)))
```
stefantalpalaru: Is this done to avoid GC-safety warnings? A regular heap allocation with `new()` would allow setting a finaliser to close the channel and it's very unlikely that it would be garbage collected before we're done with it.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+  result.add(", refCount = " & $chan.refCount)
+  result.add(")")
+  chan.releaseLock()
+
+proc rawSend(rchan: RawAsyncChannel, pbytes: pointer, nbytes: int) =
+  var cap = rchan.mask + 1
+  if rchan.count >= cap:
+    if cap == 0: cap = 1
+    var n = cast[ptr UncheckedArray[byte]](allocShared0(cap * 2 * nbytes))
+    var z = 0
```
stefantalpalaru: Can we get some descriptive variable names instead of these one-letter ones?

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+
+    rawSend(chan, unsafeAddr msg, sizeof(Msg))
+    chan.eventNotEmpty.fire()
+
+  finally:
+    chan.releaseLock()
+
+proc sendSync*[Msg](chan: AsyncChannel[Msg], msg: Msg) =
+  ## Immediately send message ``msg`` over channel ``chan``. This procedure will
+  ## block until internal channel's queue is full.
```
stefantalpalaru: Shouldn't it block until the queue is empty because the message was consumed and deleted at the destination?

chronos/achannels.nim
cheatfate: Yes it supposed to work without `--threads:on` and in such case `Lock` will not be used, because there will be not possible to perform data race.

chronos/achannels.nim
cheatfate: Nope allocation in shared heap was made to allow all the threads access this memory without any problems. If you allocate heap with `new` you will allocate in thread's local heap.

stefantalpalaru: Yes, but as long as that thread-local heap is not collected and access to it is guarded by locks, it's GC safe, even if the compiler can't verify it.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+
+proc deinitLocks(rchan: RawAsyncChannel) {.inline.} =
+  ## Deinitialize and close OS locks.
+  when hasThreadSupport:
+    deinitLock(rchan.lock)
+  else:
+    discard
+
+  close(rchan.eventNotEmpty)
+  if rchan.maxItems > 0:
```
stefantalpalaru: Why is this conditional?

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+  chan.acquireLock()
+  try:
+    if chan.refCount == 0:
+      raiseChannelClosed()
+
+    if chan.maxItems > 0:
+      # Wait until count is less then `maxItems`.
+      while chan.count >= chan.maxItems:
+        chan.releaseLock()
+        let res = await chan.eventNotFull.wait(InfiniteDuration)
```
stefantalpalaru: Are we setting ourselves up for a deadlock here (or maybe just an avoidable delay)? We want to wait until there's room in the channel's internal queue, so we can add our Msg to it, right? Looks like we're actually waiting for another channel operation to take a look at this queue's size and trigger this event for us (I see `recv()` doing that).

What if there are no other operations on this channel while we wait to send something through it? We want to be able to send something over a buffered channel before `recv()` is called on the other end, don't we?

chronos/achannels.nim
cheatfate: Because if `rchan.maxItems <= 0`, then queue is become infinite and second event `notfull` will be not created and used, so we going to close it only for limited queue.

chronos/achannels.nim
cheatfate: First of all if this function will block when queue is empty, then how we can populate queue? Also delivery confirmation will require adding more synchronization primitives.

chronos/achannels.nim
cheatfate: When Queue has limited size, it is possible that Queue will be full, in such way `send` will wait for reader to consume some data from Queue, so it can store new data to the Queue. If Queue has unlimited size this part of code will not be executed.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+
+proc releaseLock(rchan: RawAsyncChannel) {.inline.} =
+  ## Release lock in multi-threaded mode and do nothing in
+  ## single-threaded mode.
+  when hasThreadSupport:
+    release(rchan.lock)
+  else:
+    discard
+
+proc newAsyncChannel*[Msg](maxItems: int = -1): AsyncChannel[Msg] =
```
mratsim: I don't think the channel work with Ref types/seq/string so we should have a
when not supportsCopyMem(T): {.error: "Reference types are not supported in AsyncChannels".}

chronos/achannels.nim
stefantalpalaru: I misunderstood what `*Sync()` procedures are supposed to do. The comments should be updated to specify they work without an event loop on the caller's side instead of doing unbuffered sending/receiving.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+  chan.releaseLock()
+
+proc close*[Msg](chan: AsyncChannel[Msg]) =
+  ## Close channel ``chan``.
+  chan.acquireLock()
+  if chan.refCount == 0:
+    if not(isNil(chan.data)):
+      deallocShared(cast[pointer](chan.data))
+    chan.deinitLocks()
+    deallocShared(cast[pointer](chan))
```
mratsim: What happens if another thread tries to acquire the lock between

```
chan.deinitLocks()
deallocShared
```

Will it deadlock or crash?

I feel like close should start with `if chan.isNil: return` in case the thread lost a race.

chronos/achannels.nim
```diff
@@ -0,0 +1,254 @@
...
+  result.mask = -1
+  if maxItems <= 0:
+    result.maxItems = -1
+  else:
+    result.maxItems = maxItems
+  result.count = 0
+  result.wr = 0
+  result.rd = 0
+  result.refCount = 0
+  result.data = cast[ptr UncheckedArray[byte]](0)
```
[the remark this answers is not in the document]
cheatfate: Yes in initialization phase.

chronos/achannels.nim
cheatfate: @mratsim Nim's channels working with all this types, to work with it this PR needs some magic from @zah.

chronos/achannels.nim
cheatfate: If API is going to be used properly this part of code will be executed only when all the threads already performed `channel.close()`. In such way there no more threads which will going to use channel.

mratsim:
I think we should add some tests that we can use Helgrind on (with -d:useMalloc) to make sure we didn't miss any glaring issues: http://valgrind.org/docs/manual/hg-manual.html

mratsim:
I think a 2-lock Michael Scott Queue is simple enough and proven robust: https://www.cs.rochester.edu/u/scott/papers/1996_PODC_queues.pdf

DeepinScreenshot_select-area_20190802112126

cheatfate:
Sorry but proposed algorithm not satisfies `AsyncChannel[T]` requirement to have both limited/infinite queue size.

mratsim:
As discussed, it's easily adaptable.

In any case, I do not have a preference for one implementation or the other at this stage, just that we pick one that is simple but robust to start with.

After running the current implementation through helgrind (and adding a script to easily check lock usage correctness), I'm OK with the current implementation.

jangko:
what is the status of this PR? can I use it for some kind of task scheduler?
I'm planning to put it into ETH1 mining simulation along with the upcoming txpool and block finalization.

mratsim:
Can we rebase this PR and merge it?

Is there something missing? AFAIK @zah you had some idea to improve streaming/serializing?

zah:
Yes, I've developed enough of the high-level constructs now, so when this is merged it would be less than a day of work to create the final `AsyncChannel[T]`

kamilchm:
Is there a plan to move it forward?

shayanhabibi:
> Is there a plan to move it forward?

See the [status task runner](github.com/status-im/nim-task-runner) for a functioning implementation of AsyncChannels (functioning AFAIK) that works for non ref objects. Ref objects will require serialization to something like the `ThreadSafeString` that's included in the sys module in the same repo. Also includes a functioning AsyncThreadEvent impl (again AFAIK; I have had some odd troubles here and there but nothing exaggerated). Works with chronos 3.0.6 (however one change would have to be made to the PDispatcher type class to export the handles field).

michaelsbradleyjr:
> See the [status task runner](github.com/status-im/nim-task-runner) for a functioning implementation...

Make sure to use the [`impl/beta2`](https://github.com/status-im/nim-task-runner/tree/impl/beta2) branch of nim-task-runner to get the latest changes. There's a draft PR https://github.com/status-im/nim-task-runner/pull/11 to merge those changes to that repo's `master`, but work on the PR stalled. I'll see what I can do to move it forward in the near future.

The [nim-status example client](https://github.com/status-im/nim-status/blob/master/examples/client.nim) makes heavy use of the `impl/beta2` branch of nim-task-runner.

arnetheduck:
Let's take out `AsyncThreadEvent` from this PR and make it a separate PR  - this is the necessary building block to "wake up" chronos after an item has been posted on a queue meaning that the queues themselves do not necessarily have to be part of chronos

arnetheduck:
https://github.com/status-im/nim-chronos/pull/406 now provides the necessary infrastructure for projects like https://github.com/status-im/nim-taskpools/pulls, https://github.com/status-im/nim-task-runner, QT, etc to integrate with the chronos event loop from their own threads.

PR #46, opened 2019-08-05
Fix emscripten compilation errors.

cheatfate:


chronos/asyncloop.nim
```diff
@@ -165,9 +165,15 @@ export asyncfutures2, timer
 
 # TODO: Check if yielded future is nil and throw a more meaningful exception
 
+const unixPlatform = defined(macosx) or defined(freebsd) or
```
stefantalpalaru: Aren't all these platforms covered by `defined(posix)`?

cheatfate: Yes you are right, but according to https://github.com/nim-lang/Nim/blob/devel/compiler/platform.nim it is also supports MorphOS, SkyOS, Irix, AIX, QNX, Haiku, VxWorks and NintendoSwitch. And this compilation switch was introduced to avoid running chronos in non-tested environments.

cheatfate: https://github.com/nim-lang/Nim/blob/devel/compiler/options.nim#L389-L394

arnetheduck:
compiling for emscripten is the same as compiling for linux/posix - that's the whole point of emscripten - it provides an emulation layer. this fix does not help..

the problem here is that chronos unconditionally raises exceptions on import when the settings it detects from the host are not to its liking - in this case, it calls `initAPI` which calls `newSelector` from `ioselectors_epoll` which calls `getrlimit` which correctly is emulated to return 0. 

basically, even when not using the library to run a loop, selector resources are taken from the system - the epoll should not be allocated until the loop is started.

arnetheduck:
> If you are in Emscripten environment and Chronos was trying to call a missing OS function, it could be argued that the platform detection code was just not complete enough.

the problem is rather that `chronos` can't be _imported_ without raising an exception because `emscripten` correctly implements `getrlimit` (returning 0), for which the construction of a epoll selector raises, and `chronos` creates a selector as part of import - basically, just importing `chronos` causes visible side effects in the form of an epoll handle being opened, among other things, and there's no way to deal with this really.

for a library, this is a smell even if the epoll creation works - apart from the usual cadre of init order problems (which I guess for now are somewhat contained in Nim thanks to the lack of circular imports) it also interferes with dead-code-eliminiation and the principle of pay-for-what-you-use - both at odds with our goal to adapt to the platform that we run the code on. why should the epoll be created if we're only importing chronos to get access to its transport address types?

cheatfate:
@arnetheduck sorry but `getrlimit` which returns `0` is not correct emulation, because it means that you can't open any file and can't create any socket.

cheatfate:
I think we can use compromise here:

For Windows, we will initialize WinSock in global init, but we will not perform IOCP initialization until `getGlobalDispatcher()` will be used.

For Unix, we will block `SIGPIPE` signal in global init, but we will not perform initialization of system queue (epoll/kqueue) until `getGlobalDispatcher()` will be used.

arnetheduck:
sgtm

arnetheduck:
merging this now so we can use it at `interop` - the compromise solution for windows can be done at leisure when we get more thread support

PR #47, opened 2019-08-15
Await is a template now

yglukhov:


chronos/asyncmacro2.nim
```diff
@@ -353,8 +264,32 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+      var chronosInternalTmpFuture: FutureBase
+    chronosInternalTmpFuture = f
+    chronosInternalRetFuture.child = chronosInternalTmpFuture
+    yield chronosInternalTmpFuture
+    chronosInternalTmpFuture.internalCheckComplete()
+    chronosInternalRetFuture.child = nil
+    cast[type(f)](chronosInternalTmpFuture).internalRead()
+  else:
+    static:
+      assert(false, "Await only available within {.async.}")
```
stefantalpalaru: Why not raise an explicit exception instead of having it as a side effect of a failing assertion (which, BTW, doesn't work in release mode)?

yglukhov: It's in the `static:` block. Compile time only.

yglukhov: You were right. `assert` doesn't work in release :). Fixed in subsequent pr.

stefantalpalaru: Static assert() behaviour is very inconsistent:
- in Nim 0.19.6 it's disabled with -d:release
- in 0.20.2 it works with -d:release, but not with -d:danger

I'd say it's best to avoid it all the time.

PR #48, opened 2019-08-15
No Iter in stacktraces

yglukhov:


yglukhov:
Sorry, disallowing waitFor is postponed until later. Related nim issue: https://github.com/nim-lang/Nim/issues/11958

PR #55, opened 2019-10-08
TLS async stream.

cheatfate:


chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,280 @@
...
+          else:
+            # Zero length item means finish
+            rstream.state = AsyncStreamState.Finished
+            break
+
+      if (state and SSL_RECVREC) == SSL_RECVREC:
+        # TLS records required for further processing
+        length = 0'u
+        var buf = sslEngineRecvrecBuf(engine, length)
+        let res = await rstream.rsource.readOnce(buf, int(length))
```
yglukhov: This needs to be interrupted when write queue changes its size from 0 to 1. Otherwise we hang forever, while the server waits for us.

chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,334 @@
...
+  else:
+    msg.add("Internal Error")
+  var err = newException(TlsStreamProtocolError, msg)
+  err.errCode = code
+  err
+
+# proc raiseTlsStreamProtoError*[T](message: T) =
+#   raise newTlsStreamProtocolError(message)
+
+proc getStringState*(state: cuint): string =
```
yglukhov: This shouldn't be public

chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,334 @@
...
+    # Perform TLS cleanup procedure
+    sslEngineClose(engine)
+    # Becase tlsWriteLoop() is ephemeral, but we still need to keep stream state
+    # consistent.
+    rstream.buffer.forget()
+    rstream.stream = nil
+
+proc newTlsClientAsyncStream*(rsource: AsyncStreamReader,
+                              wsource: AsyncStreamWriter,
+                              serverName: string = "",
```
yglukhov: Since it's generally a bad idea to provide empty server name, this argument should be mandatory.

chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,334 @@
...
+
+  let size = max(SSL_BUFSIZE_BIDI, bufferSize)
+  result.sbuffer = newSeq[byte](size)
+  sslEngineSetBuffer(addr result.context.eng, addr result.sbuffer[0],
+                     uint(len(result.sbuffer)), 1)
+  sslEngineSetVersions(addr result.context.eng, uint16(minVersion),
+                       uint16(maxVersion))
+
+  if TLSFlags.NoVerifySN in flags:
+    let err = sslClientReset(addr result.context, "", 0)
```
yglukhov: Are you sure the servername has to be empty here? What about the usecase when we want a self-signed cert on a multidomain+signgleip server?

cheatfate:
```c
/**
 * \brief Prepare or reset a client context for a new connection.
 *
 * The `server_name` parameter is used to fill the SNI extension; the
 * X.509 "minimal" engine will also match that name against the server
 * names included in the server's certificate. If the parameter is
 * `NULL` then no SNI extension will be sent, and the X.509 "minimal"
 * engine (if used for server certificate validation) will not check
 * presence of any specific name in the received certificate.
 *
 * Therefore, setting the `server_name` to `NULL` shall be reserved
 * to cases where alternate or additional methods are used to ascertain
 * that the right server public key is used (e.g. a "known key" model).
 *
 * If `resume_session` is non-zero and the context was previously used
 * then the session parameters may be reused (depending on whether the
 * server previously sent a non-empty session ID, and accepts the session
 * resumption). The session parameters for session resumption can also
 * be set explicitly with `br_ssl_engine_set_session_parameters()`.
 *
 * On failure, the context is marked as failed, and this function
 * returns 0. A possible failure condition is when no initial entropy
 * was injected, and none could be obtained from the OS (either OS
 * randomness gathering is not supported, or it failed).
 *
 * \param cc               client context.
 * \param server_name      target server name, or `NULL`.
 * \param resume_session   non-zero to try session resumption.
 * \return  0 on failure, 1 on success.
 */
int br_ssl_client_reset(br_ssl_client_context *cc,
	const char *server_name, int resume_session);
```

cheatfate: NoVerifySN (Server Name Identification is OFF) is not same as NoVerifyHost (Server Certificate Verification is OFF)

chronos/streams/tlsstream.nim
cheatfate: of course

chronos/streams/tlsstream.nim
yglukhov: Oh I missed that. But in such case the NoVerifySN flag seems to be redundant to `serverName` argument, which should be just set to `""`. Why not just remove the flag altogether. Besides its name is pretty cryptic.

chronos/streams/asyncstream.nim
```diff
@@ -141,6 +141,24 @@ proc copyData*(sb: AsyncBuffer, dest: pointer, offset, length: int) {.inline.} =
   copyMem(cast[pointer](cast[uint](dest) + cast[uint](offset)),
           unsafeAddr sb.buffer[0], length)
 
+proc upload*(sb: ptr AsyncBuffer, pbytes: ptr byte,
```
zah: If there is a reason why `openArray` cannot be used here, let's add a comment about it. 
If there is no such reason, we should always prefer to use `openArray` as it has the potential to make the code provably memory-safe in the future (that is, in [future versions of Nim](https://forum.nim-lang.org/t/5289)).

chronos/streams/asyncstream.nim
```diff
@@ -141,6 +141,24 @@ proc copyData*(sb: AsyncBuffer, dest: pointer, offset, length: int) {.inline.} =
...
+      await sb[].transfer()
+      continue
+    else:
+      copyMem(addr sb[].buffer[sb.offset], pbytes, size)
+      sb[].offset = sb[].offset + size
+      length = length - size
+
+    if length == 0:
+      # We notify consumers that new data is available.
+      sb[].forget()
```
zah: Shouldn't this line be moved after the loop? The logic here effectively checks the loop condition twice on each iteration.

chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,334 @@
...
+    n.add("RecvRec")
+  if (state and SSL_SENDAPP) == SSL_SENDAPP:
+    n.add("SendApp")
+  if (state and SSL_RECVAPP) == SSL_RECVAPP:
+    n.add("RecvApp")
+  result = "{" & n.join(", ") & "} number (" & $state & ")"
+
+proc tlsWriteLoop(stream: AsyncStreamWriter) {.async.} =
+  var wstream = cast[TlsStreamWriter](stream)
+  var engine: ptr SslEngineContext
```
zah: It's possible to avoid the use of pointers here if the code below the `engine` variable assignment is extracted into a separate proc taking `SslEngineContext` as a parameter. It would be called like this:

```nim
tlsWriteLoopAux(stream, if wstream.kind == TLSStreamKind.Server: wstream.scontext.eng
                        else: wstream.ccontext.eng)
```

The `if` part can be further extracted into a template to make it look even more straight-forward.

chronos/streams/tlsstream.nim
```diff
@@ -0,0 +1,334 @@
...
+          let toWrite = min(int(length), item.size)
+
+          if int(length) >= item.size:
+            if item.kind == Pointer:
+              let p = cast[pointer](cast[uint](item.data1) + uint(item.offset))
+              copyMem(buf, p, item.size)
+            elif item.kind == Sequence:
+              copyMem(buf, addr item.data2[item.offset], item.size)
+            elif item.kind == String:
+              copyMem(buf, addr item.data3[item.offset], item.size)
```
zah: These lines are encoding logic that's specific to the `item` type. It seems to me that they'll have to be duplicated if there is another place where an async stream queue item must be written to memory buffer of limited size, so they can be extracted in a helper function.

Furthermore, the limited size is captured in `toWrite`, but this variable is not used elsewhere in this function?

chronos/streams/asyncstream.nim
cheatfate: @zah this procedure is `{.async.}` and i wish to change it to something else, but currently its impossible.

chronos/streams/tlsstream.nim
cheatfate: TLS streams are used not only in TCP/IP network, because we made it as a stream it can be used for every possible types of transports (e.g. pipes, files). In such use-case server name is not available by default. One more reason to leave it is to allow people, who knows little about SSL/TLS protocols to create encrypted communication channel without any problems.

chronos/streams/tlsstream.nim
cheatfate: @zah but all the procedures which will be used requires `ptr SslEngineContext`.

chronos/streams/tlsstream.nim
cheatfate: @zah maybe because you started to review too early? :) Not all ideas is implemented.

chronos/streams/tlsstream.nim
yglukhov: The *default* we want here is that users provide the server name, with an option to not provide it. That is exactly the opposite of what default argument values are. "People who know little" is an invalid argument. This is security related API, and thus security/correctness should always be favored over simplicity.

chronos/streams/tlsstream.nim
yglukhov: Setting `serverName` to empty string is very visible, if you make the argument non-optional. `NoVerifySN` just doesn't tell anything (SN is a Serial Number to me), it only adds more useless code to review and to execute. Here's a nice API example:
```nim
proc greet(name: string, nameIsCF: bool = true) =
  ## Greets name.
  ## If `nameIsCF` is true, the `name` argument is ignored,
  ## and cheatfate is greeted. We've made sure to carefully precalculate
  ## the most recommended default value for this argument.
  if nameIsCheatfate:
    echo "Hello Cheatfate"
  else:
    echo "Hello ", name
```

chronos/streams/asyncstream.nim
zah: Ah, right, pity. Still, a TODO comment may be appropriate in such places, because we might be able to remove the openarray restriction in the future.

chronos/streams/tlsstream.nim
zah: I guess you are referring to the procedures from the `BearSSL` C module? 

Nim doesn't have very good guidelines for writing FFI modules, but if a certain C function takes a pointer, you can wrap it in Nim either as a proc taking `ptr` or as a regular Nim proc taking a value type. One advantage of using regular value types is that they make it easier to stick to the safer subset of the language. If we use some kind of static analysis tool that looks for unsafe code, we'll have less flagged code that must be reviewed. 

So, perhaps the BearSSL wrapper can be modified to not user `ptr` parameters, @yglukhov, what do you think?

yglukhov: @zah, it can indeed. Current bindings were generated automatically though, and doing nice higher level bindings would require some more manual work. At this point I think it's ok for chronos to use the low level bindings as is.

chronos/streams/asyncstream.nim
```diff
@@ -141,12 +141,37 @@ proc copyData*(sb: AsyncBuffer, dest: pointer, offset, length: int) {.inline.} =
...
+  # We notify consumers that new data is available.
+  sb[].forget()
+
 template toDataOpenArray*(sb: AsyncBuffer): auto =
   toOpenArray(sb.buffer, 0, sb.offset - 1)
 
 template toBufferOpenArray*(sb: AsyncBuffer): auto =
   toOpenArray(sb.buffer, sb.offset, len(sb.buffer) - 1)
 
+template harvestItem*(dest: pointer, item: WriteItem, length: int) =
```
yglukhov: I think `copyOut` is a more commonly accepted name for the operation (`item` arg goes in front). Also please avoid templates in favor of procs (optionally inline).

chronos/streams/asyncstream.nim
```diff
@@ -270,6 +295,8 @@ proc readExactly*(rstream: AsyncStreamReader, pbytes: pointer,
   if isNil(rstream.rsource):
     try:
       await readExactly(rstream.tsource, pbytes, nbytes)
+    except CancelledError:
```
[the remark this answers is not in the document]
cheatfate: How to distinguish `TransportOsError`, `TransportIncompleteError`, `TransportLimitError` if i have only `ref Exception`?

chronos/streams/asyncstream.nim
cheatfate: Every copy procedure uses destination as first item, `item` here is source.

PR #57, opened 2019-10-30
Nim 1.0.2 cleanup and fixes.

cheatfate:


tests/config.nims
```diff
@@ -1,4 +1,4 @@
-switch("hints", "off")
-switch("verbosity", "0")
-switch("threads", "on")
+# switch("hints", "off")
+# switch("verbosity", "0")
+# switch("threads", "on")
```
arnetheduck: why disable threads here? at some point, we agreed to keep all code threads-compatible so that we don't end up with designs that will cause problems down the road..

if there are warnings that we've decided manually are safe, afair there are also new ways to disable them locally, something like `{.gcsafe.}:` or something like that, no?

cheatfate: Sorry i just forgot to enable it.

PR #63, opened 2019-12-24
Optimize timers processing

dryajov:
This PR addresses two issues:

a) The current check for `isNil` doesn't seem to make much sense in this context. I was running into errors with `fut` completing twice, checking if `fut` finished seems to make much more sense. (546cc36d79d0ed104b95304922cf331fd0ca2f4b)

b) The entire `timers` list (heap), is being scanned on every `removeTimer` invocation, this incurs in a huge performance overhead when the amount of active timers reaches some threshold. For example, introducing timeouts to nim-libp2p would trigger this in tests pretty often, bringing execution to a complete stall. This PR addresses this issue by avoiding removing timers one by one and instead delay timer removal until the end of the `poll` execution and then cleans them up in wholesale. This increases performance and makes timers usable. (b9c534724c05b4891f6cd7d331d2cc160016e8c1)

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -11,7 +11,7 @@
 include "system/inclrtl"
 
 import os, tables, strutils, heapqueue, lists, options, nativesockets, net,
-       deques
+       deques, sequtils
 import timer
 
 export Port, SocketFlag
@@ -190,9 +190,12 @@ type
   AsyncTimeoutError* = object of AsyncError
     ## Timeout exception
 
-  TimerCallback* = object
+  TimerHandle* = distinct Natural
+
+  TimerCallback* = ref object
     finishAt*: Moment
     function*: AsyncCallback
+    removed: bool
 
   TrackerBase* = ref object of RootRef
     id*: string
@@ -203,6 +206,7 @@ type
     timers*: HeapQueue[TimerCallback]
     callbacks*: Deque[AsyncCallback]
     trackers*: Table[string, TrackerBase]
+    timerHandles: seq[TimerCallback]
 
 proc `<`(a, b: TimerCallback): bool =
   result = a.finishAt < b.finishAt
@@ -227,14 +231,27 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
     result = cast[int32](res)
     result += min(1, cast[int32](mid))
 
+template removeTimers(loop) =
+  var timers = cast[seq[TimerCallback]](loop.timers)
+  timers.keepItIf( not(it.removed) )
+  loop.timers = cast[HeapQueue[TimerCallback]](timers)
+
 template processTimersGetTimeout(loop, timeout: untyped) =
   var count = len(loop.timers)
   if count > 0:
     var lastFinish = curTime
     while count > 0:
+      if loop.timers.len < 1:
+        break
+
+      if loop.timers[0].removed:
+        discard loop.timers.pop()
+        continue
+
       lastFinish = loop.timers[0].finishAt
       if curTime < lastFinish:
         break
+
       loop.callbacks.addLast(loop.timers.pop().function)
       dec(count)
     if count > 0:
@@ -255,6 +272,13 @@ template processTimers(loop: untyped) =
   var count = len(loop.timers)
   if count > 0:
     while count > 0:
+      if loop.timers.len < 1:
+        break
+
+      if loop.timers[0].removed:
+        discard loop.timers.pop()
+        continue
+
       if curTime < loop.timers[0].finishAt:
         break
       loop.callbacks.addLast(loop.timers.pop().function)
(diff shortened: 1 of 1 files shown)


chronos/asyncloop.nim
```diff
@@ -227,14 +231,27 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
     result = cast[int32](res)
     result += min(1, cast[int32](mid))
 
+template removeTimers(loop) =
+  var timers = cast[seq[TimerCallback]](loop.timers)
```
zah: What was the goal here? This cast is quite naughty :)

chronos/asyncloop.nim
```diff
@@ -721,11 +748,47 @@ elif unixPlatform:
...
+
+proc clearTimer*(handle: TimerHandle) =
+  ## Clear the timer associated with the passed handle
+  var loop = getGlobalDispatcher()
+  var timer = loop.timerHandles[handle.Natural]
+  loop.timerHandles[handle.Natural] = nil
+
+  # optimize for handle being at the top
+  if timer == loop.timers[0]:
+    discard loop.timers.pop()
```
zah: What if the second best timer was also previously removed? Shouldn't this go into a loop here popping timers until it finds a one that's not removed?

chronos/asyncloop.nim
```diff
@@ -227,14 +231,27 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
     result = cast[int32](res)
     result += min(1, cast[int32](mid))
 
+template removeTimers(loop) =
+  var timers = cast[seq[TimerCallback]](loop.timers)
+  timers.keepItIf( not(it.removed) )
```
zah: This doesn't preserve the properties of the binary heap. Consider the following heap:

```
     1
   /   \
  15    2
 / \   / \
16 20 3   4
```

If 16 and 20 are removed, the simple filtering will leave 3 and 4 as child nodes of 15, which will violate the heap ordering invariant.

chronos/asyncloop.nim
dryajov: One problem is that removing items from the heap by looping modifies the internal `seq` and breaks everything. Tho as you point out bellow, this also breaks the binary heap.

chronos/asyncloop.nim
dryajov: Not sure what you mean, this will run sequentially, so nothing should modify this.

chronos/asyncloop.nim
```diff
@@ -180,7 +180,7 @@ type
   CallbackFunc* = proc (arg: pointer = nil) {.gcsafe.}
   CallSoonProc* = proc (c: CallbackFunc, u: pointer = nil) {.gcsafe.}
 
-  AsyncCallback* = object
+  AsyncCallback* = ref object
```
cheatfate: What the reason of doing such here?

chronos/asyncloop.nim
```diff
@@ -232,9 +233,17 @@ template processTimersGetTimeout(loop, timeout: untyped) =
   if count > 0:
     var lastFinish = curTime
     while count > 0:
+      if not(loop.timers.len > 0):
```
cheatfate: why not
```nim
if loop.timers.len == 0:
  break
```
?

zah: The `count` check above is essentially doing the same thing, but it "optimised" way. I would just remove it altogether and make the loop

```nim
while loop.timers.len > 0:
  ...
```

chronos/asyncloop.nim
```diff
@@ -255,6 +264,13 @@ template processTimers(loop: untyped) =
   var count = len(loop.timers)
   if count > 0:
     while count > 0:
+      if not(loop.timers.len > 0):
```
zah: Ditto

chronos/asyncloop.nim
dryajov: Not sure if another while is going to add much for readability? Not opposed, it just seems to be more of a style thing?

chronos/asyncloop.nim
dryajov: First, this seems like it would reduce the amount of copying required, second I seem to remember you suggesting we do that?

chronos/asyncloop.nim
zah: I've meant merging the while and the if, not introducing another while.

chronos/asyncloop.nim
cheatfate: This also introduce one more allocation for every possible callback in the `chronos`. And i never suggested such thing, sorry.

chronos/asyncloop.nim
cheatfate: Totally agree with @zah proposition to replace
```nim
while count > 0:
```
with 
```nim
while loop.timers.len > 0:
```

and remove `if`

chronos/asyncloop.nim
dryajov: > This also introduce one more allocation for every possible callback in the chronos
Does it, I'm not sure where? (genuine question)

chronos/asyncloop.nim
```diff
@@ -228,17 +229,19 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
...
+    if loop.timers[0].deleted:
+      discard loop.timers.pop()
+      continue
+
+    lastFinish = loop.timers[0].finishAt
+    if curTime < lastFinish:
+      break
+
+    loop.callbacks.addLast(loop.timers.pop().function)
+  if loop.timers.len > 0:
```
cheatfate: With such approach logic was changed, so `loop.timers.len` will be always `0` at this point, but `count` can be not `0`.

chronos/asyncloop.nim
```diff
@@ -796,25 +817,25 @@ proc withTimeout*[T](fut: Future[T], timeout: Duration): Future[bool] =
...
   var moment: Moment
-  var timerPresent = false
+  var timer: TimerCallback
 
   proc continuation(udata: pointer) {.gcsafe.} =
     if not(retFuture.finished()):
       if isNil(udata):
         # Timer exceeded first.
-        fut.removeCallback(continuation)
+        clearTimer(timer)
```
cheatfate: Why `fut.removeCallback` could be replaced with `clearTimer`?

chronos/asyncloop.nim
```diff
@@ -725,7 +733,19 @@ else:
...
-proc addTimer*(at: Moment, cb: CallbackFunc, udata: pointer = nil) =
+proc setTimer(at: Moment, cb: CallbackFunc, udata: pointer = nil): TimerCallback =
+  ## Arrange for the callback ``cb`` to be called at the given absolute
+  ## timestamp ``at``. You can also pass ``udata`` to callback.
+  let loop = getGlobalDispatcher()
+  result = TimerCallback(finishAt: at,
+                          function: AsyncCallback(function: cb, udata: udata))
+  loop.timers.push(result)
+
+proc clearTimer(timer: TimerCallback) =
```
cheatfate: This procedure should be marked with `{.inline.}`.

dryajov:
> `AsyncCallback = ref object` should be reverted, and all appropriate changes should be reverted too.

done

chronos/asyncloop.nim
dryajov: Not exactly sure what you're saying, there is a break on line 240 and only if the condition of the break never hits, will `len` reach `0`.

chronos/asyncloop.nim
dryajov: Because we need to maintain `addTimer()/removeTimer()` backwards compatible, I can probably add an overload `removeTimer` to take a `TimeoutCallback`, but it's cleaner to introduce a new pair of methods that makes things more evident.

cheatfate: When timer completed first we are removing callback from Future[T], so it will be not called one more time. Your mistake here is that your clearing timer which is already completed.

chronos/asyncloop.nim
cheatfate: Please take a look on old code, `count` is set to number of items in timers queue, but its possible that inside of `while loop.timers.len > 0` you will clean all the timers queue and so `if loop.timers.len > 0:` will never be executed, while old code will be executed because it uses saved value `count` of queue length.

cheatfate: My comment is wrong here, this place is fine.

cheatfate:
Please give this PR proper name in order to avoid misunderstandings in the future.

PR #65, opened 2020-01-10
Cleanup deleted timers

dryajov:
Delete canceled timers to avoid leaks. This operation will only run when select is about to block with a timeout and a (arbitrary) threshold of timers to delete has been reached. This prevents leaking timers and doesn't incur in the penalty of running the remove operation on every `poll`.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -178,7 +178,7 @@ elif unixPlatform:
   from posix import EINTR, EAGAIN, EINPROGRESS, EWOULDBLOCK, MSG_PEEK,
                     MSG_NOSIGNAL, SIGPIPE
 
-const MaxDeletedTimers = 1024
+const MinDeletedTimers = 1024
 
 type
   CallbackFunc* = proc (arg: pointer = nil) {.gcsafe.}
@@ -234,7 +234,7 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
     result += min(1, cast[int32](mid))
 
 template removeDeletedTimers(loop: PDispatcherBase, curTimeout: untyped) =
-  if loop.deletedTimers >= MaxDeletedTimers:
+  if loop.deletedTimers >= MinDeletedTimers:
     if (curTimeout > 0 or curTimeout == -1):
       var before = Moment.now()
       var newHeap = initHeapQueue[TimerCallback]()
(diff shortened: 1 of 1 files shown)


cheatfate:
Sorry, but i dont see any improvements in this PR, you just changed scheduler behavior to produce spikes once per 1024 timer.

I think you missed how scheduler works, because even if there descriptors present scheduler timeout will be set to first minimal timer, this doesn't mean that scheduler is idle to process timers. And i wish not to make any priority for timers in scheduler.

My plan is to implement 
http://www.cs.columbia.edu/~nahum/w6998/papers/sosp87-timing-wheels.pdf
and exactly **Scheme 6: Hash Table with Unsorted Lists in each Bucket**.
So if you want to help here, we can discuss it, but i really do not see any benefits from accepting this PR, sorry.

dryajov:
This is a stopgap solution and it isn't meant to be long term. I agree that we need a more robust approach such as the `Timing Wheel`. 

However, right now timers do leak and that could be a big problem depending on the amount of timers. 

> I think you missed how scheduler works, because even if there file descriptors present, scheduler timeout will be set to first minimal timer, this do not mean that scheduler is idle... And i wish not to make any priority for timers in scheduler.

I understand how the scheduler works and if you observe the implementation closely, you'll see that I only run the cleanup cycle when `curTimeout` is greater than `0` (or `-1`), which means that there are no immediate callbacks triggering and the next timer fires in `curTimeout` milliseconds. This is non-intrusive and without any "spikes" because it will only run when there is no immediate work scheduled.

If you can have the timer wheel ready in a reasonable amount time then I don't see a need for this, otherwise this might come in handy if your solution is not ready and we're seeing high memory consumption due to dangling timers.

dryajov:
BTW, there is nothing to be sorry about 😄

cheatfate:
If you understand scheduler, then why IO events should have lower priority than timers? In your PR you only care about `pendingCallbacks`, but `pendingCallbacks` on this stage do not have any IO events yet. `pendingCallbacks` on this stage keeps only completed futures callbacks and callsooned callbacks.

Also remove item and pop item from heapqueue is O(log(n)) operation, so at one moment of time you are performing `1024 * O(log(n))`, and this is significant. Such spike depending on current CPU usage can produce network timeouts, because you are not processing IO events while deleting timers.

arnetheduck:
mostly obsoleted by #130 - timers are now "almost-free" to keep in the heap queue because the actual callback is released eagerly - there exists a pathological case though where timers created far in the future might never be deleted from the heap queue - this should probably be documented at least

PR #67, opened 2020-01-27
fix future completing twice in `wait`

dryajov:
This should fix https://github.com/status-im/nim-beacon-chain/issues/693. 

With this snippet, I can reproduce the `Feature` completing twice consistently on my machine. Both the `udata` parameter as well as the `fut.finished()` are `true` when running the snippet.

```nim
import chronos

proc main() {.async.} =
  while true:
    try:
      await wait(sleepAsync(4.millis), 4.millis)
    except:
      echo "EXCEPTION ", getCurrentExceptionMsg()

waitFor(main())
```


cheatfate:
Closing it in favor of #68.

PR #84, opened 2020-03-10
prepare nim-chronos for Nim version 1.2

Araq:


Araq:
Reason: The complier is now even more picky about its effect system.

zah:
I'm trying out 1.2 RC, so I'm merging this. Seems quite benign.

PR #85, opened 2020-03-13
some metrics for monitoring futures

stefantalpalaru:
You can see these metrics in action inside a nim-beacon-chain repo, by running `make NIMFLAGS="-d:insecure" eth2_network_simulation` and, a minute later in another terminal, `cd tests/simulation/prometheus; prometheus`.

Now you can go to [this address in your browser](http://localhost:9090/graph?g0.range_input=5m&g0.stacked=1&g0.expr=rate(chronos_processed_callbacks_total%7Bnode%3D%220%22%7D%5B2s%5D)&g0.tab=0&g1.range_input=5m&g1.stacked=1&g1.expr=rate(chronos_loop_timers_total%7Bnode%3D%220%22%7D%5B2s%5D)&g1.tab=0&g2.range_input=5m&g2.stacked=1&g2.expr=chronos_loop_timers_queue%7Bnode%3D%220%22%7D&g2.tab=0&g3.range_input=5m&g3.stacked=0&g3.expr=rate(chronos_poll_events_total%7Bnode%3D%220%22%7D%5B2s%5D)&g3.tab=0&g4.range_input=5m&g4.expr=rate(chronos_future_callbacks_total%7Bnode%3D%220%22%7D%5B2s%5D)&g4.tab=0&g5.range_input=5m&g5.stacked=1&g5.expr=rate(chronos_new_future_total%7Bnode%3D%220%22%7D%5B2s%5D)&g5.tab=0&g6.range_input=5m&g6.stacked=1&g6.expr=rate(chronos_poll_ticks_total%7Bnode%3D%220%22%7D%5B2s%5D)&g6.tab=0) and see something like this: graph

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -8,8 +8,10 @@
 #    Apache License, version 2.0, (LICENSE-APACHEv2)
 #                MIT license (LICENSE-MIT)
 
-import os, tables, strutils, heapqueue, options, deques, cstrutils
-import srcloc
+import os, tables, strutils, heapqueue, options, deques, cstrutils, sets, hashes
+when defined(metrics):
+  import metrics, locks
+import ./srcloc
 export srcloc
 
 const
@@ -73,9 +75,39 @@ type
 var currentID* {.threadvar.}: int
 currentID = 0
 
+when defined(metrics):
+  declareCounter chronos_new_future, "new Future being created"
+
 when defined(chronosFutureTracking):
   var futureList* {.threadvar.}: FutureList
   futureList = FutureList()
+
+when defined(chronosFutureTracking):
+  proc registerPendingFuture(future: var FutureBase) =
+    future.next = nil
+    future.prev = futureList.tail
+    if not(isNil(futureList.tail)):
+      futureList.tail.next = future
+    futureList.tail = future
+    if isNil(futureList.head):
+      futureList.head = future
+    futureList.count.inc()
+    when defined(metrics):
+      chronos_new_future.inc()
+      {.gcsafe.}:
+        withLock(pendingFuturesTableLock):
+          pendingFuturesTable[$future.location[LocCreateIndex]] = pendingFuturesTable.getOrDefault($future.location[LocCreateIndex]) + 1
+
+  proc unregisterPendingFuture(future: var FutureBase) =
+    if future == futureList.tail: futureList.tail = future.prev
+    if future == futureList.head: futureList.head = future.next
+    if not(isNil(future.next)): future.next.prev = future.prev
+    if not(isNil(future.prev)): future.prev.next = future.next
+    futureList.count.dec()
+    when defined(metrics):
+      {.gcsafe.}:
+        withLock(pendingFuturesTableLock):
+          pendingFuturesTable[$future.location[LocCreateIndex]] = pendingFuturesTable.getOrDefault($future.location[LocCreateIndex]) - 1
 
 template setupFutureBase(loc: ptr SrcLoc) =
   new(result)
@@ -87,14 +119,7 @@ template setupFutureBase(loc: ptr SrcLoc) =
   currentID.inc()
 
   when defined(chronosFutureTracking):
-    result.next = nil
-    result.prev = futureList.tail
-    if not(isNil(futureList.tail)):
-      futureList.tail.next = result
-    futureList.tail = result
-    if isNil(futureList.head):
-      futureList.head = result
-    futureList.count.inc()
+    registerPendingFuture(result.FutureBase)
 
 proc newFuture[T](loc: ptr SrcLoc): Future[T] =
   setupFutureBase(loc)
@@ -165,12 +190,8 @@ when defined(chronosFutureTracking):
   proc futureDestructor(udata: pointer) {.gcsafe.} =
     ## This procedure will be called when Future[T] got finished, cancelled or
     ## failed and all Future[T].callbacks are already scheduled and processed.
-    let future = cast[FutureBase](udata)
-    if future == futureList.tail: futureList.tail = future.prev
-    if future == futureList.head: futureList.head = future.next
-    if not(isNil(future.next)): future.next.prev = future.prev
-    if not(isNil(future.prev)): future.prev.next = future.next
-    futureList.count.dec()
+    var future = cast[FutureBase](udata)
+    unregisterPendingFuture(future)
 
   proc scheduleDestructor(future: FutureBase) {.inline.} =
     callSoon(futureDestructor, cast[pointer](future))
@@ -337,6 +358,13 @@ proc addCallback*(future: FutureBase, cb: CallbackFunc, udata: pointer = nil) =
   ## Adds the callbacks proc to be called when the future completes.
   ##
   ## If future has already completed then ``cb`` will be called immediately.
+
+  when defined(metrics):
+    {.gcsafe.}:
+      if future.location[0] != nil:
+        withLock(callbacksByFutureLock):
+          callbacksByFuture.inc($future.location[LocCreateIndex])
+
   doAssert(not isNil(cb))
   if future.finished():
     callSoon(cb, udata)
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -12,7 +12,9 @@ include "system/inclrtl"
 
 import os, tables, strutils, heapqueue, lists, options, nativesockets, net,
        deques
-import timer
+when defined(metrics):
+  import metrics, locks, algorithm, sequtils
+import ./timer, ./srcloc
 
 export Port, SocketFlag
 export timer
@@ -253,8 +255,69 @@ template processTimersGetTimeout(loop, timeout: untyped) =
     if len(loop.callbacks) != 0:
       timeout = 0
 
+when defined(metrics):
+  var
+    callbacksByFuture* = initCountTable[string]()
+    callbacksByFutureLock*: Lock
+    pendingFuturesTable* = initTable[string, int]()
+    pendingFuturesTableLock*: Lock
+  initLock(callbacksByFutureLock)
+  initLock(pendingFuturesTableLock)
+
+  declareCounter chronos_loop_timers, "loop timers"
+  declareGauge chronos_loop_timers_queue, "loop timers queue"
+  declareCounter chronos_poll_ticks, "Chronos event loop ticks"
+  declareHistogram chronos_poll_duration_seconds, "Chronos event loop - duration of poll()",
+    buckets = [0.25, 0.5, 1, 2, 4, 8, Inf]
+  declareCounter chronos_poll_events, "Chronos poll events", ["event"]
+  declareCounter chronos_future_callbacks, "Future callbacks", ["location"]
+  declareCounter chronos_processed_callbacks, "total number of processed callbacks"
+  declareGauge chronos_pending_futures, "pending futures", ["location"]
+
+
+  proc processFutureMetrics() {.gcsafe.} =
+    # Wait until we have a decent amount of data and pick the most frequently
+    # seen futures.
+    const
+      ticksBetweenChecks = 50
+      minimumCallbacksPerCheck = 1000
+      maximumPicksPerCheck = 5
+
+    if chronos_poll_ticks.value.int64 mod ticksBetweenChecks == 0:
+      {.gcsafe.}:
+        withLock(callbacksByFutureLock):
+          var sum = 0
+          for val in callbacksByFuture.values:
+            sum += val
+          if sum >= minimumCallbacksPerCheck:
+            callbacksByFuture.sort()
+            var i = 0
+            for futureLocation, val in callbacksByFuture:
+              if i == maximumPicksPerCheck:
+                break
+              chronos_future_callbacks.inc(val.int64, labelValues = [futureLocation])
+              i.inc()
+
+            {.gcsafe.}:
+              # buggy compiler is buggy
+              callbacksByFuture.clear()
+
+      {.gcsafe.}:
+        withLock(pendingFuturesTableLock):
+          const minimumPendingFutures = 10
+          var i = 0
+          for futureLocation in sorted(toSeq(pendingFuturesTable.keys()),
+                                      proc (x, y: string): int = cmp(pendingFuturesTable[x], pendingFuturesTable[y]),
+                                      SortOrder.Descending):
+            if i == maximumPicksPerCheck or pendingFuturesTable[futureLocation] < minimumPendingFutures:
+              break
+            chronos_pending_futures.set(pendingFuturesTable[futureLocation].int64, labelValues = [futureLocation])
+            i.inc()
+
 template processTimers(loop: untyped) =
   var curTime = Moment.now()
+  when defined(metrics):
+    chronos_loop_timers_queue.set(loop.timers.len.int64)
   while loop.timers.len > 0:
     if loop.timers[0].deleted:
       discard loop.timers.pop()
@@ -263,6 +326,8 @@ template processTimers(loop: untyped) =
     if curTime < loop.timers[0].finishAt:
       break
     loop.callbacks.addLast(loop.timers.pop().function)
+    when defined(metrics):
+      chronos_loop_timers.inc()
 
 template processCallbacks(loop: untyped) =
   var count = len(loop.callbacks)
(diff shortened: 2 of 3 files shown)


chronos/asyncfutures2.nim
```diff
@@ -284,6 +287,9 @@ proc addCallback*(future: FutureBase, cb: CallbackFunc, udata: pointer = nil) =
   ## Adds the callbacks proc to be called when the future completes.
   ##
   ## If future has already completed then ``cb`` will be called immediately.
+  {.gcsafe.}:
```
arnetheduck: not sure if chronos is thread safe, but this is certainly not

also, it seems wrong to be creating strings here and counting the callbacks like this - normally what's done is that if a particular counter is expensive, you count for example 1% of the instances and scale the result accordingly, getting a statistical approximation - something that the metrics library should take care of - if metrics require you to pre-aggregate stuff, something is wrong with the library.

finally, `when metrics` since it's not necessarily part of every codebase to be running this stuff

arnetheduck:
pretty graphs, looks useful as a feature

chronos/asyncfutures2.nim
stefantalpalaru: I don't want all future types ending up in the metrics. Just the most frequently used ones, hence the aggregation, sorting and truncating (further down, in the other file). I know it's ugly, but dragging around hundreds of labels and trying to sort them with Prometheus queries is uglier.

If I add `when defined(metrics):` all over the code, it will get less readable. Is the performance gain worth it?

No, it's not thread-safe. When we start using multiple event loops in parallel and we want combined metrics, we'll need locks.

arnetheduck: It's not just about runtime performance but also compile time and dependency hygiene.. 

also, like this, there's a significant chunk of metrics code intertwined with chronos' callback / future logic - code-wise, the metrics stuff should be factored out into its own functions at least so that it takes up less code-space where the important things are going on - it's already a readability issue that needs to be fixed

chronos/asyncloop.nim
```diff
@@ -729,6 +747,28 @@ elif unixPlatform:
     # poll() call.
     loop.processCallbacks()
 
+    # Wait until we have a decent amount of data and pick the most frequently
+    # seen futures.
+    const
```
zah: I must agree this feature needs more compile-time configuration and perhaps it should be disabled by default for a library such as Chronos.

cheatfate:
First of all `chronos` already has trackers which can be considered as `metrics` replacement. And this feature is cross-platform and used already.

Currently only to track transport or server leaks. But it can be easily extended to track any values inside of `chronos`.

Also i dont think `metrics` are valuable source of information, at least in proposed manner, because even if you have number of callbacks processed and number of timers processed, this do not solve any problems. And actually this problems are already known. Maybe in future we can introduce some glue layer to `nim-metrics` or even migrate `nim-metrics` to chronos (not asyncdispatch). But right now i dont think its a valuable addition.

Please consider that metrics will work only on *nix platforms.

stefantalpalaru:
- added "--skipUserCfg --skipParentCfg" to a test case, which is useful when testing from inside nim-beacon-chain
- made the "metrics" module an optional dependency
- protected the global count table with a lock, when threading is enabled, making the code thread-safe
- moved the "callbacks by future" processing code in its own proc and used that also on Windows
- added another metric to the Windows poll()

chronos/asyncloop.nim
```diff
@@ -12,13 +12,20 @@ include "system/inclrtl"
...
 import ./timer, ./srcloc
 
 export Port, SocketFlag
 export timer
 
-var callbacksByFuture* = initCountTable[string]()
+when defined(metrics):
+  var callbacksByFuture* = initCountTable[string]()
+  when defined(threads):
+    var callbacksByFutureLock*: Lock
```
cheatfate: `callbacksByFuture` must not be global, because async dispatcher will be single-threaded even in multi-threaded environment, so for N threads there will be present N dispatchers.

chronos/asyncloop.nim
```diff
@@ -256,12 +263,47 @@ template processTimersGetTimeout(loop, timeout: untyped) =
     if len(loop.callbacks) != 0:
       timeout = 0
 
-declareCounter chronos_loop_timers, "loop timers"
-declareGauge chronos_loop_timers_queue, "loop timers queue"
+when defined(metrics):
+  declareCounter chronos_loop_timers, "loop timers"
```
cheatfate: all this metrics variables must not be globals.

chronos/asyncloop.nim
cheatfate: Also because dispatcher will be single-threaded locks are not needed.

chronos/asyncloop.nim
```diff
@@ -256,12 +263,47 @@ template processTimersGetTimeout(loop, timeout: untyped) =
...
+  declareGauge chronos_loop_timers_queue, "loop timers queue"
+  declareCounter chronos_poll_ticks, "Chronos event loop ticks"
+  declareCounter chronos_poll_events, "Chronos poll events", ["event"]
+  declareCounter chronos_future_callbacks, "Future callbacks", ["location"]
+  declareCounter chronos_processed_callbacks, "total number of processed callbacks"
+
+
+  proc processCallbacksByFuture() =
```
cheatfate: What the reason to count callbacks per Future?
Almost every possible Future[T] will have only 1 or 2 callbacks, in rare cases it will have more then 2 (allFutures, one), but i really do not understand why this code ever needed?

cheatfate:
Also i understand that you do not care, but all this code can be and should be implemented using internal chronos mechanisms, there no reason to bundle one more dependency.

chronos/asyncloop.nim
stefantalpalaru: > What the reason to count callbacks per Future?

Per Future type, not instance.

> but i really do not understand why this code ever needed

So I can see which Future types are used more often. In NBC's case, it's those produced by readExactly() and write(), peaking at 140-180 per second, every few seconds.

It shows inefficient internal reads and writes, because they correspond to peaks of 20 OS read events and 50 write events per second.

stefantalpalaru:
> all this metrics can be and should be implemented more safely using internal chronos mechanisms

You mean these objects you keep in a table inside the global dispatcher threadvar? https://github.com/status-im/nim-chronos/blob/7ed9f1431a0a8262f988553e9927676cad54e470/chronos/asyncloop.nim#L198

Why resort to that ugly hack when I have proper metrics available, complete with atomic operations (those that confused you when you started commenting that metrics should not be global).

chronos/asyncloop.nim
stefantalpalaru: > Also because dispatcher will be single-threaded locks are not needed.

You can have multiple event loops running in different threads and collect metrics from all of them.

zah:
> I understand that you do not care, but all this metrics can be and should be implemented more safely using internal Chronos mechanisms.

Our experience as C/C++ programmers has thought us that a library without dependencies is something to strive for, a virtue. But why is this the case? One reason is that unfocused dependencies can introduce code bloat (e.g. you need a portable tray icon and you end up adding Qt to your project). But the more probable reason is that in C and C++, integrating dependencies in your build can be quite painful.

In Nim, both of these problems can be avoided if we strive to have small and granular dependencies, especially when we are building all of the libraries ourselves. The alternative to having dependencies is to re-implement the same functionality in multiple places. These end up linked in the same final binary, leading to an unnecessary increase of its code size (both for humans to review and for machines to load in memory).

Reusing the same components also changes the economics of development. It's unlikely that we'll ever have the time to work on the Chronos internal metrics, but we might invest in creating real-time visualisation and analysis tools for nim-metrics if it's used everywhere in our codebases. The smaller and more focused our libraries are, the better.

With all this said, there will be a problem with a dependency to nim-metrics here. To avoid the cyclic graph, we'll need to have a separate package called `metrics-server` that will depend on `metrics(-core)` and `chronos` in the future.

cheatfate:
@stefantalpalaru because dispatchers will run in per-thread mode, there will be no global dispatcher which will cover all the threads, so there no reason to use atomics and globals.

@zah because its dependency which perform `inc(value)` only. And this dependency even not supports custom trackers of some other library, so you need to integrate "globals" and other unsafe constructs to library. So from my point of view if `chronos` are able to count metrics by its own, `nim-metrics` should be able to obtain this values without applying changes to `chronos` scheduler which can seriously affect performance.

cheatfate:
So in future when `chronos` will start support creating multi-threaded servers there will be N threads and N dispatchers which should work in parallel for every connected peer. But all this dispatchers will be required to sync with single Lock at every possible `poll()` call. And i'm not talking there about applying `sort` operation on `CountTable`.

stefantalpalaru:
> So in future when chronos will start support creating multi-threaded servers there will be N threads and N dispatchers which should work in parallel

You support that right now.

> all this dispatchers will be required to sync with single Lock at every possible poll() call

Every 50 calls, because of `if chronos_poll_ticks.value.int64 mod ticksBetweenChecks == 0:`

> there no reason to use atomics and globals

There's no harm either.

> if chronos are able to count metrics by its own, nim-metrics should be able to obtain this values

How?

zah:
> @zah because its dependency which perform inc(value) only. And this dependency even not supports custom trackers of some other library, so you need to integrate "globals" and other unsafe constructs to library. So from my point of view if chronos are able to count metrics by its own, nim-metrics should be able to obtain this values without applying changes to chronos scheduler which can seriously affect performance.

nim-metrics allows you to instantiate the counters as fields within objects as well. Saying that it's just about `inc(value)` misses the point a bit - a metrics library exists so we can have tools that can monitor the data in real-time, archive it, analyse it, etc.

When the responsibilities are cleanly divided between different libraries, we can invest the time to perfect them. If you have some ideas how nim-metrics can be made less intrusive, go ahead and suggest changes to the API - let's work towards creating the best possible metrics library.

cheatfate:
@zah i dont think discussing `nim-metrics` features in `nim-chronos` repository is a right place.
`nim-metrics` it self its just a `inc(value)` and nothing more, and if we going to make it less intrusive it should stay with just `inc(value)`. All other dependencies (web servers, udp broadcasters) should be optional.

Right now, if we going to accept this PR we are including low-quiality web server into low-level networking framework. So question is does "BSD sockets" library must have kernel space web server just to allow you to monitor some metrics? I think approach should be different.

Chronos itself is able to perform `inc(value)` and `nim-metrics` integration should be able to pass or read this values, without any changes to chronos scheduler.

Second problem is that all `nim-metrics` values are global variables, and `chronos` are per thread engine, in such way if we going to use `chronos` in many threads this variables will be not totally correct or totally incorrect. This also creates many problems with `gcsafe`.

stefantalpalaru:
> nim-metrics it self its just a inc(value) and nothing more

You should read the documentation: https://github.com/status-im/nim-metrics/#nim-metrics

> Right now, if we going to accept this PR we are including low-quiality web server into low-level networking framework.

Here's the diff, show me the web server that gets included into Chronos: https://github.com/status-im/nim-chronos/pull/85/files

> Second problem is that all nim-metrics values are global variables, and chronos are per thread engine, in such way if we going to use chronos in many threads this variables will be not totally correct or totally incorrect. This also creates many problems with gcsafe.

No, of course it doesn't create any problem. GC safety is already guaranteed by nim-metrics.

> Chronos itself is able to perform inc(value) and nim-metrics integration should be able to pass or read this values, without any changes to chronos scheduler.

How? How do I get the rate of Future callbacks executed for a specific Future type out of Chronos right now? What's the API for that? Is that data available now, or a couple of years in the future?

zah:
The web server component in nim-metrics is already optional. Merely using metrics in some project doesn't create any security risks. Furthermore, the spirit of the library is that all monitoring should be optional and I suggested that we should disable it by default here.

cheatfate:
@zah right now i see changes in scheduler and this is not acceptable. It seriously affects performance and create single lock sync problems, if `chronos` will be used in multi threaded environment. Also because of global counters it produce incorrect numbers which will be sum of all working dispatchers.

zah:
@stefantalpalaru's argument is that the performance impact is not severe, because there is a notion of sampling (you run the measuring code every N-th poll, where `N` can be controlled).

Furthermore, the performance impact will be zero if the feature is disabled by default which is something I've suggested.

cheatfate:
@zah and every N-th poll you will perform sort operation of thousands Future[T] identifiers just to get metrics information which is not always useful, and after that second thread will perform sort operation of equal dataset. So we are going to freeze all our dispatchers for some time controlled by `N`?

cheatfate:
As i said this metrics should be carefully applied using internal chronos mechanisms, after that we can make a method to transfer this values to any other compatible framework.

stefantalpalaru:
All the new metrics, showing a Medalla node (`make NIMFLAGS="-d:insecure -d:chronosFutureTracking" medalla`) in the initial block syncing phase. No attestation processing yet.

img

zah:
While having metrics from Chronos is an extremely valuable addition to NBC, this PR has been merged without establishing a consensus within the team which is something we should strive to avoid. Even more relaxed practices such as [rough consensus](https://doist.com/blog/decision-making-flat-organization/) would have been violated, so the best way forward is to revert the change as soon as possible and continue the discussion from there.

The concerns of both sides are valid:

1) We need metrics from Chronos
2) The integration was not done in an acceptable way (due to the use of global variables and locking)

The right course of action would have been improving the PR until there are no more blocking objections. Through the use of instance-bound counters (either coming from nim-metrics or through regular variables), the undesired locking would have been avoided and this was a clear direction for how the PR could have been improved.

That said, I would also point out that progress requires both sides to cooperate. This PR has been lingering for too long and we've been deprived from the benefits it brings for no good reason (forcing our hand a little bit to merge it in the current emergency). The metrics library is envisioned to become a very comprehensive solution with instance-bound counters offering fine-grained compile-time configuration mechanisms that are unpractical to develop internally in Chronos and thus it can make Chronos a better library in the long-term. The presented objections towards having it as dependency has been largely addressed (e.g. It was already pointed out that the server components there are optional) and I wish @cheatfate would have more good will in general towards the nim-metrics vision shared by the team.

cheatfate:
@zah 

Sorry, but i can't see how i can cooperate if people totally ignoring my concerns.

You said this metrics are very valuable asset, but i never seen that something got improved using metrics. So the only way when metrics can be used as a tool to find and fix something is when you put metric "file_xxx_line_yyy_metric_zzz" at every line of your code. With current approach the only way of using metrics is to find correlation between one even and another event... What i can say if somebody wants to see such correlations it means there no understanding on how code is working...

For example i can't understand what number of `pending` timers or `pending` callbacks can give to the developer at some moment of time... Without diving deep into this `callbacks` and `timers` this counters are absolutely useless.

Poor understanding of how `async` is working generates poor PRs.

zah:
@cheatfate, it's very hard to argue that having metrics is not useful. Would you claim that you were never surprised by a result we obtained from metrics? As scientists often say, many discoveries don't come from deductions, but rather from observations that makes you say "Hmm, that's strange".

cheatfate:
@zah sorry but i have never been surprised by metrics, usually i have mentioned problems highlighted by metrics many months before it was highlighted by metrics.

Also `nim-metrics` has pretty poor API and design which forces everybody to declare metrics in global scope, as a consequences you are required to put `gcsafe` pragma everywhere where you working with this metric variables... So you turning off pretty useful compiler feature just because you want to use `metrics`. I dont think metrics are worth it.

zah:
> Also nim-metrics has pretty poor API and design which forces everybody to declare metrics in global scope, as a consequences you are required to put gcsafe pragma everywhere where you working with this metric variables... So you turning off pretty useful compiler feature just because you want to use metrics. I dont think metrics are worth it.

This is not true and it's an example for the lack of cooperation that I'm highlighting. The API already allows you to use the counters as fields inside objects and in the future this will get even better because it will be possible to use a compile-time switch that automatically removes the field of a particular metric and all the corresponding updates to it. This is the kind of capability that takes some effort to perfect in a library like nim-metrics and that it doesn't make sense for other libraries to re-invent.

Give the other side the benefit of the doubt, work with them towards improving the API instead of claiming it's just not good and treating it as if it was developed by a third party team that you cannot influence.

cheatfate:
@zah if its **true** why every possible interaction with `metrics` requires to set `{.gcsafe.}`? I think we are all confused here because of `metrics` here.

stefantalpalaru:
> this PR has been merged without establishing a consensus within the team

Libp2p developers jumping through hoops to use the Chronos branch with metrics counts as a consensus where it matters.

> the best way forward is to revert the change as soon as possible and continue the discussion from there

I disagree. I see a very useful feature being blocked for almost 5 months based on non-technical reasons. If anything, we need to stop driving with this handbrake on.

If you hold some fringe beliefs about real programmers debugging without instrumentation, you don't get to veto features useful to us - lesser programmers doing the actual debugging. The feature gets merged sooner, rather than later and whenever you manage to reimplement those metrics you say you understand a lot better, you can make another PR.

This policy is based on common sense and it would allow us to iterate faster, which is very important in this alpha stage.

PR #87, opened 2020-03-31
Add allFinished() primitive.

cheatfate:


stefantalpalaru:
You can also add a deprecation message to `allFutures()` pointing API users to this new procedure. Might also need to replace it in tests.

chronos/asyncfutures2.nim
```diff
@@ -701,6 +701,43 @@ proc allFutures*[T](futs: varargs[Future[T]]): Future[void] =
...
   return retFuture
 
+proc allFinished*[T](futs: varargs[Future[T]]): Future[seq[Future[T]]] =
+  ## Returns a future which will complete only when all futures in ``futs``
+  ## will be completed, failed or canceled.
+  ##
+  ## Returned sequence will hold all the Future[T] objects passed to
+  ## ``allCompleted`` with the order preserved.
+  ##
+  ## If the argument is empty, the returned future COMPLETES immediately.
```
arnetheduck, line 711: if I await the future and it's cancelled, does cancelling raise an exception? 

should probably mention in docs that no exceptions are raised from all itself

chronos/asyncfutures2.nim
```diff
@@ -701,6 +701,43 @@ proc allFutures*[T](futs: varargs[Future[T]]): Future[void] =
...
+
+  proc cb(udata: pointer) {.gcsafe.} =
+    if not(retFuture.finished()):
+      inc(completedFutures)
+      if completedFutures == totalFutures:
+        retFuture.complete(nfuts)
+
+  proc cancellation(udata: pointer) {.gcsafe.} =
+    # On cancel we remove all our callbacks only.
+    for i in 0..<len(nfuts):
```
arnetheduck, line 728: why not `for f in nfuts.mitems`?

chronos/asyncfutures2.nim
arnetheduck: ie this has been something of a question in our talks about all, its raise behaviour - we should probably clarify here in the documentation that even if some of the collected futures "raise", the future returned from all will not raise on its own

zah: This API strikes me a bit as an example of [Groupthink](https://en.wikipedia.org/wiki/Groupthink). Jacek wanted a Result-based API, Eugene maintained that Future is a superset of Result and now we have an API that add very little value and comes with two short-comings:

1) It allocates memory while the previous API didn't need that
2) It increases the code size of Chronos. Both the run-time and the one that we have to refactor when introducing changes.

arnetheduck: the core issue is scheduling: `allFutures` uses a single scheduling/completion/poll slot while awaiting the schedules individually uses many, in the chronos loop. eventually, one would want to keep a set of futures in a collection with optimized scheduling and have "group" operations on that set which would make everyone happy

arnetheduck: also, once/if nim gets move support, it'll be fine - the API was changed due to how rarely it was used correctly

PR #95, opened 2020-04-30
ServerFlags.ReusePort support on POSIX

stefantalpalaru:


stefantalpalaru:
libuv prefers SO_REUSEPORT to SO_REUSEADDR on \*BSD and macOS: https://github.com/libuv/libuv/blob/43b456dcb375319ad0e45931ddddd88fd3dd7ee1/src/unix/udp.c#L495

"On the BSDs, SO_REUSEPORT implies SO_REUSEADDR but with some additional refinements for programs that use multicast." - https://github.com/libuv/libuv/blob/43b456dcb375319ad0e45931ddddd88fd3dd7ee1/src/unix/udp.c#L467

See also: https://github.com/libuv/libuv/blame/5736658bee0eede05caa55d66b9ad337e3b77266/docs/src/udp.rst#L38

PR #97, opened 2020-05-05
Remove unreachable raise from initTAddress

kdeme:
Avoiding to have unnecessary exceptions in `raises:[]`.


zah:
This is a fine change, how come it's not being merged?

PR #107, opened 2020-07-03
Fix cancellation race when low level futures are already completed, w…

cheatfate:
Fix cancellation race while cancelling already finished futures.
Added test.

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -49,7 +49,7 @@ template createCb(retFutureSym, iteratorNameSym,
           {.gcsafe.}:
             next.addCallback(identName)
     except CancelledError:
-      retFutureSym.cancel()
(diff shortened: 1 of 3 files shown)


chronos/asyncmacro2.nim
```diff
@@ -273,8 +273,10 @@ template await*[T](f: Future[T]): auto =
     chronosInternalTmpFuture = f
     chronosInternalRetFuture.child = chronosInternalTmpFuture
     yield chronosInternalTmpFuture
-    chronosInternalTmpFuture.internalCheckComplete()
     chronosInternalRetFuture.child = nil
+    if chronosInternalRetFuture.mustCancel:
+      raise newException(CancelledError, "")
```
arnetheduck, line 278: set a message here like "Future operation cancelled"?

PR #108, opened 2020-07-06
Make Future tracking and stack traces optional

cheatfate:
- Make Future tracking optional via `-d:chronosFutureTracking` compilation flag.
- Make Future stack traces optional via `-d:chronosStackTraces` compilation flag.
- Address review comment https://github.com/status-im/nim-chronos/pull/107#pullrequestreview-442368768

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -36,8 +36,9 @@ type
     stackTrace: StackTrace ## For debugging purposes only.
     mustCancel*: bool
     id*: int
-    next*: FutureBase
-    prev*: FutureBase
+    when defined(chronosFutureTracking):
+      next*: FutureBase
+      prev*: FutureBase
 
   # ZAH: we have discussed some possible optimizations where
   # the future can be stored within the caller's stack frame.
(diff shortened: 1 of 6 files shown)


chronos/asyncfutures2.nim
```diff
@@ -36,8 +36,9 @@ type
     stackTrace: StackTrace ## For debugging purposes only.
     mustCancel*: bool
     id*: int
-    next*: FutureBase
-    prev*: FutureBase
+    when defined(chronosFutureTracking):
```
arnetheduck, line 39: shouldn't this cover the stack trace as well? it's a pure debugging feature, and one of the most expensive things nbc does, according to profiling, when not syncing

cheatfate: stack trace is not related to `chronosFutureTracking`.

arnetheduck: yeah, but maybe it should be, naming this feature `chronosDebug` or something like that?

cheatfate: `chronosStackTraces`.

PR #109, opened 2020-07-07
start documenting the library

stefantalpalaru:
This should get the ball rolling.


README.md
```diff
@@ -8,23 +8,36 @@
...
+
+### Concepts
+
+Chronos implements the async/await paradigm in a self-contained library, using
+macros, with no specific helpers from the compiler.
+
+Our event loop is called a "dispatcher" and a single instance per thread is
+created, as soon as one is needed.
+
+To trigger a dispatcher's processing step, we need to call `poll()` - either
```
cheatfate, line 31: Even if its possible this is not recommended way, so recommended way is to use `waitFor()` and `runForever()` only.

stefantalpalaru: It depends. If you want to exit gracefully by checking a global variable between event loop ticks, it's cleaner to use `poll()`.

Anyway, these are high-level concepts. The recommendations will come later, when we get to the API and examples.

cheatfate: @stefantalpalaru with such approach you can stuck forever in `poll()` call, it is much better to use `callSoon(someCallback)` and check your variable in `someCallback`, and this will never stuck.

stefantalpalaru: If `poll()` gets stuck, callbacks are no longer being processed, right?

cheatfate: For your specific use-case when you want to check some global variable for some flag and call `poll()` it can happens that if there no `callbacks` and `timers` but some descriptors are waiting (for example accept() waiting for new client) you could stuck in `poll()` until new client will not be accepted. Its why you should always use `callSoon()` and populate dispatcher's callbacks list with at least one callback so it will be called all the time... This of course very bad idea too, because in such case there will be no pauses and your CPU usage can increase significantly.

cheatfate: So its why cycles like this:
```nim
while someGlobalFlagIsTrue:
  poll()
```
Is very bad idea, but my `yieldAsync` is also not a proper solution, just because it leads to infinite CPU usage.

stefantalpalaru:
You're worried about this? https://en.wikipedia.org/wiki/Busy_waiting

I don't think we can avoid it, in an event loop (unless we introduce artificial pauses that affect our responsiveness), but you should know that `runForever()` does the same busy waiting: https://github.com/status-im/nim-chronos/blob/ce6e7d17b11b33d5dd17123827dda43af553a7eb/chronos/asyncloop.nim#L939-L942

Maybe we should consider `poll()` getting stuck a bug, instead of a supported state. Same for hogging the event loop with long-running tasks in the same thread.

README.md
stefantalpalaru: You're worried about this? https://en.wikipedia.org/wiki/Busy_waiting

I don't think we can avoid it, in an event loop (unless we introduce artificial pauses that affect our responsiveness), but you should know that `runForever()` does the same busy waiting: https://github.com/status-im/nim-chronos/blob/ce6e7d17b11b33d5dd17123827dda43af553a7eb/chronos/asyncloop.nim#L939-L942

Maybe we should consider `poll()` getting stuck a bug, instead of a supported state. Same for hogging the event loop with long-running tasks in the same thread.

cheatfate:
@stefantalpalaru just remove `poll()` from documentation, it is not supposed to be **public** API procedure, but it is used in tests to protect/check behavior.

stefantalpalaru:
> just remove poll() from documentation

Do you understand that the main purpose of this documentation is to help people make sense of the library, as it's used right now in NBC?

cheatfate:
@stefantalpalaru i'm totally understand the point, and its why i think that `poll()` should not be used even in `nbc`.

README.md
arnetheduck: instead of not documenting it, it's better to document the pitfalls of using it, with a recommendation to avoid it - when it's part of public api, people will use it - less information is always worse, in such cases

cheatfate: There two procedures which should be used as public API its `runForever` and `waitFor` all other procedures MUST not be used at any cost. I wish to disable `*` for `poll()` but i can't. I can write book explaining why you should not use `poll()` and most of the readers will drop this book because it will become boring. That's why documentation should avoid sentences like this:

> To trigger a dispatcher's processing step, we need to call `poll()`.

You should not call `poll()` directly at any cost, this should be present in documentation.

README.md
```diff
@@ -74,10 +74,28 @@ proc waitFor*[T](fut: Future[T]): T =
   return fut.read()
 ```
 
+### Async procedures and methods
+
+The `{.async.}` pragma will transform a procedure (or a method) returning a
+specialised `Future` type into a closure iterator. If there is no return type
```
cheatfate, line 81: `{.async.}` pragma do not transform procedure into closure iterator. It transforms procedure to 3 procedures. One of this 3 procedures is a closure iterator.

README.md
```diff
@@ -74,10 +74,28 @@ proc waitFor*[T](fut: Future[T]): T =
...
+```nim
+proc p() {.async.} =
+  await sleepAsync(100.milliseconds)
+
+echo p().type # prints "Future[system.void]"
+```
+
+Whenever `await` is encountered inside an async procedure, control is passed
+back to the dispatcher for as many steps as it's necessary for the awaited
+future to complete (or be cancelled). `await` calls the equivalent of
```
cheatfate, line 93: complete, failed or be cancelled.

README.md
```diff
@@ -36,7 +36,8 @@ processed.
 `Future` objects encapsulate the result of an async procedure, upon successful
 completion, and a list of callbacks to be scheduled after completion or
 cancellation. (These explicit callbacks are rarely used outside Chronos, being
```
cheatfate, line 38: callbacks are scheduled when Future changes it state from `FutureState.Pending`. So callbacks are scheduled also when Future got failed or cancelled.

cheatfate: > A Promise/Future is a proxy for a value not necessarily known when the promise/Future is created. It allows you to associate handlers with an asynchronous action's eventual success value or failure reason. This lets asynchronous methods return values like synchronous methods: instead of immediately returning the final value, the asynchronous method returns a promise to supply the value at some point in the future.

README.md
```diff
@@ -1,30 +1,201 @@
...
+```nim
+proc p4() {.async.} =
+  await p1()
+  await p2()
+  # Also takes a single second for both futures sleeping concurrently.
+
+waitFor p4()
+```
+
+Don't let `await`'s behaviour of giving back control to the dispatcher surprise
```
cheatfate, line 129: This statement can be easily transformed to more simple one:
> Do not use `await` if you want to be safe.

But if you not using `await` you will be unable to perform concurrent actions.

chronos/asyncmacro2.nim
```diff
@@ -272,7 +272,19 @@ template await*[T](f: Future[T]): auto =
       var chronosInternalTmpFuture {.inject.}: FutureBase
     chronosInternalTmpFuture = f
     chronosInternalRetFuture.child = chronosInternalTmpFuture
+
+    # This "yield" is meant for a closure iterator in the caller.
     yield chronosInternalTmpFuture
+
+    # By the time we get control back here, we're guaranteed that the Future we
+    # just yielded has been completed, through a very complicated mechanism in
```
cheatfate, line 280: we're guaranteed that the Future just yielded has changed its state.

cheatfate:
@stefantalpalaru Is it general documentation, or its some kind of "do not use it" warning?

README.md
stefantalpalaru: So you make a choice: no changes to the global state or concurrency.

stefantalpalaru:
> Is it general documentation, or its some kind of "do not use it" warning?

It's an unusually friendly documentation, after what I experienced reading the code.

README.md
stefantalpalaru: We don't want to scare them off so soon. Let's break it to them in small chunks.

README.md
stefantalpalaru: How else do you stop the dispatcher for a controlled shutdown?

README.md
stefantalpalaru: Chronos defines "completion" as success, failure or cancellation:
https://github.com/status-im/nim-chronos/blob/f856c885facfb45074b95c5df42f6468f2cc86da/chronos/asyncfutures2.nim#L147-L154

stefantalpalaru: Terminology clarified.

README.md
stefantalpalaru: Clarified.

README.md
stefantalpalaru: > Do not use `await` if you want to be safe.

Too generic to be of any use to newcomers.

chronos/asyncmacro2.nim
stefantalpalaru: Clarified.

README.md
cheatfate: If you want to document `poll()` document everything properly. `poll()` is exactly not for newcomers.

README.md
cheatfate: Use `waitFor` if you need controlled shutdown.

README.md
```diff
@@ -1,30 +1,203 @@
...
+Exceptions inheriting from `Defect` are treated differently, being raised
+directly. Don't try to catch them coming out of `poll()`, because this would
+leave behind some zombie futures.
+
 ## TODO
   * Pipe/Subprocess Transports.
   * Multithreading Stream/Datagram servers
-  * Future[T] cancelation
-  
+  * Future[T] cancellation
```
arnetheduck, line 199: cancellation support   is there already, no? @cheatfate

cheatfate: yep it should be removed.

stefantalpalaru: Removed.

PR #111, opened 2020-07-27
more docs

stefantalpalaru:


stefantalpalaru:
Should `CancelledError` even be documented, or is it an internal implementation detail that the user is not supposed to catch?

If so, why does it inherit from `CatchableError` so it risks being caught by mistake?

zah:
I've pondered about `CancelledError` myself and I've almost convinced myself that it belongs to a third category of exceptions (i.e. we can define a new exception type called `ExecutionAbortion` that derives directly from Exception). APIs similar to Chrono's `cancel` or Java's [`Thread.interrupt`](https://stackoverflow.com/questions/3590000/what-does-java-lang-thread-interrupt-do) can introduce specific exception types in that category.

arnetheduck:
ah - we're now fully on the slippery slope ;) you're now beginning to understand why the divide between catchable and defect makes little sense - and why exceptions in general are hard to work with - because _all_ code must be aware of _all_ other code and possible exceptions, including transitive dependencies, that each may raise their own idea of  unknown exceptions that "they" define as special, at runtime, with no static typing to help.

you're in for a world of pain trying to reason about the resulting code with two categories (as seen in nbc, if the example needs to be fresh and "in nim"), specially when you add your requirement that `Defect` should "sometimes" be catchable - that "sometimes" ends up being "always" once you venture outside the 80% "it looks like it works" comfort zone you got to in 20% of the time. 

adding a third category doesn't help - it only shows that in the initial design, you thought about a subset of problems and now as you expand the solution, you're finding that it was incomplete - it will keep being incomplete because exceptions and inheritance trees generally are open types where anything goes and everyone can make their own rules that impact not just the surrounding modules but all code across the application.

stefantalpalaru:
What is currently being done to avoid catching `CancelledError` and interfering with the Future cancellation machinery:

https://github.com/status-im/nim-beacon-chain/blob/0a96e5f5648cf186e3cf86a829db45d73f2ed249/beacon_chain/eth2_network.nim#L277-L281

stefantalpalaru:
I made CancelledError an object of Exception in order to get rid of that ridiculous requirement of doing

```nim
except CancelledError:
  raise
```
before any `except CatchableError:`.

I also corrected the docs and some comments.

arnetheduck:
> I made CancelledError an object of Exception

well, in doing so you're going against the (broken) norm that Nim itself is trying to establish: https://nim-lang.org/docs/manual.html#exception-handling-exception-hierarchy - that's probably worse than having to be explicit about cancellation (ie being explicit about the fact that cancellations may happen, and that they follow a different flow, is not a bad thing)

stefantalpalaru:
> being explicit about the fact that cancellations may happen, and that they follow a different flow, is not a bad thing

It is a very bad thing to have hidden exception handling rules in order not to break cancellation propagation. Telling the user to always catch `CatchableError` is one thing, because it should be a best practise all over the Nim codebase, but having to catch and raise `CancelledError` each and every time before catching `CatchableError` is too much to ask from an API user.

Case in point: NBC and nim-libp2p not managing to do that consistently.

arnetheduck:
> Case in point: NBC and nim-libp2p not managing to do that consistently.

neither of NBC or libp2p manage to catch exceptions consistently at all, regardless of their type, origin and nature - perhaps one needs to start there ;) the dynamic type involved means one cannot consistently handle these situations without writing lots of exception-handling code - once you start doing it, the code becomes convoluted, but at least with explicit code, after many iterations and bugs and missing exception handling added, it becomes possible to understand what's going on

arnetheduck:
basically, just because `CancelledError` represents a different flow does not mean that you're not supposed to clean up resources - which is one of the main reasons to write an exception handler - ie what goes inside `except CancelledErrors` is the same code that goes inside `except CatchableError`, a lot of the time. 

Hide it, and you will merely have a new type of bugs.

stefantalpalaru:
> clean up resources

You're supposed to do that with Future callbacks. Eventually by capturing the future pointer (ref object) in their closure environment and checking its state if needed.

arnetheduck:
> You're supposed to do that with Future callbacks.

well.. uh, I thought the whole point of `async` / `await` was to avoid callbacks and write code that looks like regular code and not lots of callbacks?

zah:
@arnetheduck, I don't find your arguments too convincing. In fact, I would turn the tables on you and claim that `CancelledError` is something that's very difficult to express in a world without exceptions. It's a pretty good example for the term ["cross-cutting concern"](https://en.wikipedia.org/wiki/Cross-cutting_concern), where a pattern or a property of the code is duplicated across the entire codebase and it meddles with all other abstractions.

To have something like `CancelledError` in the world of error codes and Results, you would need to:

1) Turn every return type into a logical enum (because the cancellation needs to be handled in a special way)
2) If you already had an enum return type, you'll need to add `Cancelled` as an extra possibility there. Please note that each enum will have its own unique value for `Cancelled`, so you'll have to write code to remap between them.

Point 2) in particular is one of the reasons why a language such as Zig that pushes very strongly towards error codes has settled on something called ["Error Sets"](https://ziglang.org/documentation/master/#toc-Error-Set-Type) instead of regular enums.

With exceptions, the designer of the API has more options, so let's discuss the merits of the different designs.

As Stefan argued, requiring the user to take care of `CancelledError` on every level seems like an inappropriate default, because the potential for error is very high and the likehood for the intention of the code to differ from the "standard" behavior is quite low.

A `CancelledError` is not a `Defect` though, because it's obviously intentionally triggered in the code and it must be handled at some point. It should be noted that in C# for example, the design of the whole API is slightly different. The cancellation is handled through an object called `CancellationToken` which is implicitly passed forward in the code. The API requires you to explicitly specify the boundary where the cancellable operation starts, so attempts to read the result of that particular operation must handle the possibility for cancellation.

Nevertheless, even with this design we have a new category of stack unwinding. Most of the code should transparently handle it, but it's not supposed to trigger the usual diagnostic data gathering associated with Defects. We can resolve this in two ways:

1) Introduce a 3rd category of Exceptions
2) Introduce a new type sitting above `Defect` (e.g. `UntrackedError`)

Option 1) is easier to push at this point, although 2) may be slightly better in a more holistic world-view (it will be able to also express the category of errors I've called "catastrophic system failures" - e.g. running out of memory).

zah:
Resource clean up should be performed in destructors and `finally` blocks. Not in `if` statements, duplicated code in `except CatchableError: ... except CancelledError: ...` block or other similar fragile techniques that can be broken with early returns, `break`, `continue`, etc.

dryajov:
> It's a pretty good example for the term "cross-cutting concern", where a pattern or a property of the code is duplicated across the entire codebase and it meddles with all other abstractions.

Yep, without taking this into account, any cancelation api is going to end up being very very hard to consume, if not at all broken. 

We should definitely make `CanceledError` a different type, ie not inherit from `CatchableError`, if not a different category altogether, as it has been already mentioned - but I haven't giving this too much thought, so I'll be neutral about it for now.

arnetheduck:
> error codes

I might have missed something, but we were not discussing error codes or Result here?

> destructors

I think we're talking about a different language - Nim is (still) ref-counted, thus doesn't have destructors, the way they're normally defined - this leads to designs where references outlive their scope (that's the point) - which in turns means that you cannot place cleanup in `finally` because.. well.. you're not supposed to do the cleanup there but rather register the instance for some other cleanup scheme, move on and happily use your ref.

If you're talking about an imaginary future where Nim, at its core, is no longer ref-counted and chronos turns deterministic, well, that's a different discussion than deviating from the established practise of the language as it is defined _now_.

zah:
I took the detour because you tried to deliver your typical critique towards exceptions without discussion the actual issue yourself.

Nim does have non-ref [automatic](https://en.wikipedia.org/wiki/Automatic_variable) stack variables whose destructors will work in a very similar way to C++, so many of the same RAII techniques apply equally well here.

arnetheduck:
> established practise 

ie the compiler prints a warning when you create new error categories because.. well.. you're not supposed to do that.

arnetheduck:
> actual issue

the actual issue here is that consumers of the library don't know which exceptions to catch, when. The suggested solution is to flip create yet another category of exceptions outside of what the language recommends so that sometimes when you're supposed to catch exceptions, they go through unnoticed.

When you do that, you will get the typical bugs that the typical critique is about.

stefantalpalaru:
> I thought the whole point of async / await was to avoid callbacks

No, just to hide them, but sometimes you need the escape hatch.

There is even a dedicated `Future.cancelCallback` pseudo-field in the public API to register a callback that is only called on cancellation.

> Introduce a new type sitting above Defect (e.g. UntrackedError)

I thought about it, but something tells me I can't rely on catching defects without fighting the compiler (or having to do major refactoring in the near future). I mean, it's already nagging me about inheriting directly from `Exception`...

PR #113, opened 2020-07-31
Rate limiter implementation.

cheatfate:
This is not very precise but incredibly fast rate limiter, I was able to see 23M requests/second on my 10y old Xeon.
Implementation was inspired by this article https://blog.cloudflare.com/counting-things-a-lot-of-different-things/


cheatfate:
This is dependency to fix https://github.com/status-im/nim-beacon-chain/issues/1359

arnetheduck:
per discussions from a distant past, we'll instead be going with a token / leaky bucket algorithm for rate limiting

Issue #116, opened 2020-08-05
Async transformation do not protects Future[void] result assignment.


cheatfate:
```nim
import chronos
proc someFoo() {.async.} =
  result = newFuture[int]("another one")
when isMainModule:
  waitFor someFoo()
```
This example is able to get compiled.

PR #117, opened 2020-08-05
Produce an error when the result variable is used in void async procs

zah:


cheatfate:
This fixes https://github.com/status-im/nim-chronos/issues/116

PR #120, opened 2020-08-09
yieldAsync()

stefantalpalaru:


cheatfate:
With continuous issues with timers `yieldAsync` implementation using `sleepAsync` looks dumb, it should be implemented using `callSoon` because `callSoon` puts callback directly into `pendingCallbacks` sequence.

First of all presence of this primitive allows developers to workaround and not fix actual problems.

stefantalpalaru:
> With continuous issues with timers yieldAsync implementation using sleepAsync looks dumb, it should be implemented using callSoon because callSoon puts callback directly into pendingCallbacks sequence.

Go ahead and make a PR for this proposal of yours.

> presence of this primitive allows developers to workaround and not fix actual problems

Breaking blocking tasks into smaller pieces and giving control back to the even loop is an established pattern in the async world.

They use `setTimeout()` to do that in JavaScript: https://stackoverflow.com/questions/10344498/best-way-to-iterate-over-an-array-without-blocking-the-ui/10344560#10344560

PR #121, opened 2020-08-11
asyncCheck(): don't raise exception on cancellation

stefantalpalaru:
When we document this procedure, we need to make it clear that exceptions are being raised in `poll()` by that new callback.

The name doesn't help at all.


cheatfate:
All cancellations should be handled and not swallowed, if you still using this primitive use
```
try:
  <procedure code>
except CancellationError as exc:
  discard
```
to consume cancellation exception explicitely.

PR #124, opened 2020-09-03
Bugfix: Correct origin locations for futures created in the async macro

zah:


cheatfate:
Closing it because found actual problem and fixed macro code https://github.com/status-im/nim-chronos/pull/125

PR #127, opened 2020-09-09
small fixes

arnetheduck:
* more efficient codegen for nil check (much less code)
* release futures earlier in AsyncEvent
* release finished future earlier in AsyncQueue
* avoid searches for futures (deque variant unused / broken)
* avoid catching defects

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -142,9 +142,7 @@ proc newAsyncEvent*(): AsyncEvent =
   # Workaround for callSoon() not worked correctly before
   # getGlobalDispatcher() call.
   discard getGlobalDispatcher()
-  result = new AsyncEvent
-  result.waiters = newSeq[Future[void]]()
-  result.flag = false
+  AsyncEvent(waiters: newSeq[Future[void]](), flag: false)
 
 proc wait*(event: AsyncEvent): Future[void] =
   ## Block until the internal flag of ``event`` is `true`.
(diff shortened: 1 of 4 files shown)


chronos/asyncsync.nim
```diff
@@ -145,22 +146,19 @@ proc newAsyncEvent*(): AsyncEvent =
...
+  var w = newFuture[void]("AsyncEvent.wait")
+
   if not(event.flag):
-    var w = newFuture[void]("AsyncEvent.wait")
     event.waiters.add(w)
-    try:
-      await w
-    finally:
-      event.removeWaiter(w)
+  else:
```
cheatfate, line 158: This procedure could be cancelled, please return `try/finally` block.

chronos/asyncsync.nim
```diff
@@ -198,32 +197,13 @@ proc wakeupNext(waiters: var seq[Future[void]]) {.inline.} =
...
-    waiters.delete(index)
-
-proc removeWaiter(waiters: var Deque[Future[void]],
-                  fut: Future[void]) {.inline.} =
-  var nwaiters = initDeque[Future[void]]()
-  while len(waiters) > 0:
-    var waiter = waiters.popFirst()
-    if waiter != fut:
-      nwaiters.addFirst(waiter)
+  waiters.delete(0, i - 1)
```
cheatfate, line 206: It looks like you do not like to keep an order here, what is actually important.

arnetheduck: `delete` keeps order

chronos/asyncsync.nim
arnetheduck: it's no longer `async` so cancellations don't apply (this way we avoid a roundtrip to the event loop and all the extra cancellation future code)

chronos/asyncsync.nim
arnetheduck: the significant fix here is that we remove finished futures even if there is no finished future (if it was cancelled for example) - this allows the gc to recycle it earlier.

PR #129, opened 2020-09-10
Fix AsyncLock.locked flag to be consistent.

cheatfate:
Refactor `AsyncLock` to not use `result`.
Add test for `locked` flag.

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -9,8 +9,8 @@
 #                MIT license (LICENSE-MIT)
 
 ## This module implements some core synchronization primitives
-import std/sequtils
-import asyncloop, deques
+import std/[sequtils, deques]
+import ./asyncloop
 
 type
   AsyncLock* = ref object of RootRef
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -71,24 +71,22 @@ proc newAsyncLock*(): AsyncLock =
   # Workaround for callSoon() not worked correctly before
   # getGlobalDispatcher() call.
   discard getGlobalDispatcher()
-  result = new AsyncLock
-  result.waiters = newSeq[Future[void]]()
-  result.locked = false
+  AsyncLock(waiters: newSeq[Future[void]](), locked: false)
 
-proc wakeUpFirst(lock: AsyncLock) {.inline.} =
+proc wakeUpFirst(lock: AsyncLock): bool {.inline.} =
   ## Wake up the first waiter if it isn't done.
   for fut in lock.waiters.mitems():
     if not(fut.finished()):
       fut.complete()
-      break
+      return true
+  return false
 
 proc checkAll(lock: AsyncLock): bool {.inline.} =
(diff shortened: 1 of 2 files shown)


chronos/asyncsync.nim
```diff
@@ -70,24 +70,22 @@ proc newAsyncLock*(): AsyncLock =
...
+  AsyncLock(waiters: newSeq[Future[void]](), locked: false)
 
-proc wakeUpFirst(lock: AsyncLock) {.inline.} =
+proc wakeUpFirst(lock: AsyncLock): bool {.inline.} =
   ## Wake up the first waiter if it isn't done.
   for fut in lock.waiters.mitems():
     if not(fut.finished()):
       fut.complete()
-      break
+      return true
```
arnetheduck, line 80: best would be to remove the finished futures from the waiters list here since they no longer serve any purpose in the list

chronos/asyncsync.nim
```diff
@@ -125,8 +118,8 @@ proc release*(lock: AsyncLock) =
   ## other coroutines are blocked waiting for the lock to become unlocked,
   ## allow exactly one of them to proceed.
   if lock.locked:
-    lock.locked = false
-    lock.wakeUpFirst()
+    if not(lock.wakeUpFirst()):
```
arnetheduck, line 121: I think add a comment here explaining the logic why we don't set locked to false - in the new flow, `locked = true` will still run in `acquire` which on a casual reading of the code looks "unbalanced"

tests/testsync.nim
```diff
@@ -41,6 +41,75 @@ suite "Asynchronous sync primitives test suite":
...
+      res = true
+    return res
+
+  proc testDoubleRelease(): Future[bool] {.async.} =
+    var lock = newAsyncLock()
+    var fut0 = lock.acquire()
+    var fut1 = lock.acquire()
+    var res = false
+    asyncSpawn fut0
+    asyncSpawn fut1
```
arnetheduck, line 105: this depends on `asyncSpawn` not yielding control - is that the intent? is that a guarantee that will be maintained?

chronos/asyncsync.nim
```diff
@@ -71,53 +72,48 @@ proc newAsyncLock*(): AsyncLock =
   # Workaround for callSoon() not worked correctly before
   # getGlobalDispatcher() call.
   discard getGlobalDispatcher()
-  result = new AsyncLock
-  result.waiters = newSeq[Future[void]]()
-  result.locked = false
+  AsyncLock(waiters: newSeq[Future[void]](), locked: false, released: true)
```
arnetheduck, line 75: ideally we'd have a `released` flag that starts with `false`  / the default value for bool - that leads to less bugs generally in nim, would have to change name

tests/testsync.nim
cheatfate: `asyncSpawn` here is to raise `Defect` on if `fut0` or `fut1` will be cancelled or failed.

arnetheduck (approved):

PR #134, opened 2020-11-11
Small fixes

cheatfate:


chronos/asyncloop.nim
```diff
@@ -803,6 +805,31 @@ proc sleepAsync*(ms: int): Future[void] {.
      inline, deprecated: "Use sleepAsync(Duration)".} =
   result = sleepAsync(ms.milliseconds())
 
+proc stepsAsync*(number: int): Future[void] =
```
arnetheduck, line 808: aka `yieldAsync`?

cheatfate: Actually not, because `yieldAsync` should not introduce 100% CPU consumption, this procedure could easily achieve it.

cheatfate: So the purpose of this primitive is to help creating complex tests...

arnetheduck: ok, that could be documented here - also, there's no such thing as "for testing", if it's there, it should be generally supported - ie this would be useful for pausing the queue processing in nbc and letting networking do its work

cheatfate: At any case this primitive MUST not be used for pausing the queue processing and let networking do its work.

chronos/asyncfutures2.nim
```diff
@@ -761,22 +788,24 @@ proc oneValue*[T](futs: varargs[Future[T]]): Future[T] {.
...
   proc continuation(udata: pointer) {.gcsafe.} =
     if not(retFuture.finished()):
       retFuture.complete()
-  fut.addCallback(continuation)
-
-  # Start the cancellation process. If `fut` has children, multiple event loop
-  # steps will be needed for it to complete.
-  fut.cancel()
-
+  proc cancellation(udata: pointer) {.gcsafe.} =
```
arnetheduck, line 800: @cheatfate unused

PR #145, opened 2021-01-11
Fix callSoon to raise Defect only.

cheatfate:


chronos/asyncloop.nim
```diff
@@ -382,17 +383,12 @@ when defined(windows) or defined(nimdoc):
...
-    if not gDisp.isNil:
-      doAssert gDisp.callbacks.len == 0
-    gDisp = disp
-
-  proc getGlobalDispatcher*(): PDispatcher =
-    ## Returns current thread's dispatcher instance.
-    if gDisp.isNil:
-      setGlobalDispatcher(newDispatcher())
-    result = gDisp
+  proc setThreadDispatcher*(disp: PDispatcher) {.gcsafe.}
```
arnetheduck, line 386: these forward declarations should move to the non-platform specific part?

cheatfate: But `PDispatcher` object is platform specific.

arnetheduck (approved):

PR #149, opened 2021-01-20
Refactoring of AsyncStreams.

cheatfate:
1. Implement all read() primitives using readLoop() like it was done in streams.
2. Fix readLine() bug.
3. Add readMessage() primitive.
4. Fixing exception hierarchy, handling code and simplification of (break/continue + exception).
5. Fix TLSStream closure procedure.
6. Add BoundedStream stream and tests.
7. Remove `result` usage from the code.
8. Add `bytesCount` argument to catch number of bytes written or read by stream.
9. Optimize ChunkedStream hexadecimal decoding.
10. Remove usage of `awaitne` primitive.
11. Avoid `CatchableError` exceptions in read/write loops.


cheatfate:
Closing in favor of #153

PR #154, opened 2021-02-04
export selector and handles fields from PDispatcher to support third-party extensions

michaelsbradleyjr:
While working on the `nim-task-runner` repo, which imports `nim-chronos`, it was found that the `selector` and `handles` fields on `PDispatcher` needed to be exported so that objects returned from `getThreadDispatcher()` can have those fields accessed.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -310,7 +310,7 @@ when defined(windows) or defined(nimdoc):
 
     PDispatcher* = ref object of PDispatcherBase
       ioPort: Handle
-      handles: HashSet[AsyncFD]
+      handles*: HashSet[AsyncFD]
       connectEx*: WSAPROC_CONNECTEX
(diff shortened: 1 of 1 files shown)


cheatfate:
I understand that you need something to work, but explanation is also needs to be more comprehensive. Because currently i dont see any reasons to accept your PR, because it could lead to crashes and undefined behavior if i allow third-party extensions to modify this data structures.

jakubgs:
Based on [comment](https://github.com/status-im/status-desktop/pull/7786#issuecomment-1280801982) from @iurimatias the Desktop team has no intention to upgrade `nimbus-build-system`, so this might not become an issue in the long term, but the fact remains that `status-desktop` does use two versions of `nim-chronos` in their submodules.

arnetheduck:
replaced with #363 which is the actual version used right now

PR #135, opened 2020-11-15
ORC build (experiment)

sinkingsugar:
Removed the old incompatible code for now, but maybe @Araq has a better solution.
All tests almost pass, but it's failing in asyncstream.. I think it's bad codegen by looking at `C` code.
```c
nimln_(801, "E:\\develop\\status\\nim-chronos\\chronos\\streams\\asyncstream.nim");
nimln_(335, "E:\\develop\\status\\nim-chronos\\chronos\\asyncfutures2.nim");
nimln_(801, "E:\\develop\\status\\nim-chronos\\chronos\\streams\\asyncstream.nim");
colontmpD_ = 0;
nimln_(240, "E:\\develop\\status\\nim-chronos\\chronos\\asyncfutures2.nim");
eqcopy___WVHzlBztR3durWxAW3awcg_2(&colontmpD_, (*(*colonenv_).rw1).Sup.future);
if (NIM_UNLIKELY(*nimErr_)) goto LA1_;
T27_ = (tyObject_FutureBasecolonObjectType___V9aHQccB2Pyjfqk9bdE5ZRnw*)0;
T27_ = &colontmpD_->Sup;
nimln_(335, "E:\\develop\\status\\nim-chronos\\chronos\\asyncfutures2.nim");
T28_ = (tyObject_SrcLoc__9cugm8iqHlSb5d0xCTLlcbg*)0;
T28_ = srcLocImpl__9aKv9c9bPL5yWoiTwJ0le1uqQ();
if (NIM_UNLIKELY(*nimErr_)) goto LA1_;
T29_ = (NIM_BOOL)0;
T29_ = cancel__ZYuu0FydmjOOjOhwx69cq1w(T27_, T28_);
if (NIM_UNLIKELY(*nimErr_)) goto LA1_;
(void)(T29_);
```
`colontmpD_ = 0;` -> `eqcopy___WVHzlBztR3durWxAW3awcg_2(&colontmpD_, (*(*colonenv_).rw1).Sup.future);` 
is not really correct I think.

PR #161, opened 2021-02-26
Enable --gc:orc testing for nim-chronos

Araq:
Don't forget to "squash" my commits.


sinkingsugar:
Tests are green, maybe this can be closed as well https://github.com/status-im/nim-chronos/pull/135

sinkingsugar:
Curious about perf btw, some many months ago orc builds were quite a perf hit on reference heavy code due to cycles.
Of course the way I wrote that code was with the GC in mind so the blame is not really on the ORC.
Yet it's a paradigm shift, old codebases might need big refactoring.
Chronos uses lot of cycling references afaik.

Araq:
In my experience raw throughput is worse (by about 20%) but latency is better by orders of magnitude because there is no "stop the world" phase and WCET can be measured effectively (it's very predictable). Memory consumption can be reduced by a factor of 2. You can also tweak Orc with the new annotations in ways that were previously impossible -- I never look back once a project works with Orc.

sinkingsugar:
> In my experience raw throughput is worse (by about 20%) but latency is better by orders of magnitude because there is no "stop the world" phase and WCET can be measured effectively (it's very predictable). Memory consumption can be reduced by a factor of 2. You can also tweak Orc with the new annotations in ways that were previously impossible -- I never look back once a project works with Orc.

Yeah, that's for sure.
But my point is that the old GC tricked us into very bad coding habits often, way too many refs and cycles. With ORC and most of all destructors/constructors value types are to be preferred when possible.

arnetheduck (approved):

PR #165, opened 2021-03-12
Split HTTPS and HTTP servers.

cheatfate:
This PR splits HTTP and HTTPS server implementations, so it become possible to use HTTP server without SSL library dependency.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -64,6 +64,7 @@ type
     acceptLoop*: Future[void]
     lifetime*: Future[void]
     headersTimeout: Duration
+    bufferSize: int
     maxHeadersSize: int
     maxRequestBodySize: int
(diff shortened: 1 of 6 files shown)


chronos/apps/http/httpserver.nim
```diff
@@ -450,47 +469,30 @@ proc getRequest(conn: HttpConnectionRef): Future[HttpRequestRef] {.async.} =
   except AsyncStreamLimitError:
     raiseHttpCriticalError("Maximum size of request headers reached", Http431)
 
+proc init*(value: var HttpConnection, server: HttpServerRef,
+           transp: StreamTransport) =
+  value.server = server
```
arnetheduck, line 474: a way to write an `init` and guarantee that all fields are reset is `value = HttpConnection(a: b, ...)`, if that is the intent

arnetheduck (approved):

PR #166, opened 2021-03-14
exception tracking

arnetheduck:
This PR adds minimal exception tracking to chronos, moving the goalpost
one step further.

In particular, it becomes invalid to raise exceptions from `callSoon`
callbacks: this is critical for writing correct error handling because
there's no reasonable way that a user of chronos can possibly _reason_
about exceptions coming out of there: the event loop will be in an
indeterminite state when the loop is executing an _random_ callback.

As expected, there are several issues in the error handling of chronos:
in particular, it will end up in an inconsistent internal state whenever
the selector loop operations fail, because the internal state update
functions are not written in an exception-safe way. This PR turns this
into a Defect, which probably is not the optimal way of handling things
- expect more work to be done here.

Some API have no way of reporting back errors to callers - for example,
when something fails in the accept loop, there's not much it can do, and
no way to report it back to the user of the API - this has been fixed
with the new accept flow - the old one should be deprecated.

Finally, there is information loss in the API: in composite operations
like `poll` and `waitFor` there's no way to differentiate internal
errors from user-level errors originating from callbacks.

* store `CatchableError` in future
* annotate proc's with correct raises information
* `selectors2` to avoid non-CatchableError IOSelectorsException
* `$` should never raise
* remove unnecessary gcsafe annotations
* fix exceptions leaking out of timer waits
* fix some imports
* functions must signal raising the union of all exceptions across all
platforms to enable cross-platform code
* switch to unittest2
* add `selectors2` which supercedes the std library version and fixes
several exception handling issues in there

An optional strict exception mode is also introduced, to be used when code correctly declares what exceptions it's raising - it is enabled with `-d:chronosStrictException`.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -44,11 +44,11 @@ type
     Empty, Prepared, Sending, Finished, Failed, Cancelled, Dumb
 
   HttpProcessCallback* =
-    proc(req: RequestFence): Future[HttpResponseRef] {.gcsafe.}
+    proc(req: RequestFence): Future[HttpResponseRef] {.gcsafe, raises: [Defect].}
 
   HttpConnectionCallback* =
     proc(server: HttpServerRef,
-         transp: StreamTransport): Future[HttpConnectionRef] {.gcsafe.}
+         transp: StreamTransport): Future[HttpConnectionRef] {.gcsafe, raises: [Defect].}
 
   HttpServer* = object of RootObj
     instance*: StreamServer
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -44,7 +44,8 @@ type
     Empty, Prepared, Sending, Finished, Failed, Cancelled, Dumb
 
   HttpProcessCallback* =
-    proc(req: RequestFence): Future[HttpResponseRef] {.gcsafe, raises: [Defect].}
+    proc(req: RequestFence): Future[HttpResponseRef] {.
+      gcsafe, raises: [Defect, CatchableError].}
 
   HttpConnectionCallback* =
     proc(server: HttpServerRef,
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -506,6 +506,19 @@ proc mget*[T](future: FutureVar[T]): var T =
   ## Future has not been finished.
   result = Future[T](future).value
 
+template taskFutureLocation(future: FutureBase): string =
+  let loc = future.location[0]
+  "[" & (
+    if len(loc.procedure) == 0: "[unspecified]" else: $loc.procedure & "()"
+    ) & " at " & $loc.file & ":" & $(loc.line) & "]"
+
+template taskErrorMessage(future: FutureBase): string =
+  "Asynchronous task " & taskFutureLocation(future) &
+  " finished with an exception \"" & $future.error.name & "\"!\nStack trace: " &
+  future.error.getStackTrace()
+template taskCancelMessage(future: FutureBase): string =
+  "Asynchronous task " & taskFutureLocation(future) & " was cancelled!"
+
 proc asyncSpawn*(future: Future[void]) =
   ## Spawns a new concurrent async task.
   ##
@@ -520,32 +533,17 @@ proc asyncSpawn*(future: Future[void]) =
   ## and processed immediately.
   doAssert(not isNil(future), "Future is nil")
 
-  template getFutureLocation(): string =
-    let loc = future.location[0]
-    "[" & (
-      if len(loc.procedure) == 0: "[unspecified]" else: $loc.procedure & "()"
-      ) & " at " & $loc.file & ":" & $(loc.line) & "]"
-
-  template getErrorMessage(): string =
-    "Asynchronous task " & getFutureLocation() &
-    " finished with an exception \"" & $future.error.name & "\"!"
-  template getCancelMessage(): string =
-    "Asynchronous task " & getFutureLocation() & " was cancelled!"
-
   proc cb(data: pointer) =
     if future.failed():
-      raise newException(FutureDefect, getErrorMessage())
+      raise newException(FutureDefect, taskErrorMessage(future))
     elif future.cancelled():
-      raise newException(FutureDefect, getCancelMessage())
+      raise newException(FutureDefect, taskCancelMessage(future))
 
   if not(future.finished()):
     # We adding completion callback only if ``future`` is not finished yet.
     future.addCallback(cb)
   else:
-    if future.failed():
-      raise newException(FutureDefect, getErrorMessage())
-    elif future.cancelled():
-      raise newException(FutureDefect, getCancelMessage())
+    cb(nil)
 
 proc asyncCheck*[T](future: Future[T]) {.
     deprecated: "Raises Defect on future failure, fix your code and use asyncSpawn!".} =
@@ -553,7 +551,20 @@ proc asyncCheck*[T](future: Future[T]) {.
   ## the given future failed - there's no way to handle such exceptions so this
   ## function is now an alias for `asyncSpawn`
   ##
-  asyncSpawn(future)
+  when T is void:
+    asyncSpawn(future)
+  else:
+    proc cb(data: pointer) =
+      if future.failed():
+        raise newException(FutureDefect, taskErrorMessage(future))
+      elif future.cancelled():
+        raise newException(FutureDefect, taskCancelMessage(future))
+
+    if not(future.finished()):
+      # We adding completion callback only if ``future`` is not finished yet.
+      future.addCallback(cb)
+    else:
+      cb(nil)
 
 proc asyncDiscard*[T](future: Future[T]) {.
     deprecated: "Use asyncSpawn or `discard await`".} = discard
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -706,7 +706,7 @@ proc all*[T](futs: varargs[Future[T]]): auto {.
               retFuture.fail(nfut.error)
               break
             else:
-              retValues[k] = nfut.read()
+              retValues[k] = nfut.value
           if not(retFuture.failed()):
             retFuture.complete(retValues)
 
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -9,7 +9,7 @@
 #                MIT license (LICENSE-MIT)
 
 import std/[os, tables, strutils, heapqueue, options, deques, cstrutils, sequtils]
-import srcloc
+import ./srcloc
 export srcloc
 
 const
@@ -29,7 +29,7 @@ type
     cancelcb*: CallbackFunc
     child*: FutureBase
     state*: FutureState
-    error*: ref Exception ## Stored exception
+    error*: ref CatchableError ## Stored exception
     mustCancel*: bool
     id*: int
 
@@ -171,7 +171,7 @@ proc done*(future: FutureBase): bool {.inline.} =
   completed(future)
 
 when defined(chronosFutureTracking):
-  proc futureDestructor(udata: pointer) {.gcsafe.} =
+  proc futureDestructor(udata: pointer) =
     ## This procedure will be called when Future[T] got finished, cancelled or
     ## failed and all Future[T].callbacks are already scheduled and processed.
     let future = cast[FutureBase](udata)
@@ -271,7 +271,7 @@ template complete*[T](futvar: FutureVar[T], val: T) =
   ## Any previously stored value will be overwritten.
   complete(futvar, val, getSrcLocation())
 
-proc fail[T](future: Future[T], error: ref Exception, loc: ptr SrcLoc) =
+proc fail[T](future: Future[T], error: ref CatchableError, loc: ptr SrcLoc) =
   if not(future.cancelled()):
     checkFinished(FutureBase(future), loc)
     future.error = error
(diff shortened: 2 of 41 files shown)


chronos/asyncmacro2.nim
```diff
@@ -304,7 +316,8 @@ template await*[T](f: Future[T]): auto =
     if chronosInternalRetFuture.mustCancel:
       raise newCancelledError()
     chronosInternalTmpFuture.internalCheckComplete()
-    cast[type(f)](chronosInternalTmpFuture).internalRead()
+    when T isnot void:
```
cheatfate, line 319: `Future[void]` still can has an exception inside, and in such case you swallow all the exceptions inside failed `Future[void]`.

chronos/transports/datagram.nim
```diff
@@ -601,7 +631,12 @@ proc close*(transp: DatagramTransport) =
   else:
     if {ReadClosed, WriteClosed} * transp.state == {}:
       transp.state.incl({WriteClosed, ReadClosed})
-      closeSocket(transp.fd, continuation)
+      try:
+        closeSocket(transp.fd, continuation)
+      except IOSelectorsException as exc:
+        raiseAssert exc.msg
```
cheatfate, line 637: This is absolutely possible that `close` could raise, and it should not be `Defect`.

chronos/transports/stream.nim
```diff
@@ -700,12 +703,20 @@ when defined(windows):
...
 
-      var raddress = windowsAnyAddressFix(address)
+      var raddress = try: windowsAnyAddressFix(address)
+      except TransportAddressError as exc:
+        raiseAssert exc.msg
 
       toSAddr(raddress, saddr, slen)
       proto = Protocol.IPPROTO_TCP
-      sock = createAsyncSocket(raddress.getDomain(), SockType.SOCK_STREAM,
+      sock = try: createAsyncSocket(raddress.getDomain(), SockType.SOCK_STREAM,
```
cheatfate, line 712: This procedure do not raise any exceptions except `OSError`.

arnetheduck: it depends on the platform you're compiling for: on windows it raises `OSError`, on linux it raises `IOSelectorException` and `ValueError` - this  is because it calls `register` which behaves differently

chronos/transports/datagram.nim
arnetheduck: well, this is the tricky part: `closeSocket` will raise only if the internal state of the ioselector is inconsistent - `closeSocket` raises when an invalid `fd` is passed to it - later, it raises an unrecoverable error in `unregister` as well, but that happens in a continuation which can no longer be caught _here_

chronos/asyncmacro2.nim
arnetheduck: `internalCheckComplete` checks for the error, `internalRead` just copies the value (it would possibly make sense to make `internalRead` a template such that a copy can be avoided, but `lent` serves a similar purpose in future Nim versions

chronos/transports/datagram.nim
cheatfate: `closeSocket` could raise `EBADF`, `EINTR` (not handled right now), `EIO` (i'm talking about it).

chronos/asyncfutures2.nim
```diff
@@ -611,7 +611,8 @@ proc `or`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
   ##
   ## If cancelled, ``fut1`` and ``fut2`` futures WILL NOT BE cancelled.
   var retFuture = newFuture[void]("chronos.or")
-  proc cb(udata: pointer) {.gcsafe.} =
+  var cb: proc(udata: pointer) {.gcsafe, raises: [Defect].}
+  cb = proc(udata: pointer) =
```
dryajov, line 615:
```suggestion
  proc cb(udata: pointer) {.gcsafe, raises: [Defect].} =
```

arnetheduck: https://github.com/nim-lang/Nim/issues/17369

chronos/asyncmacro2.nim
```diff
@@ -247,6 +246,19 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
     closureIterator.pragma = newNimNode(nnkPragma, lineInfoFrom=prc.body)
     closureIterator.addPragma(newIdentNode("closure"))
 
+    # TODO when push raises is active in a module, the iterator here inherits
```
dryajov, line 249: It should probably inherit the exceptions from the `{.async.}` proc? Should be a matter of reading them from the annotated proc and attaching to the iterator?

arnetheduck: it needs to either a) deduce exceptions which it can't do easily, or b) use explicit annotations which come with their own issues - once it knows what exceptions are raised, it needs to rewrite all `await` calls with some additional macro logic - this is far beyond this PR to solve.

dryajov: Fair, I think (a) is probably closer to what is needed and it should be possible to read annotations from the wrapped proc, this will happen on macro expansion at compile time so the original exception info should still be present on the wrapped proc/method?

arnetheduck: the issue is `await` that in its present form always causes a `CatchableError` raises effect because it reads from the future which doesn't hold specific exception type - one would need to transport the error type more precisely in some shape or form - there are a couple proposals for that - the main point of _this_ branch is to get rid of `Exception` effects that leak out and also address the `poll` exception leaks once and for all.

chronos/transports/datagram.nim
arnetheduck: going through the code, it turns closeSocket doesn't raise in its present form but perhaps should - that's outside the scope of this PR however

chronos/asyncmacro2.nim
dryajov: OK, what I mean is that the _desirable_ option is probably (a) and it's fine we address it as part of another PR.

chronos/apps/http/httpserver.nim
```diff
@@ -44,11 +44,11 @@ type
     Empty, Prepared, Sending, Finished, Failed, Cancelled, Dumb
 
   HttpProcessCallback* =
-    proc(req: RequestFence): Future[HttpResponseRef] {.gcsafe.}
+    proc(req: RequestFence): Future[HttpResponseRef] {.gcsafe, raises: [Defect].}
```
cheatfate, line 47: This is not needed annotation because all the exceptions handled here https://github.com/status-im/nim-chronos/blob/master/chronos/apps/http/httpserver.nim#L589-L598

chronos/asyncfutures2.nim
```diff
@@ -611,7 +609,8 @@ proc `or`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
   ##
   ## If cancelled, ``fut1`` and ``fut2`` futures WILL NOT BE cancelled.
   var retFuture = newFuture[void]("chronos.or")
-  proc cb(udata: pointer) {.gcsafe.} =
+  var cb: proc(udata: pointer) {.gcsafe, raises: [Defect].}
```
cheatfate, line 612: Why this continuation callback needs explicit `raises: [Defect]`. Why not mark every possible continuation callback as `raises:[Defect]`?

chronos/asyncloop.nim
```diff
@@ -1027,11 +1019,15 @@ proc wait*[T](fut: Future[T], timeout = InfiniteDuration): Future[T] =
             when T is void:
               retFuture.complete()
             else:
-              retFuture.complete(fut.read())
+              try:
+                retFuture.complete(fut.read())
+              except CatchableError as exc:
+                retFuture.fail(exc)
```
cheatfate, line 1025: This should be `Defect`. Because this situation is not possible. There present check at https://github.com/status-im/nim-chronos/pull/166/files#diff-95090cf0d6917b2c1eae2c05fe84b275d0f77b28ac5f0638651bd2d672e756f8R1016 which protects reading `fut` value from raising exception.

chronos/asyncloop.nim
```diff
@@ -1045,7 +1041,10 @@ proc wait*[T](fut: Future[T], timeout = InfiniteDuration): Future[T] =
       when T is void:
         retFuture.complete()
       else:
-        retFuture.complete(fut.read())
+        try:
+          retFuture.complete(fut.read())
+        except CatchableError as exc:
+          retFuture.fail(exc)
```
cheatfate, line 1047: This should be `Assertion` or `Defect` error, because situation that `fut.read()` will raise is impossible.

chronos/asyncloop.nim
```diff
@@ -1069,15 +1068,21 @@ proc wait*[T](fut: Future[T], timeout = -1): Future[T] {.
...
     wait(fut, timeout.milliseconds())
 
+{.pop.}
+
 include asyncmacro2
 
-proc runForever*() =
+{.push raises: [Defect].}
+
+proc runForever*() {.raises: [Defect, CatchableError].} =
```
cheatfate, line 1077: I think this procedure should not leak `CatchableError`.

chronos/asyncloop.nim
```diff
@@ -1069,15 +1068,21 @@ proc wait*[T](fut: Future[T], timeout = -1): Future[T] {.
...
   ## Begins a never ending global dispatcher poll loop.
+  ## Raises different exceptions depending on the platform.
   while true:
     poll()
 
-proc waitFor*[T](fut: Future[T]): T =
+proc waitFor*[T](fut: Future[T]): T {.raises: [Defect, CatchableError].} =
```
cheatfate, line 1083: I think this procedure should not leak `CatchableError`.

chronos/transports/stream.nim
```diff
@@ -864,10 +879,12 @@ when defined(windows):
                                            DWORD(server.bufferSize),
                                            DWORD(0), nil)
           if pipeHandle == INVALID_HANDLE_VALUE:
-            raiseTransportOsError(osLastError())
+            raiseAssert $osLastError()
```
cheatfate, line 882: `osErrorMsg(osLastError())` but not `$`.

chronos/transports/stream.nim
```diff
@@ -1072,7 +1092,12 @@ when defined(windows):
                                              nil, flags)
           # Start tracking transport
           trackStream(ntransp)
-          server.createAcceptPipe()
+          try:
+            server.createAcceptPipe()
```
cheatfate, line 1096: trackStream(ntransp) should be inside try/except branch, so only successful attempts will be tracked.

chronos/transports/stream.nim
```diff
@@ -1072,7 +1092,12 @@ when defined(windows):
...
+          except CatchableError as exc:
+            # TODO close transp!
```
cheatfate, line 1098: closeHandle(sock) ?

chronos/asyncfutures2.nim
arnetheduck: nim bug: https://github.com/nim-lang/Nim/issues/17369

chronos/asyncloop.nim
arnetheduck: it raises the `CatchableError` that is potentially inside the future - it would have to return the future (or a `Result`) in order to only raise when `poll` raises

chronos/apps/http/httpserver.nim
arnetheduck: fixed

chronos/transports/stream.nim
arnetheduck: moved it after the try/catch which should have that same effect?

chronos/transports/stream.nim
arnetheduck: ok, pushed something, but take a look - the flow here is quite hard to follow

chronos/asyncmacro2.nim
```diff
@@ -247,13 +290,23 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+    when defined(chronosStrictException):
+      closureIterator.addPragma(nnkExprColonExpr.newTree(
+        newIdentNode("raises"),
+        nnkBracket.newTree(
+          newIdentNode("Defect"),
+          newIdentNode("CatchableError")
+        )
+      ))
+    else:
+      closureIterator.addPragma(nnkExprColonExpr.newTree(
```
dryajov, line 302: Doesn't `Exception` make all `Defect` and `CatchableError` reduntant?

arnetheduck: sort of, but it also don't hurt - ie it's easier to see what the hack is this way.

chronos/asyncfutures2.nim
```diff
@@ -561,8 +547,20 @@ proc asyncSpawn*(future: Future[void]) =
...
 
-proc asyncDiscard*[T](future: Future[T]) {.deprecated.} = discard
-  ## This is async workaround for discard ``Future[T]``.
+proc asyncCheck*[T](future: Future[T]) {.
+    deprecated: "Raises Defect on future failure, fix your code and use asyncSpawn!".} =
+  ## This function used to raise an exception through the `poll` call if
+  ## the given future failed - there's no way to handle such exceptions so this
+  ## function is now an alias for `asyncSpawn`
+  ##
+  asyncSpawn(future)
```
kdeme, line 556: Won't there be some `Future[void]` versus `Future[T]` incompatiblities here? 
I noticed this error at least in nim-eth's (old) kademlia when testing by running the p2p tests.

arnetheduck: yes, I have a fix locally for that which discards the T

arnetheduck: pushed

tests/testsoon.nim
```diff
@@ -46,9 +46,11 @@ suite "callSoon() tests suite":
       await sleepAsync(100.milliseconds)
       timeoutsTest1 += 1
 
-  proc callbackProc(udata: pointer) {.gcsafe.} =
+  var callbackproc: proc(udata: pointer) {.gcsafe, raises: [Defect].}
+  callbackproc = proc (udata: pointer) {.gcsafe, raises: [Defect].} =
     timeoutsTest2 += 1
-    callSoon(callbackProc)
+    {.gcsafe.}:
```
zah, line 52: Why is this override needed here?

arnetheduck: probably it isn't with `unittest2`, but before, callbackproc was a global - I'll see if it can be removed

PR #168, opened 2021-03-24
Add raises annotation to withTimeout callbacks

kdeme:
I was updating vendor modules in nim-waku and noticed withTimeout usage would still give issues.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -951,7 +951,9 @@ proc withTimeout*[T](fut: Future[T], timeout: Duration): Future[bool] =
   var timer: TimerCallback
   var cancelling = false
 
-  proc continuation(udata: pointer) {.gcsafe.} =
+  # TODO: raises annotation shouldn't be needed, but likely similar issue as
+  # https://github.com/nim-lang/Nim/issues/17369
+  proc continuation(udata: pointer) {.gcsafe, raises: [Defect].} =
     if not(retFuture.finished()):
       if not(cancelling):
         if not(fut.finished()):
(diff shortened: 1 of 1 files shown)


chronos/asyncloop.nim
```diff
@@ -951,7 +951,7 @@ proc withTimeout*[T](fut: Future[T], timeout: Duration): Future[bool] =
   var timer: TimerCallback
   var cancelling = false
 
-  proc continuation(udata: pointer) {.gcsafe.} =
+  proc continuation(udata: pointer) {.gcsafe, raises: [Defect].} =
```
arnetheduck, line 954: this shouldn't be needed - in fact, `{.gcsafe.}` should be removed as well (the compiler will deducee it) - are you actually seeing an error?

kdeme: Yes, compiling nim-waku is giving this error after bumping nim-chronos, nim-libp2p (and nim-eth):

```
/home/deme/repos/nim-waku/waku/v2/node/peer_manager.nim(140, 69) template/generic instantiation of `async` from here
/home/deme/repos/nim-waku/waku/v2/node/peer_manager.nim(157, 22) template/generic instantiation of `withTimeout` from here
/home/deme/repos/nim-waku/vendor/nim-chronos/chronos/asyncloop.nim(974, 10) Error: type mismatch: got <Future[connection.Connection], proc (udata: pointer){.closure, gcsafe, locks: <unknown>.}>
but expected one of: 
proc removeCallback(future: FutureBase; cb: CallbackFunc; udata: pointer = nil)
  first type mismatch at position: 2
  required type for cb: CallbackFunc
  but expression 'continuation' is of type: proc (udata: pointer){.closure, gcsafe, locks: <unknown>.}
  The `.raises` requirements differ.
proc removeCallback[T](future: Future[T]; cb: CallbackFunc)
  first type mismatch at position: 2
  required type for cb: CallbackFunc
  but expression 'continuation' is of type: proc (udata: pointer){.closure, gcsafe, locks: <unknown>.}
  The `.raises` requirements differ.

expression: removeCallback(fut, continuation)
```

https://github.com/status-im/nim-waku/blob/master/waku/v2/node/peer_manager.nim#L156

arnetheduck:
let's add a TODO link to https://github.com/nim-lang/Nim/issues/17369 - I suspect it's the same issue, then use this workaround for now

arnetheduck (approved):

PR #171, opened 2021-03-29
please Nim 1.4+ compiler

narimiran:
`%`, used in `error("Expected return type of 'Future' got '$1'" % typeName)`,  can raise `ValueError`.
(Just adding `raises: [ValueError]` pragma moves the problem around, and doesn't solve it.)


arnetheduck:
FYI the problem is most likely not the code but rather a bug in Nim - `verifyReturnType` is not covered by any `raises` constraints so there should be no error - for these cases, we open an issue in Nim and link it to the workaround with a todo comment like so: https://github.com/status-im/nimbus-eth2/blob/1af07ad1247b43bd79e775b7b466e767ec3f0ecf/beacon_chain/networking/eth2_network.nim#L491

PR #172, opened 2021-03-30
Ensures that asyncmacro2 does not raise exceptions

markspanbroek:
Changed all places where asyncmacro2 might raise an exception:
- calls to `macros.$` are replaced by calls to `strVal`
- `%` formatting is replaced by string concatenation
- call to `parseExpr` (which might raise `ValueError`) is replaced by `newNimNode`

These changes also ensure that chronos works with Nim 1.4.4.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -1073,11 +1073,7 @@ proc wait*[T](fut: Future[T], timeout = -1): Future[T] {.
   else:
     wait(fut, timeout.milliseconds())
 
-{.pop.}
-
 include asyncmacro2
-
-{.push raises: [Defect].}
 
 proc runForever*() {.raises: [Defect, CatchableError].} =
   ## Begins a never ending global dispatcher poll loop.
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -10,8 +10,6 @@
 
 {.push raises: [Defect].}
 
-include "system/inclrtl"
-
 import std/[os, tables, strutils, heapqueue, lists, options, nativesockets, net,
         deques]
 import ./timer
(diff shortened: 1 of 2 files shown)


arnetheduck:
This is likely a bug in Nim that's being investigated - the code is not covered by `push raises` so should be allowed to raise as it pleases - before we merge this, report upstream with some repro instructions so their investigation can continue. It affects 1.2 also.

See also https://github.com/status-im/nim-chronos/pull/171

markspanbroek:
Yes, I'll work on a reproduction scenario and report upstream.

markspanbroek:
I can reproduce it as follows:
```nim
{.push raises: [Defect].}

include "system/inclrtl"

{.pop.}

proc foo* =
  raise newException(CatchableError, "boo!")

{.push raises: [Defect].}
```

The include of `system/inclrtl` pushes [sinkInference](https://github.com/nim-lang/Nim/blob/af2a7e8cbc3729ac62be64b5683e4a2187160b2f/lib/system/inclrtl.nim#L50), which is what gets popped instead of what you might expect.

I could argue both ways: it could be considered a bug in Nim, or in chronos. What do you think @arnetheduck , should I report this upstream?

dryajov:
I thought that push pragmas are module level? Why is the push from inclrtl escaping it's module?

arnetheduck:
`include` includes the text of the file, doesn't create a separate module

markspanbroek:
Upstream issue: https://github.com/nim-lang/Nim/issues/17583

markspanbroek:
Removed the include, all tests pass.

dryajov:
Ah, right missed the include part, makes sense then.

PR #194, opened 2021-05-27
add shortcuts to create completed futures

arnetheduck:


chronos/asyncfutures2.nim
```diff
@@ -243,6 +243,12 @@ template complete*[T](future: Future[T], val: T) =
   ## Completes ``future`` with value ``val``.
   complete(future, val, getSrcLocation())
 
+template complete*[T](F: type Future[T], val: T, fromProc: static[string] = ""): F =
```
cheatfate, line 246: I see this named confusing, and this could introduce some sandwich problem in the future, i think such templates should have different name, like `completed` for example.

cheatfate:
What the main goal of this PR? Right now i see it very subjective.

arnetheduck:
there's a few places in nbc where we create completed or failed futures, and this is a convenient way to do it - it's similar to how you create a `Result[A,B].ok(value)` - instead of first creating an unfinished future, you can use these constructors with `let` and other features that make code less risky - ie `if a: Future[T].complete(x) else: Future[T].fail(b)` for example.

Yes, to a certain extent, it's subjective, but it's aimed at using the safety features of the language better. `completed` is taken already, to check if the future is complete. parameter overloading like this should be safe, everything is in the same module so we shouldn't have sandwitch problems. `Result` looks like this all the way and doesn't have issues.

The test updates on the other hand are a consequence of `unittest2` already putting each test in a separate `proc` - no need to do it manually any more which makes the tests slightly easier to understand: the `check`s can now give more detailed errors instead of values like "300" that don't mean anything.

cheatfate:
Manual `Future[T]` management is unsafe by design.

All this `Future[T].addCallback()` and closure procedures do not make code simpler. Right now `chronos` allows both paradigms to be combined (`callback soup` and `async/await`), but at the end i wish to prohibit usage of `callback soup`. So i can consider this API as `unsafe` and should only be used inside of `chronos`. All other consumers should avoid this primitives and use `async/await` instead.

arnetheduck:
> Manual Future[T] management is unsafe by design.

maybe, but it's needed to extend chronos itself sometimes to more precisely control execution (locks, special queues, etc) - these functions are rarely used for adding callbacks, but rather for beginning execution chains that `async` functions consume. Exactly because these use cases are advanced and more difficult to write, it's better to have as safe tools as possible to do it, and constructors like these help safety in generrral.

In a few exceptional places, they're also needed to work around inefficiencies in Nim itself, in particular the version we're using today - ie constructing a closure environment is quite expensive in some cases, as is waiting for an entire poll loop.

cheatfate:
Sorry but i'm still not totally convinced by adding this primitives, while tests refactoring could be merged.

chronos/asyncfutures2.nim
```diff
@@ -243,6 +243,12 @@ template complete*[T](future: Future[T], val: T) =
...
+  ## Completes a new ``Future[T]`` with value ``val``.
+  let future = newFuture[T](fromProc)
```
cheatfate, line 248: I think `let` should be replaced by `var`, because `complete()` operation actually changes `Future[T]` state.

arnetheduck: that's not correct though, `Future` is a ref object - `var` would refer to mutability in terms of switching the future variable to a different instance, not changing the contents of the future currently referred.

arnetheduck:
```
let a = newFuture[T](fromProc)
a = newFuture[T](fromProc) # invalid
a.field = x # valid

var b = newFuture[T](fromProc)
b = newFuture[T](fromProc) # valid
b.field = x # also valid
```

the rules for `var` are a bit unfortunate for `ref object` - it's another reason to avoid `ref object` in general - this makes a lot more sense when the code spells out `ref`..

arnetheduck:
tests merged through #196 to focus the discussion on the helpers

arnetheduck:
superseded by #350

PR #201, opened 2021-06-19
use slim test

bung87:
well , slim needs nim >= 1.4 , appveyor seems use a cached old nim version


arnetheduck:
what is slim test? 

In general, we keep chronos compatible with the 1.2 series.

bung87:
it's a fork version of nimble, with slim can drop `nim-unittest2` dependency in main requires, instead requires in before test.

PR #205, opened 2021-06-28
add getData to kqueue,epoll

bung87:
just port from nim std


cheatfate:
I understand that you trying to make it compatible with std, but i don't see any reasons why this calls should be present.

bung87:
https://github.com/bung87/httpbeast/tree/chronos Am trying to replace httpbeast asyncdispatch when verified will port this part to my web framework

cheatfate:
I'm going to reject this PR because of following reasons:
1. Solution provided is not cross-platform, e.g. Windows support is not provided.
2. There no actual reasons for `chronos` was presented.

bung87:
I agree with 2. only need selectors when no `async` used,  I dont understand 1. `getData` already in `ioselectors_select.nim `

PR #206, opened 2021-06-29
Fix accept defect

cheatfate:


Menduist:
https://github.com/status-im/nim-libp2p/runs/2949724337
Works great :+1:

PR #221, opened 2021-09-09
CI: test with multiple Nim branches

stefantalpalaru:


arnetheduck:
this still shows red on nim 1.6/devel, how so? those versions should be allowed-to-fail

stefantalpalaru:
> how so?

GitHub bug: https://github.com/status-im/nim-libp2p/pull/624#issuecomment-917050082

cheatfate:
I want to get some explanations from @Araq because i don't understand why `{.gcsafe.}` is needed if parts of code even do not try to access any global variables.

chronos/asyncsync.nim
```diff
@@ -571,8 +571,9 @@ proc emit[T](bus: AsyncEventBus, event: string, data: T, loc: ptr SrcLoc) =
 
     # Schedule subscriber's callbacks, which are subscribed to the event.
     for subscriber in item.subscribers:
+      let subscriberCopy = subscriber
```
cheatfate, line 574: What the purpose of this changes? `items()` iterator already makes copy of every returned item...

chronos/asyncsync.nim
```diff
@@ -583,8 +584,9 @@ proc emit[T](bus: AsyncEventBus, event: string, data: T, loc: ptr SrcLoc) =
 
   # Schedule subscriber's callbacks which are subscribed to all events.
   for subscriber in bus.subscribers:
+    let subscriberCopy = subscriber
```
cheatfate, line 587: Same question as above^^^.

chronos/asyncmacro2.nim
```diff
@@ -33,21 +33,23 @@ when defined(chronosStrictException):
...
-              break
-
-          if next == nil:
-            if not(retFutureSym.finished()):
-              const msg = "Async procedure (&" & strName & ") yielded `nil`, " &
-                          "are you await'ing a `nil` Future?"
-              raiseAssert msg
-          else:
-            next.addCallback(identName)
+        {.gcsafe.}:
```
cheatfate, line 36: This problem should be fixed in Nim, we should not hide/workaround problem in our code.

chronos/asyncsync.nim
Araq: `items` does not copy anymore in 1.4 and later.

chronos/asyncmacro2.nim
```diff
@@ -70,21 +72,23 @@ else:
...
```
cheatfate, line 75: This problem should be fixed in Nim, we should not hide/workaround problem in our code.

chronos/asyncsync.nim
Araq: Same answer.

cheatfate:
This PR workaround bugs in Nim code.
1. Issue with `.borrow.`.
2. Issue with macro generated iterators.

You can monitor fixes here.
https://github.com/nim-lang/Nim/issues/18877

narimiran:
> Issue with `.borrow.`.

This is now fixed (https://github.com/nim-lang/Nim/pull/18882) and it is already available in `version-1-6` branch of Nim.

chronos/asyncsync.nim
arnetheduck: @Araq what is the intended semantic then, for lent for loop instances in local callbacks / closures? depending on the lifetime of the closure, a copy should or should not be made, for optimal performance

Araq: Currently you cannot access a borrowed location via a closure. The compiler enforces the explicit copy here.

PR #225, opened 2021-09-28
CI: test multiple Nim branches in a daily cron job

stefantalpalaru:
This is an alternative to https://github.com/status-im/nim-chronos/pull/221 which would have marked all PRs and commits as failing.

The downside of a daily CI job is that you can forget about it, but at least it's there when you need it.

### DIFF
--- a/.github/workflows/multi_nim.yml
+++ b/.github/workflows/multi_nim.yml
@@ -1,0 +1,158 @@
+name: Daily
+on:
+  schedule:
+    - cron: "30 5 * * *"
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - os: linux
+            cpu: amd64
+          - os: linux
+            cpu: i386
+          - os: macos
+            cpu: amd64
+          - os: windows
+            cpu: amd64
+          #- os: windows
+            #cpu: i386
+        branch: [version-1-2, version-1-4, version-1-6, devel]
+        include:
+          - target:
+              os: linux
+            builder: ubuntu-18.04
+            shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/multi_nim.yml
```diff
@@ -0,0 +1,157 @@
...
+          cat << EOF > external/bin/g++
+          #!/bin/bash
+          exec $(which g++) -m32 "\$@"
+          EOF
+          chmod 755 external/bin/gcc external/bin/g++
+          echo '${{ github.workspace }}/external/bin' >> $GITHUB_PATH
+
+      - name: MSYS2 (Windows i386)
+        if: runner.os == 'Windows' && matrix.target.cpu == 'i386'
+        uses: msys2/setup-msys2@v2
```
cheatfate, line 75: Sorry but i still do not like this approach, because this setup is almost impossible to use not in CI. 
So i prefer old approach with different URLs, this distributions do not have any dependency and do not require you to setup `msys2` which is not compatible with `Git For Windows`.

cheatfate: https://github.com/status-im/nim-chronos/blob/master/.github/workflows/ci.yml#L83-L94

stefantalpalaru: You might want to see how the sausage is made, first: https://github.com/brechtsanders/winlibs_mingw/issues/30

cheatfate: I'm fine with it. And i really do not care how the sausage is made but i do not like sausages which could be eaten only with predefined plate and sauce. But looks like you like to eat sausages which will become paper if you change plate or sauce.

PR #226, opened 2021-10-04
Prepare chronos for Nim version 1.6

Araq:
It turns out that Chronos generates code like

```nim

proc p() =
  iterator it() = p()
  it()

```

where both `p` and `it` are `.gcsafe` but due to the mutual calls the poor compiler cannot infer that. Hence we patch the macro to help out. This only happened to work for previous Nim compilers where the `.gcsafe`-ty inference was wrong.


Araq:
RFC is here: https://github.com/nim-lang/RFCs/issues/435

PR #230, opened 2021-11-03
make `libbacktrace` a proper dependency

narimiran:
This is why `chronos` was failing on Nim's CI — this change is needed for us to be able to test it and ensure that it keeps working both with Nim 1.2.x and Nim 1.6.x

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,5 +153,4 @@ jobs:
         working-directory: nim-chronos
         run: |
           nimble install -y --depsOnly
-          nimble install -y libbacktrace
           nimble test
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -11,6 +11,7 @@ requires "nim > 1.2.0",
          "stew",
          "bearssl",
          "httputils",
+         "libbacktrace",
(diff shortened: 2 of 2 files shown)


stefantalpalaru:
You can't add per-package testing dependencies in your CI setup?

narimiran:
> You can't add per-package testing dependencies in your CI setup?

What's the problem in having `libbacktrace` as a dependency, just like you have `stew`, `bearssl`, `httputils` and `unittest2`?

Also, if this is an "optional dependency", can this be reflected in the `chronos.nimble` file? Currently you unconditionally test with `--import:libbacktrace` (line 24).

stefantalpalaru:
It's only used in testing. Most users don't need it.

It would make more sense to make a separate "chronos.nimble" task for testing with libbacktrace and just call that explicitly in our CI, while you don't on yours.

narimiran:
> It would make more sense to make a separate "chronos.nimble" task for testing with libbacktrace and just call that explicitly in our CI, while you don't on yours.

Can you (plural) do it, or should I send a PR (and probably do it in non-idiomatic way ;))?

stefantalpalaru:
Done.

PR #231, opened 2021-11-03
CI: separate libbacktrace test in its own task

stefantalpalaru:
This is an alternative to https://github.com/status-im/nim-chronos/pull/230


narimiran:
Anything stopping this from merging?

PR #232, opened 2021-11-10
Ignore HTTP clients that closed before `accept`.

etan-status:
When a client connects to `HttpServerRef` and then closes the connection
before the server had a chance to `accept` and register it, various
POSIX errors may occur. These errors are currently treated as critical
and non-recoverable, resulting in the accept loop being exited. Future
connection requests would no longer be handled by the server; clients
start filling up the listener backlog until full, then get rejected.
This patch handles this case by closing and skipping sockets that were
already closed before any meaningful information could be extracted.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -833,8 +833,47 @@ proc acceptClientLoop(server: HttpServerRef) {.async.} =
       # if server.maxConnections > 0:
       #   await server.semaphore.acquire()
 
-      let transp = await server.instance.accept()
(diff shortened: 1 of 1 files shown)


etan-status:
Ran into this some time with a HTTP client that kept retrying to access a REST API, but the response computation taking longer than the retry period. The issue does not show consistently, usually took about 30 minutes to reproduce.

etan-status:
https://github.com/status-im/nim-chronos/pull/234
That other PR fixes this issue in a way that hides a couple fewer real `TransportOsError` than this simplified PR here does. It also fixes the problem on Windows. Closing this PR as the other one is a better fix.

Issue #233, opened 2021-11-10
`HttpServerRef` exits accept loop when HTTP client closes connection before being accepted.


etan-status:
When a client connects to `HttpServerRef` and then closes the connection before the server had a chance to `accept` and register it, various POSIX errors may occur. These errors are currently treated as critical and non-recoverable, resulting in the accept loop being exited. Future connection requests would no longer be handled by the server; clients start filling up the listener backlog until full, then get rejected.

Ran into this some time with a HTTP client that kept retrying to access a REST API,  but the response computation taking longer than the retry period.  The issue does not show consistently, usually took about 30 minutes to reproduce.

List of known errors that may indicate this problem during `server.instance.accept`.
- accept: [ECONNABORTED]
   A connection has been aborted.
- accept: [EPROTO]
   A protocol error has occurred.
 - setsockopt: [EINVAL]
   The specified option is invalid at the specified socket level or the socket has been shut down.

Note that the presence of these errors indicates that a client connected but immediately closed.
It may still be worthwhile to pass this information onward from `stream.nim` instead of hiding it there.

Unfortunately, the specific errors are underspecified and platform specific (there may be additional ones).
Furthermore, the current implementation hides the errors behind a generic `TransportOsError` that no longer carries info about the specific syscall that failed. Note that this is in line with Nim, e.g., `createAsyncNativeSocket` in `asyncdispatch.nim`.

Ideally, programming errors (i.e., accepting a socket NEVER works) should be caught by assertions next to the call site, and only intermittent socket-specific errors should bubble up as `TransportOsErrors`. e.g., the non-blocking syscall would be a point for an assertion / this is not something that typically fails on just some sockets but not on others.

On top of the errors during `accept`, the `HttpServerRef` furthermore also exits the accept loop when the client closes the connection between the socket being accepted by the server, but before the client's identity is being fetched for the first time (before it is cached).

List of errors that may indicate this during the initial `transp.getId()` .
- getpeername: [EINVAL]
  The socket has been shut down.
- getsockname: [EINVAL]
  The socket has been shut down.

Related PR that fixes this: https://github.com/status-im/nim-chronos/pull/232
However, not sure if this problem also affects non-HTTP servers.

cheatfate:
Fixed in #234.

PR #234, opened 2021-11-11
Handle accept() errors properly.

cheatfate:
Address #233

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -246,9 +246,12 @@ proc dumbResponse*(): HttpResponseRef {.raises: [Defect].} =
   ## Create an empty response to return when request processor got no request.
   HttpResponseRef(state: HttpResponseState.Dumb, version: HttpVersion11)
 
-proc getId(transp: StreamTransport): string {.inline.} =
+proc getId(transp: StreamTransport): Result[string, string]  {.inline.} =
   ## Returns string unique transport's identifier as string.
-  $transp.remoteAddress() & "_" & $transp.localAddress()
+  try:
+    ok($transp.remoteAddress() & "_" & $transp.localAddress())
+  except TransportOsError as exc:
+    err($exc.msg)
 
 proc hasBody*(request: HttpRequestRef): bool {.raises: [Defect].} =
   ## Returns ``true`` if request has body.
@@ -665,7 +668,8 @@ proc `keepalive=`*(resp: HttpResponseRef, value: bool) =
 proc keepalive*(resp: HttpResponseRef): bool {.raises: [Defect].} =
   HttpResponseFlags.KeepAlive in resp.flags
 
-proc processLoop(server: HttpServerRef, transp: StreamTransport) {.async.} =
+proc processLoop(server: HttpServerRef, transp: StreamTransport,
+                 connId: string) {.async.} =
   var
     conn: HttpConnectionRef
     connArg: RequestFence
@@ -675,7 +679,7 @@ proc processLoop(server: HttpServerRef, transp: StreamTransport) {.async.} =
     conn = await server.createConnCallback(server, transp)
     runLoop = true
   except CancelledError:
-    server.connections.del(transp.getId())
+    server.connections.del(connId)
     await transp.closeWait()
     return
   except HttpCriticalError as exc:
@@ -822,7 +826,7 @@ proc processLoop(server: HttpServerRef, transp: StreamTransport) {.async.} =
       # need to close it.
       await conn.closeWait()
 
-  server.connections.del(transp.getId())
+  server.connections.del(connId)
   # if server.maxConnections > 0:
   #   server.semaphore.release()
 
@@ -832,10 +836,16 @@ proc acceptClientLoop(server: HttpServerRef) {.async.} =
     try:
       # if server.maxConnections > 0:
       #   await server.semaphore.acquire()
-
       let transp = await server.instance.accept()
-      server.connections[transp.getId()] = processLoop(server, transp)
-
+      let resId = transp.getId()
+      if resId.isErr():
+        # We are unable to identify remote peer, it means that remote peer
+        # disconnected before identification.
+        await transp.closeWait()
+        breakLoop = false
+      else:
+        let connId = resId.get()
+        server.connections[connId] = processLoop(server, transp, connId)
     except CancelledError:
       # Server was stopped
       breakLoop = true
@@ -845,10 +855,12 @@ proc acceptClientLoop(server: HttpServerRef) {.async.} =
     except TransportTooManyError:
       # Non critical error
       breakLoop = false
+    except TransportAbortedError:
+      # Non critical error
+      breakLoop = false
     except CatchableError:
       # Unexpected error
       breakLoop = true
-      discard
 
     if breakLoop:
       break
--- a/chronos/handles.nim
+++ b/chronos/handles.nim
@@ -105,10 +105,6 @@ proc createAsyncSocket*(domain: Domain, sockType: SockType,
   if not setSocketBlocking(handle, false):
     close(handle)
     return asyncInvalidSocket
-  when defined(macosx) and not defined(nimdoc):
-    if not setSockOpt(AsyncFD(handle), SOL_SOCKET, SO_NOSIGPIPE, 1):
-      close(handle)
-      return asyncInvalidSocket
   register(AsyncFD(handle))
   AsyncFD(handle)
 
@@ -119,10 +115,6 @@ proc wrapAsyncSocket*(sock: SocketHandle): AsyncFD {.
   if not setSocketBlocking(sock, false):
     close(sock)
     return asyncInvalidSocket
-  when defined(macosx) and not defined(nimdoc):
-    if not setSockOpt(AsyncFD(sock), SOL_SOCKET, SO_NOSIGPIPE, 1):
-      close(sock)
-      return asyncInvalidSocket
   register(AsyncFD(sock))
   AsyncFD(sock)
 
--- a/chronos/transports/common.nim
+++ b/chronos/transports/common.nim
@@ -635,16 +635,15 @@ when defined(windows):
 template getTransportTooManyError*(code: int = 0): ref TransportTooManyError =
   let msg =
     when defined(posix):
-      case code
-      of 0:
+      if code == 0
         "Too many open transports"
-      of EMFILE:
+      elif code == EMFILE:
         "[EMFILE] Too many open files in the process"
-      of ENFILE:
+      elif code == ENFILE:
         "[ENFILE] Too many open files in system"
-      of ENOBUFS:
+      elif code == ENOBUFS:
         "[ENOBUFS] No buffer space available"
-      of ENOMEM:
+      elif code == ENOMEM:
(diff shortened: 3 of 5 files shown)


chronos/handles.nim
```diff
@@ -105,10 +105,6 @@ proc createAsyncSocket*(domain: Domain, sockType: SockType,
   if not setSocketBlocking(handle, false):
     close(handle)
     return asyncInvalidSocket
-  when defined(macosx) and not defined(nimdoc):
-    if not setSockOpt(AsyncFD(handle), SOL_SOCKET, SO_NOSIGPIPE, 1):
-      close(handle)
-      return asyncInvalidSocket
```
etan-status, line 111: Hmm. Not sure these are correct.

To avoid `SIGPIPE` during `send` on a socket that is being closed during the send, one of these needs to be done:
- `SO_NOSIGPIPE` on the socket itself (only on BSD)
- `MSG_NOSIGNAL` for each `send` call (Linux)
- `signal(SIGPIPE, SIG_IGN)` on the process (POSIX, but blocks it for the entire process not just for sockets)

https://github.com/tarsnap/libcperciva even goes so far as to temporarily ignore SIGPIPE / restore the handler on each syscall, but that is likely too crazy here xD

https://github.com/nim-lang/Nim/blob/v1.6.0/lib/pure/asyncdispatch.nim#L1655-L1664 uses the same approach as was used here before (with exceptions instead of `invalidSocket` though – here, the caller is throwing the exception though, so same end result).

etan-status: Maybe less disruptive:

Check for the error after the `setSockOpt`, throw `TransportAbortedError` on `EINVAL` and `TransportOsError` otherwise? For `setSocketBlocking`, only `TransportOsError` is good enough (that's what the caller currently makes of of both of the errors).

chronos/transports/common.nim
```diff
@@ -625,3 +631,72 @@ when defined(windows):
...
+
+template getConnectionAbortedError*(code: int): ref TransportAbortedError =
+  let msg =
+    when defined(posix):
+      if code == 0:
+        "[ECONNABORTED] Connection has been aborted before being accepted"
+      elif code == EPERM:
+        "[EPERM] Firewall rules forbid connection"
+      elif code == ETIMEDOUT:
+        "[ETIMEDOUT] Operation has been timed out"
```
etan-status, line 682: Interesting one, haven't encountered the `ETIMEDOUT` one so far on `accept`. But it shows that it is difficult to reliably distinguish between "critical error" and "early close error".

chronos/transports/common.nim
```diff
@@ -625,3 +631,72 @@ when defined(windows):
...
+      "[ECONNABORTED] " & m
+  newException(TransportAbortedError, msg)
+
+template getConnectionAbortedError*(code: int): ref TransportAbortedError =
+  let msg =
+    when defined(posix):
+      if code == 0:
+        "[ECONNABORTED] Connection has been aborted before being accepted"
+      elif code == EPERM:
+        "[EPERM] Firewall rules forbid connection"
```
etan-status, line 680: I have seen `EPERM` without firewalls on some obscure TCP stack before, so it _could_ also be other reasons, but I like the general idea of trying to give some context about what could have gone wrong.

The POSIX spec is more generic here: "A protocol error has occurred; [OB XSR] [Option Start]  for example, the STREAMS protocol stack has not been initialized. [Option End]"
https://pubs.opengroup.org/onlinepubs/9699919799/functions/accept.html

chronos/transports/common.nim
```diff
@@ -625,3 +631,72 @@ when defined(windows):
...
+      of WSAENETRESET:
+        "[ENETRESET] Network dropped connection on reset"
+      of WSAECONNRESET:
+        "[ECONNRESET] Connection reset by peer"
+      of WSAETIMEDOUT:
+        "[ETIMEDOUT] Connection timed out"
+      else:
+        "[" & $code & "] Connection has been aborted"
+    else:
+      "[" & $code & "] Connection has been aborted"
```
etan-status, line 700: The `code` number itself sadly is also platform-specific, so it may have to be combined with a `osErrorMsg` to be useful in isolation (just the number itself is only useful if the underlying platform is known, which is not always provided when someone reports bug)

chronos/transports/stream.nim
```diff
@@ -1049,7 +1051,8 @@ when defined(windows):
                 # was already scheduled, so we failing it not with OS error.
                 retFuture.fail(getServerUseClosedError())
               else:
-                retFuture.fail(getTransportOsError(err))
+                let errorMsg = osErrorMsg(err)
+                retFuture.fail(getConnectionAbortedError(errorMsg))
```
etan-status, line 1055: Hmm, not sure if _all_ of the error codes should become an `ConnectionAbortedError`. Things like `WSANOTINITIALISED` would still be a usage error / a `TransportOsError`.

chronos/transports/stream.nim
```diff
@@ -1063,11 +1066,18 @@ when defined(windows):
...
             server.asock.closeSocket()
             retFuture.fail(getServerUseClosedError())
             server.clean()
+          of OsErrorCode(common.WSAENETDOWN), OSErrorCode(common.WSAENETRESET),
+             OSErrorCode(common.WSAECONNABORTED),
+             OSErrorCode(common.WSAECONNRESET),
+             OSErrorCode(common.WSAETIMEDOUT):
+            server.asock.closeSocket()
+            retFuture.fail(getConnectionAbortedError(int(ovl.data.errCode)))
+            server.clean()
```
etan-status, line 1080: Seems okay, I'm not familiar enough with Windows networking to verify correctness though.

chronos/transports/stream.nim
```diff
@@ -1135,16 +1145,19 @@ when defined(windows):
...
+        except CatchableError as exc:
+          retFuture.fail(exc)
+          return retFuture
 
       if server.asock == asyncInvalidSocket:
         let err = osLastError()
-        if int32(err) == ERROR_TOO_MANY_OPEN_FILES:
-          retFuture.fail(getTransportTooManyError())
+        case int(err)
+        of ERROR_TOO_MANY_OPEN_FILES, WSAENOBUFS, WSAEMFILE:
```
etan-status, line 1159: I wonder who ever had the idea that we need 3 things to specify "out of files" haha. But good catch.

Actually makes me think whether EAGAIN / EWOULDBLOCK should also both be handled (on some weird platforms they are not the same value, and currently we only handle EAGAIN).

chronos/transports/common.nim
etan-status: Will create a followup to fix `osErrorMsg` to use `strerror_r`. `strerror` is not thread-safe.

chronos/transports/stream.nim
```diff
@@ -1734,19 +1753,22 @@ else:
...
+                let errorMsg = osErrorMsg(osLastError())
+                retFuture.fail(getConnectionAbortedError(errorMsg))
             else:
-              let err = osLastError()
-              if int(err) == EINTR:
+              let err = int(osLastError())
+              if err == EINTR:
                 continue
-              elif int(err) == EAGAIN:
+              elif err == EAGAIN:
```
etan-status, line 1762: EWOULDBLOCK as well.

chronos/transports/stream.nim
```diff
@@ -1734,19 +1753,22 @@ else:
...
                 # This error appears only when server get closed, while accept()
                 # continuation is already scheduled.
                 retFuture.fail(getServerUseClosedError())
-              elif int(err) == EMFILE:
-                retFuture.fail(getTransportTooManyError())
+              elif err in {EMFILE, ENFILE, ENOBUFS, ENOMEM}:
+                retFuture.fail(getTransportTooManyError(err))
+              elif err in {ECONNABORTED, EPERM, ETIMEDOUT}:
+                retFuture.fail(getConnectionAbortedError(err))
```
etan-status, line 1769: LGTM. (have not personally seen `ETIMEDOUT` yet, but at least from the error code definition it seems to also be a case of "early close").

etan-status (approved):
Thanks for the fix. This hides fewer `TransportOsError` than my initial fix did.
Also appreciate the Windows support.

Not sure about the removal of `SO_SIGPIPE` – if the processes that use the library already ignore `SIGPIPE` globally that's fine, otherwise that section may need another pass.

chronos/transports/stream.nim
cheatfate: EWOULDBLOCK should be never happened, even if it mentioned in manual `EWOULDBLOCK` means that async framework is not working, unable to properly deliver events that socket is ready for non-blocking operation.

chronos/apps/http/httpserver.nim
```diff
@@ -246,9 +246,12 @@ proc dumbResponse*(): HttpResponseRef {.raises: [Defect].} =
...
   HttpResponseRef(state: HttpResponseState.Dumb, version: HttpVersion11)
 
-proc getId(transp: StreamTransport): string {.inline.} =
+proc getId(transp: StreamTransport): Result[string, string]  {.inline.} =
   ## Returns string unique transport's identifier as string.
-  $transp.remoteAddress() & "_" & $transp.localAddress()
+  try:
+    ok($transp.remoteAddress() & "_" & $transp.localAddress())
+  except TransportOsError as exc:
+    err($exc.msg)
```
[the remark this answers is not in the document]
cheatfate: `EINVAL` for this functions could have different meaning, so i wish not to rename it to `AbortedError`. And there is no need to distinguish because if you unable to identify local or remote side of socket you definitely lost this socket... So doesn't mater at what condition you are you should close the socket. Also `TransportOsError` ins this case should be handled as it handled for `send/receive` (which also could raise `TransportOsError`). Its not a critical error.

chronos/handles.nim
cheatfate: If we unable to perform `setSockOpt` on socket it means we can't continue (we could not make socket non-blocking). So we definitely should close this socket and return an error, that connection could not be made for whatever reason. `EINVAL` has here 2 meanings, so i wish to return `EINVAL` instead of describing what we see here.

chronos/apps/http/httpserver.nim
```diff
@@ -845,10 +855,12 @@ proc acceptClientLoop(server: HttpServerRef) {.async.} =
     except TransportTooManyError:
       # Non critical error
       breakLoop = false
+    except TransportAbortedError:
+      # Non critical error
+      breakLoop = false
```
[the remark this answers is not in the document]
cheatfate: There present some bugs in Nim's exception transformation inside of closure iterators (all async procedures are actually closure iterators). This why all this flags are present, it could be much more easier to use `finally` for example. But problem with loops, breaks and exceptions are present.

chronos/transports/common.nim
cheatfate: Usually i really like to have number, not string which could be confusing because you can't fit everything into one sentence.

etan-status:
Ran my code that originally triggered this problem every ~30 minutes for ~3 hours, and the problem has disappeared.

chronos/transports/common.nim
etan-status: Followup re `strerror_r`: https://github.com/nim-lang/Nim/pull/19126

PR #239, opened 2021-11-20
Nim-1.6 fixes

stefantalpalaru:


stefantalpalaru:
Would you look at that! It also works with Nim-1.4, on most platforms: https://github.com/status-im/nim-chronos/actions/runs/1483286791

PR #244, opened 2021-11-28
posix: fast path for write

arnetheduck:
When `write` is called on a `StreamTransport`, the current sequence of
operations is:

* copy data to queue
* register for "write" event notification
* return unfinished future to `write` caller
* wait for "write" notification (in `poll`)
* perform one `send`
* wait for notification again if there's more data to write
* complete the future

In this PR, we introduce a fast path for writing:

* If the queue is empty, try to send as much data as possible
  * If all data is sent, return completed future without `poll` round
  * If there's more data to send than can be sent in one go, add the
rest to queue
* If the queue is not empty, enqueue as above
* When notified that write is possible, keep writing until OS buffer is
full before waiting for event again

With this flow, as long as there is space in the OS buffer, we will
avoid data copying and poll loops - this above all provides great
performance benefits when there are many small writes to many sockets.

Also fixes an issue where sockets would not be deregistered from the
writer notifications on close.

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -1242,6 +1242,8 @@ else:
   proc removeWriter(transp: StreamTransport) =
     try:
       transp.fd.removeWriter()
+      # For debugging, record that we're no longer getting write notifications
+      transp.state.incl WritePaused
     except IOSelectorsException as exc:
       raiseAsDefect exc, "removeWriter"
     except ValueError as exc:
@@ -1628,9 +1630,16 @@ else:
 
   proc resumeWrite(transp: StreamTransport) {.inline.} =
     if transp.queue.len() == 1:
-      # writeStreamLoop keeps writing until queue is empty
+      # writeStreamLoop keeps writing until queue is empty - we should not call
+      # resumeWrite under any other condition than when the items are
+      # added to a queue - if the flag is not set here, it means that the socket
+      # was not removed from write notifications at the right time, and this
+      # would mean an imbalance in registration and deregistration
+      doAssert WritePaused in transp.state
       try:
         addWriter(transp.fd, writeStreamLoop, cast[pointer](transp))
+
+        transp.state.excl WritePaused
       except IOSelectorsException as exc:
         raiseAsDefect exc, "addWriter"
       except ValueError as exc:
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -1258,105 +1258,83 @@ else:
       return
 
     if WriteClosed in transp.state:
-      let error = getTransportUseClosedError()
-      failPendingWriteQueue(transp.queue, error)
-      # TODO where does removeWriter happen?
-    else:
-      # We exit this loop in two ways:
-      # * The queue is empty: we call removeWriter to disable further callbacks
-      # * EWOULDBLOCK is returned and we need to wait for a new notification
+      doAssert transp.queue.len() == 0, "`close` should have emptied queue"
+      return
 
-      while len(transp.queue) > 0:
-        template handleError() =
-          let err = osLastError()
+    # We exit this loop in two ways:
+    # * The queue is empty: we call removeWriter to disable further callbacks
+    # * EWOULDBLOCK is returned and we need to wait for a new notification
 
-          if cint(err) == EINTR:
-            # Signal happened while writing - try again with all data
-            transp.queue.addFirst(vector)
-            continue
+    while len(transp.queue) > 0:
+      template handleError() =
+        let err = osLastError()
 
-          if cint(err) in [EWOULDBLOCK, EAGAIN]:
-            # Socket buffer is full - wait until next write notification
-            transp.queue.addFirst(vector)
-            return
+        if cint(err) == EINTR:
+          # Signal happened while writing - try again with all data
+          transp.queue.addFirst(vector)
+          continue
 
-          # The errors below will clear the write queue and removeWriter
-          # will be called next as a result
-          if isConnResetError(err):
-            # Soft error happens which indicates that remote peer got
-            # disconnected, complete all pending writes in queue with 0.
-            transp.state.incl({WriteEof})
+        if cint(err) in [EWOULDBLOCK, EAGAIN]:
+          # Socket buffer is full - wait until next write notification - in
+          # particular, ensure removeWriter is not called
+          transp.queue.addFirst(vector)
+          return
+
+        # The errors below will clear the write queue, meaning we'll exit the
+        # loop
+        if isConnResetError(err):
+          # Soft error happens which indicates that remote peer got
+          # disconnected, complete all pending writes in queue with 0.
+          transp.state.incl({WriteEof})
+          if not(vector.writer.finished()):
+            vector.writer.complete(0)
+          completePendingWriteQueue(transp.queue, 0)
+        else:
+          transp.state.incl({WriteError})
+          let error = getTransportOsError(err)
+          if not(vector.writer.finished()):
+            vector.writer.fail(error)
+          failPendingWriteQueue(transp.queue, error)
+
+      var vector = transp.queue.popFirst()
+      case vector.kind
+      of VectorKind.DataBuffer:
+        let res =
+          case transp.kind
+          of TransportKind.Socket:
+            posix.send(fd, vector.buf, vector.buflen, MSG_NOSIGNAL)
+          of TransportKind.Pipe:
+            posix.write(cint(fd), vector.buf, vector.buflen)
+          else: raiseAssert "Unsupported transport kind: " & $transp.kind
+
+        if res >= 0:
+          if vector.buflen == res:
             if not(vector.writer.finished()):
-              vector.writer.complete(0)
-            completePendingWriteQueue(transp.queue, 0)
+              vector.writer.complete(vector.size)
           else:
-            transp.state.incl({WriteError})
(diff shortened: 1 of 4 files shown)


cheatfate:
I think this 
```
Also fixes an issue where enqueued data items would not be sent on
EINTR.
```
should be removed no?

arnetheduck:
> should be removed no?

fixed

chronos/transports/stream.nim
```diff
@@ -1251,171 +1258,91 @@ else:
...
-                if int(err) == EINTR:
-                  continue
-                else:
-                  try:
-                    transp.fd.removeWriter()
-                  except IOSelectorsException as exc:
-                    raiseAsDefect exc, "removeWriter"
-                  except ValueError as exc:
-                    raiseAsDefect exc, "removeWriter"
+      if transp.queue.len > 0:
```
cheatfate, line 1261: `transp.state.incl({WritePaused})` missing here.

chronos/transports/stream.nim
```diff
@@ -1251,171 +1258,91 @@ else:
...
+          # particular, ensure removeWriter is not called
+          transp.queue.addFirst(vector)
+          return
+
+        # The errors below will clear the write queue, meaning we'll exit the
+        # loop
+        if isConnResetError(err):
+          # Soft error happens which indicates that remote peer got
+          # disconnected, complete all pending writes in queue with 0.
+          transp.state.incl({WriteEof})
```
cheatfate, line 1292: `WritePaused` missing here.

chronos/transports/stream.nim
```diff
@@ -1251,171 +1258,91 @@ else:
...
+          if not(vector.writer.finished()):
+            vector.writer.complete(0)
+          completePendingWriteQueue(transp.queue, 0)
+        else:
+          transp.state.incl({WriteError})
```
cheatfate, line 1297: `WritePaused` missing here.

tests/teststream.nim
```diff
@@ -1168,7 +1147,9 @@ suite "Stream Transport test suite":
       await sleepAsync(100.milliseconds)
 
       for i in 0 ..< len(futs):
-        if futs[i].failed() and (futs[i].error of TransportUseClosedError):
+        # writes may complete via fast write
+        if futs[i].completed() or (
```
cheatfate, line 1151: This test check become logically wrong, e.g. if it failed or not i do not care... You should fix the test first, you need to perform writes until you get `EAGAIN` and it starts filling queue.

chronos/transports/stream.nim
```diff
@@ -2101,15 +2027,56 @@ proc getUserData*[T](server: StreamServer): T {.inline.} =
...
+          pbytes = cast[ptr byte](cast[uint](pbytes) + cast[uint](res))
+          rbytes -= res
+
+          if rbytes == 0:
+            retFuture.complete(nbytes)
+            return retFuture
+          # Not all bytes written - keep going
+        else:
+          let err = osLastError()
+          if cint(err) in [EAGAIN, EWOULDBLOCK]:
```
cheatfate, line 2048: why check with array? why not `set`?

chronos/transports/stream.nim
```diff
@@ -2101,15 +2027,56 @@ proc getUserData*[T](server: StreamServer): T {.inline.} =
...
+
+          if cint(err) == EINTR:
+            continue
+
+          if isConnResetError(err):
+            transp.state.incl({WriteEof})
+            retFuture.complete(0)
+            return retFuture
+          else:
+            transp.state.incl({WriteError})
```
cheatfate, line 2059: `WritePaused` is missing

chronos/transports/stream.nim
```diff
@@ -2101,15 +2027,56 @@ proc getUserData*[T](server: StreamServer): T {.inline.} =
...
+        else:
+          let err = osLastError()
+          if cint(err) in [EAGAIN, EWOULDBLOCK]:
+            break # No bytes written, add to queue
+
+          if cint(err) == EINTR:
+            continue
+
+          if isConnResetError(err):
+            transp.state.incl({WriteEof})
```
cheatfate, line 2055: `WritePaused` is missing

chronos/transports/stream.nim
cheatfate: Please restore all the `WritePaused` flags everywhere, because it helps a lot with debugging when you can easily understand is writing loop is running or not. Your check with `transp.queue` feels flaky (i'm not sure is it possible to race there but it smells).

chronos/transports/stream.nim
```diff
@@ -1251,171 +1258,91 @@ else:
...
+
+      var vector = transp.queue.popFirst()
+      case vector.kind
+      of VectorKind.DataBuffer:
+        let res =
+          case transp.kind
+          of TransportKind.Socket:
+            posix.send(fd, vector.buf, vector.buflen, MSG_NOSIGNAL)
+          of TransportKind.Pipe:
+            posix.write(cint(fd), vector.buf, vector.buflen)
```
cheatfate, line 1311: i'm not sure is this conversion is needed...

chronos/transports/stream.nim
```diff
@@ -1251,171 +1258,91 @@ else:
...
+            vector.writer.fail(error)
+          failPendingWriteQueue(transp.queue, error)
+
+      var vector = transp.queue.popFirst()
+      case vector.kind
+      of VectorKind.DataBuffer:
+        let res =
+          case transp.kind
+          of TransportKind.Socket:
+            posix.send(fd, vector.buf, vector.buflen, MSG_NOSIGNAL)
```
cheatfate, line 1309: you can remove `MSG_NOSIGNAL` here.

chronos/transports/stream.nim
arnetheduck: In the new logic for `writeStreamLoop`, there is a direct relationship between the two things: every time there is something in the queue, `writeStreamLoop` needs to be called to remove it from the queue - the only way `writeStreamLoop` gets called is when you register the socket for event notification and the only time you should register the socket for write notification is when the queue is empty. There's an important interplay with the other flags here as well: `writeClosed` is needed so that nothing is _added_ to the queue after `close`, because there is no longer any way to create the necessary notification, because the socket has been closed.

It was the removal of the flag that led me to finding the bug with missing deregistration from the writer: when there are two ways to express the same thing ("I need a write notification"), subtle bugs like this happen easily.

I'm always wary of having two ways to express a concept or idea because during refactoring or optimization, they tend to go out of sync, as happened with `removeWriter` - I can add it back as a debugging tool, but then, I think we should add a `doAssert` that checks the flag vs the queue length as an invariant - any other "ordering" of operations is a bug. Thoughts?

chronos/transports/stream.nim
arnetheduck: can't use `set` with `int`

tests/teststream.nim
arnetheduck: Ok, updated the test, though in reality we're lying a little here: it's no big difference whether the data was dropped by chronos or the OS in this case, and yet we're making it seem like so in the test and in the future we returned.

chronos/transports/stream.nim
arnetheduck: `fd` is a `SocketHandle` which is a `distinct cint` - `write` takes a `cint` - one could argue that the underlying types in the `posix` module, and the types in `chronos`, are quite inconsistent with the reality of `posix` - one could also argue that `posix` itself is inconsistent, and that would be true - basically we could use `write` for both socket and pipe writes - the distinction is cosmetic

chronos/transports/stream.nim
arnetheduck: it was there before, and I didn't really want to change it in this PR

chronos/transports/stream.nim
arnetheduck: I'm not sure I understand why - to me, it looked like a bug to set it here, one that might introduce race conditions.

I looked at where `WritePaused` is being read, and the way it was used was to determine if the socket should be added to write event notifications or not. That is a separate concerned with respect to a write error happening on the socket - basically, when an error happens, we set `WriteEof` or `WriteError`, but why should we set `WritePaused` here? 

The correct place to set the flag would be either inside `completePendingWriteQueue` or in `removeWriter` - this way it would "balance" the operation that causes it to be cleared.

chronos/transports/stream.nim
arnetheduck: `writePause` has been readded in `resumeWrite`/`removeWriter` along with a `doAssert` to ensure internal consistency with respect to the queue length: 611a8230e25dc04a4633fac59821ee5a3675d394

chronos/transports/stream.nim
arnetheduck: 611a8230e25dc04a4633fac59821ee5a3675d394

chronos/transports/stream.nim
arnetheduck: 611a8230e25dc04a4633fac59821ee5a3675d394

chronos/transports/stream.nim
arnetheduck: 611a8230e25dc04a4633fac59821ee5a3675d394

chronos/transports/stream.nim
arnetheduck: 611a8230e25dc04a4633fac59821ee5a3675d394

PR #251, opened 2021-12-27
Raise tracking

Menduist:
EDIT:
This PR adds the possibility to do check exceptions in async procedures. More info from the updated README:

----
By specifying a `asyncraises` list to an async procedure, you can check which
exceptions can be thrown by it.
```nim
proc p1(): Future[void] {.async, asyncraises: [IOError].} =
  assert not (compiles do: raise newException(ValueError, "uh-uh"))
  raise newException(IOError, "works") # Or any child of IOError
```

Under the hood, the return type of `p1` will be rewritten to another type,
which will convey raises informations to await.

```nim
proc p2(): Future[void] {.async, asyncraises: [IOError].} =
  await p1() # Works, because await knows that p1
             # can only raise IOError
```
----


arnetheduck:
This is indeed one of the options we've discussed in the past, to annotate future with an error type akin to `Result`, and make it customizable.

However, it would indeed be nice to reuse the raises instead, and this _should_ be doable eventually - the way forward here _should_ be that we move any given raises annotations to the generated iterator such that the compiler enforces them - what remains is to introspect code snippets with the `effectsOf` pragma from future nim versions, to transport type information about caught exceptions to the continuation site.

There are smaller incremental steps we can take however: one is an `{.atask.}` replacement for `{.async.}` where the only difference is that it acts as a full exception barrier, enforcing the use of `try/catch` - its primary use case would be `asyncSpawn`. A strengthening of strict exceptions, this would change the raises of the generated iterator to `raises: [Defect]` - this is already a requirement that we enforce at runtime.

Finally, there's already a few places experimenting with `Future[Result...` - one pragmatic way forward thus becomes to combine `atask` with special `Result` support in the library to ensure that no exceptions leak when using `Result`.

zah:
This is one of the two possible approaches to solve this problem. Besides changing the `Future` type, it's also possible to introduce an additional wrapper type that will appear in place of the `T` in the `Future` instantiation. This has several potential benefits:

* Less disruption, existing code continues to compile as it is
* Ability to introduce gradually in selected modules first
* The wrapper type can enforce the error handling at the call site if desired

I've written about the design and implemented the key parts of it here:

https://github.com/status-im/nim-stew/blob/b3a6b52c7764da0b351e748b1080127677c734eb/docs/error_handling.md
https://github.com/status-im/nim-stew/pull/26/files

Having that said, changing the `Future` type is quite reasonable as well.

arnetheduck:
> it's also possible to introduce an additional wrapper type that will appear in place

this is good for experiments experiments and proof-of-concepts, but I think ultimately we will want to land in a "native" solution for `nim-chronos` where we don't pay the EH and wrapper overhead (cognitive and implementation-wise) in general, except where those features are used - right now, the flow with exceptions across await barriers is quite expensive because we need to catch exceptions and repropagate them even in code that doesn't raise - better would be that code that doesn't use EH doesn't have to pay the cost either, in any way (not even in the Future object - for that to work, we'd need a more fine-grained `Future` implementation - one that also _maybe_ disconnects writers from readers akin to C++) - basically, if code can be deduced not to raise, it doesn't need to catch, then store, then reraise, and this should be part of the final implementation. 

> Having that said, changing the Future type is quite reasonable as well.

does it even work to do `Future[T, E = CatchableError]` with a default `E`? I wonder how much stuff breaks then.

Menduist:
> does it even work to do Future[T, E = CatchableError] with a default E?

Couldn't make it work. However, we could do 
```nim
type
  EFuture[T, E] =... 
  Future[T] = EFuture[T, (CatchableError)]
```
And then, the async pragma could automatically transform the return Future into a EFuture using the raises as E

> However, it would indeed be nice to reuse the raises instead 

Issue is, iirc, the push pragma is applied after async transformation, so we don't know the raises coming from push yet. 

> transport type information about caught exceptions to the continuation site. 

To my current knowledge, the only way to do this is how I did it here 
We need to know, at compile time, which exceptions can be thrown by a Future given to `await`. Meaning, from the future type, we must be able to get the exception list

arnetheduck:
> To my current knowledge, the only way to do this is how I did it here

that's what `effectsOf` does, in future nim versions - though I anticipate that it will work poorly, given that it has to fully resolve and deduce the effects of the called code, something that nim tends to do quite poorly.

arnetheduck:
> so we don't know the raises coming from push yet.

a starting point might be to get explicit `{.async, raises ...}` to work

zah:
> "native" solution

This is a false dilemma in general. You can implement exactly the same specializations of the `Future` type and the surrounding code using the additional wrapper type.

arnetheduck:
> You can

I'm saying that we must, compared to the proposal in nim-stew, if we're going to actually land it in something like nim-chronos that's used for real.. it's not so much a dilemma as a requirement so as to turn the error handling proposal into a pragmatic and real solution

chronos/asyncmacro2.nim
```diff
@@ -242,19 +235,26 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
 
-template await*[T](f: Future[T]): untyped =
+macro checkMagical(f, e: typed): untyped =
+  result = newNimNode(nnkStmtList)
+
+  let types = getType(e)
+  if types.len < 1: return
+  for errorType in types[1..^1]:
+    result.add quote do:
+      if not isNil(`f`.error) and `f`.error is `errorType`:
```
zah, line 252: The `is` operator works at compile-time. It will compare the [type-erased](https://en.wikipedia.org/wiki/Type_erasure) `error` field to the specified exception types.  To correct the code, you need to use the `of` operator which will work at run-time by examining the concrete exception type that ended being stored in the `error` field.

Having that said, it's a bit unnecessary to go through the trouble of doing these run-time checks. You only need to fool the compiler that the expression can raise all these exceptions even when you directly `raise f.error` without any casting. 

Theoretically, you should be able to do that with a `cast(raises: [...])` block:

```nim
{.cast(raises: [ValueError]).}:
  echo foo
  echo bar
```

... but this requires Nim 1.6. You can also fool the compiler by putting the raises statements in an arbitrary block of code that appears "reachable" to the compiler (even when it's not actually reachable). Accidentally, this is exactly what you did here by using the `is` operator.

zah:
> that's what effectsOf does, in future nim versions?

`effectsOf` has limited utility to the problem at hand here. It can be only useful if `await` is defined as macro that looks at the call expression being awaited. In other words, `await foo()` would discover that `foo()` can raise certain exceptions and it would modify the call-site accordingly. The usefulness is lost when you allow the result of `foo()` to be assigned to a future variable before being awaited and thus it cannot deliver a complete solution. On the other hand, a "compete" solution based on modifying the Future type as it's done here doesn't need to use `effectsOf` in any way.

zah:
> I'm saying that we must, compared to the proposal in nim-stew, if we're going to actually land it in something like nim-chronos that's used for real.. it's not so much a dilemma as a requirement so as to turn the error handling proposal into a pragmatic and real solution

I don't understand any bit of this sentence. What I said is that you can define the error-information-carrying wrapper type at a layer that is above Chronos (or within Chronos) and you can still implement features such as "A Future for an async op that can't raise an  exception doesn't have an `error` field". This renders the argument claiming that there are certain "optimisation" benefits of the `Future[T, E]` approach invalid and I just wanted to clarify that in the interest of having a fair discussion.

chronos/asyncmacro2.nim
Menduist: Oh, right, my bad, should have used of

Though in 1.2, I don't see a better way, not only I have to tell the compiler that this expression can raise E errors, but also that it can't raise anything else

I couldn't find a way to force the raises effect in 1.2, like you did

Menduist:
I did a V2:
```nim
  FuturEx*[T, E] = ref object of FutureBase ## Typed future.
    [..]
    
  Future*[T] = FuturEx[T, (CatchableError)]

converter toFuture*[T, E](futurex: FuturEx[T, E]): Future[T]
```
This allows it to be backward compatible, you can use a Future directly but loose raises info, or use FuturEx to have it

The async macro will read the raises pragma and put info in the return type:
```nim
proc test1(): Future[void] {.async, raises: [ValueError].} =
  raise newException(ValueError, "Value")
```
Will become
```nim
proc test1(): FuturEx[void, (CancelledError, ValueError)] {.stackTrace: off,
    raises: [], gcsafe.} =
  iterator test1_436207619(chronosInternalRetFuture: FuturEx[void,
      (CancelledError, ValueError)]): FutureBase {.closure, gcsafe,
      raises: [CancelledError, ValueError].} =
    block:
      raise newException(ValueError, "Value")
    complete(chronosInternalRetFuture)

  var resultFuture = newFuturEx[void, (CancelledError, ValueError)]("test1")
  resultFuture.closure = test1_436207619
  futureContinue(resultFuture)
  return resultFuture
```

If no `raises` are found, `CatchableError` is added for back. compatibility


Works:
```nim
proc test1(): Future[void] {.async, raises: [ValueError].} =
  raise newException(ValueError, "Value")

proc test2() {.async, raises: [].} =
  let toto = test1()
  try:
    await toto
    echo "Mhh?"
  except ValueError as exc:
    echo "Works"
```
(and also shows why the errors must be stored in the future type, and effectsOf may not help us if we keep our current syntax)

Doesn't compile:
```nim
proc test1(): Future[void] {.async, raises: [ValueError].} =
  raise newException(ValueError, "Value")

proc test2() {.async, raises: [].} =
  await test1()
```

Doesn't compile:
```nim
proc test1(): Future[void] {.async, raises: [ValueError].} =
  raise newException(ValueError, "Value")

proc test2() {.async, raises: [ValueError].} =
  let fut: Future[void] = test1() # We lost raise info here
  await fut # can raise an unlisted exception: CatchableError
```

Menduist:
> one is an {.atask.} replacement for {.async.} where the only difference is that it acts as a full exception barrier

You would still need to inform the caller that you can't raise, otherwise eg
```nim
proc test1() {.atask.} =
  discard

proc test2() {.atask.} =
  await test1() # Can't compile, can raise unlisted exceptions
```

Menduist:
The branches I had to create to make this work with nimbus:
https://github.com/status-im/nim-libp2p/commit/7eb3225d19d0f42be4345380dd0f91fe14069371
https://github.com/status-im/nim-websock/commit/b31ff99b9d5200eea175520c820a3c9f91093bbf
[same issues in json_rpc and nim-presto, but didn't make a branch for theses one yet]

With theses branches, everything compiles correctly

As you can see, the major issue is people (including me) putting wrong raises in async
We could use a different pragma in async context, to avoid this issue and keep backward compatibility, but not sure it make total sense

There is also a weird issue with libp2p on nim devel, need to investigate
(`libp2p/stream/connection.nim(91, 1) Error: Defect can raise an unlisted exception: Defect` ?? looks like a bug on method + raises: [])

wdyt?

chronos/asyncmacro2.nim
```diff
@@ -81,6 +81,27 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+    possibleExceptionsTuple = nnkTupleConstr.newTree(newIdentNode("CancelledError"))
+    foundRaises = -1
+
+  when (NimMajor, NimMinor) < (1, 4):
+    possibleExceptions.add(newIdentNode("Defect"))
+
+  for index, pragma in pragma(prc):
+    if pragma.kind == nnkExprColonExpr and pragma[0] == ident "raises":
+      foundRaises = index
+      for possibleRaise in pragma[1]:
```
dryajov, line 95: shouldn't this be `pragma[foundRaises]`?

dryajov: nvm, this is correct, you're iterating over all the pragmas...

Menduist: The ast returned by pragma looks like that:
```
Pragma
  Ident "gcsafe"
  ExprColonExpr
    Ident "raises"
    Bracket
      Ident "ValueError"
      Ident "CatchableError"
```

The first loop is responsible to find the "raises" pragma, and store it's index in the `foundRaises` (to be deleted later)
Next, we iterate over each exception in the Bracket node, which is `pragma[1]`

dryajov:
Overall this looks really promising! I'd love to see this moving forward and it would be invaluable given the current status-quo!

chronos/asyncfutures2.nim
```diff
@@ -60,6 +60,10 @@ type
       closure*: iterator(f: Future[T]): FutureBase {.raises: [Defect, CatchableError, Exception], gcsafe.}
     value: T ## Stored value
 
+  # Future with a tuple of possible exception types
+  # eg FuturEx[void, (ValueError, OSError)]
+  FuturEx*[T, E] = Future[T]
```
dryajov, line 65: Does this essentially creates an alias instead of a new type? It is a bit of a puzzling construct.

zah: This construct cannot produces a reliable solution indeed. You cannot count that the compiler won't eliminate the information about `E` during the generic instantiation process. Nim tries very hard to normalize generic alias types to their ultimate destination type in `semtypinst` in the compiler.

zah: To clarify further, defining this as a distinct type would work, but what would be the implications of this?

arnetheduck:
Thinking about the Futurex type, one of the goals I had for the future type family was that it should carry no exceptions for code that doesn't raise - `raises` has the weakness that it forces dynamic type on the user which prevents several forms of static analysis, and code that doesn't raise shouldn't have to pay this price.

There are two ways to go forward here, I think - either a base type or allowing E to be void - this is also one of the reasons Result is more fundamentally more powerful as a vehicle for transporting error information: it allows both static and dynamic typing to be used, with the former allowing more powerful static analysis down the line - in a base layer library / type structure like Future, it's an important flexibility to have.

We have to anticipate that whatever type we add here will end up being used in user code as well - although the majority of code uses the macros, there are special cases to consider where the transformation ends up generating poor code and manually written Future code simply is better.

arnetheduck:
> As you can see, the major issue is people (including me) putting wrong raises in async

putting raises on `async` in the current version is wrong indeed - however, `push raises` should not interfere - at least it hasn't interfered prior to this branch.

chronos/asyncfutures2.nim
Menduist: As long as this snippet works, FuturEx will work: https://play.nim-lang.org/#ix=3LXl
And it would seem weird to me if it didn't work

Basically, FuturEx sole purpose in life is to give CT information to await that this future can raise X exceptions

Having a distinct type would probably create havoc in some way, require converter or borrows which are not great

Menduist:
> it should carry no exceptions for code that doesn't raise

Every async procedure can raise CancelledError

> push raises should not interfere

The issue I have in libp2p devel is that this PR will add to the async procedure (not the iterator):
```
1.2: raises: [Defect]
> 1.2: raises: []
```

But in libp2p, we push raises: [Defect] everywhere.
This caused an issue in >1.2 when a async method (with raises: []) is extended by a non-async method (which inherits the pushed raises: [Defect])

Pretty rare case, I think, and libp2p faults for pushing push Defect on >1.2

arnetheduck:
> Every async procedure can raise CancelledError

indeed - this would need addressing at the core level of future as well - this is a problem in general, with the current model: it lacks the ability to express cancellation distinctly from "user code" errors - basically, if the user code raises a `CancelledError`, the model breaks, and this introduces unwelcome fragility at the core.

The fully expressive model would be something like: `FutureNoRaises[Option[Result[T, E]]]]` (where `E` can be an exception, in which case the "exception mode" kicks in and raises, as a way to coat the user code with sugar) - here, every concept is kept apart: the "future" aspect, the "cancellation aspect" and the "outcome of running the user code" aspect - it's verbose and ugly of course, which would need addressing, but that's a separate concern - the semantics of the `await` keyword itself perform the same sin: cancellation is intertwined with "general user error" using the same "reporting" mechanism for both. Key here would be to have primitives in the library that reasonably allow discovery of each of these concepts separately - then, on top one can add sugar for the most common "idioms".

Anyway, I'm writing it down mainly to provide a dump of how far I came along - it's not a blocker per se for improved raises tracking, but it should be considered when chaning the Future type overall.

Menduist:
Yes, current implementation is more geared toward backward compatibility than correctness, unfortunately
After a while it would make sense to do a chronos V2 and break everything we don't like, but that's for another day..

chronos/asyncfutures2.nim
zah: I'm explaining that the compiler will happily erase that information behind your back. You really cannot count on it unless you make your `Gen[T]` type a `distinct int`. You can trust me because this is the area of the compiler that I have most heavily contributed to, but if you insist I can construct some more complex scenarios that will break your example.

zah: There you go:

```nim
type Gen[G] = object
type Gen2[G, E] = Gen[G]

proc test: Gen2[bool, int] = discard

proc checkTestReturnType(T: type, checkMsg: string) =
  let x = test()
  assert type(x) is T, checkMsg
  
checkTestReturnType Gen2[bool, float], "this should be true"
checkTestReturnType Gen2[bool, int], "this should be false"
checkTestReturnType Gen2[string, int], "this shouldn't be reached"
```

Fails with "this shouldn't be reached"

Menduist: I trust you, but need to understand ;)

I do expect him to erase information when converting a `FuturEx` to a `Future`, since we need implicit conversion in that case, but indeed, he seems to erase in more cases than I would like:
```nim
let a: FuturEx[void, (CatchableError,)] = FuturEx[void, (Defect,)].new # compiles, which is not good
# not consistent with
type Gen[T] = int
# let a: Gen[float] = Gen[bool](5) # doesn't compile
type Gen2[T, E] = Gen[T]
let b: Gen2[bool, float] = Gen2[bool, bool](5) # compiles, which is not good
```
So, a "useless" generic parameter pointing to a concrete type will be checked, but not a "useless" generic parameter pointing to a generic type, afaict?

This will cause issue if you do
```nim
proc test1 {.async, raises: [].} = discard
proc test2 {.async, raises: [ValueError].} = raise newException(ValueError, "surprise!")
var fut = test1()
await fut
fut = test2() # should be illegal cause fut kept his type
await fut # will throw defect
```

So that's pretty bad. Now, as I said, I don't see a way to make it work with distinct or others. To keep backward compatibility, we need an implicit FuturEx -> Future conversion, but a locked Future -> FuturEx conversion.

So, we have a few options:
- We find a way to improve that, which would be ideal. The inconsistency makes me thing we could say it's a nim issue? Or if we find a solution to make it distinct or other
- We keep it as-is, and warn about misusages. I think it's reasonable because it fails on "advanced" use cases, unless we found other ways it fail
- We say goodbye to backward compatibility
- We say goodbye to checked exceptions

Personally, now that I've tested a bit of checked exceptions in libp2p with this branch, I can't think going without them.

Menduist: Ok github doesn't seem to auto-update the comments anymore, nice.. So I guess we've hit the same issue independently :)

zah: That's not a Nim issue. A type alias is a statement that two types are equivalent. The code below is correct, because:

```nim
let a: FuturEx[void, (CatchableError,)] = FuturEx[void, (Defect,)].new # compiles, which is not good
# is the same as
let a: Future[void] = Future[void].new
```

You can say goodbye to backwards compatibility or you can go with my suggested approach of encoding the information with an additional type:

```
proc someAsync(): Future[Raising[T, (ValueError, KeyError)]
```

`await` can be specilized to skip over the `raising` type if desired.

dryajov: Interesting suggestion, `Raising` indeed seems like a good meddle ground. I wonder why `Future[T, E = (CatchableError)]` wouldn't work tho?

Menduist: @zah would
```nim
type
  FuturEx[T, E] = ref object of FutureBase
     [...]
  Future[T] = FuturEx[T, (CatchableError,)]
```
Be OK? I've surprisingly managed to make it work with few adjustments, which indicates me that it probably also causes issue. But intuitively, I don't see how the nim compiler could throw away my types if it's part of the "source" of the alias.
At first glance, it seems pretty ok (the example I gave earlier no longer compiles).

Otherwise, need to think about the Raising type

dryajov: Ah, thats even better - hopefully it works. Worst case, `Raising` would be viable path forward IMO.

dryajov: > which indicates me that it probably also causes issue

I'm speculating, but it looks like a more backwards compatible approach - `Future[T] = FuturEx[T, (CatchableError,)]` is essentially exactly the same as `Future[T] = ref object of FutureBase`, because `E` gets erased in most cases? 

If that is indeed the case, I think it's a safer overall approach and we can upgrade the codebase piecemeal without introducing unexpected breakage. One issue I think the current approach has is that it essentially changes the shape of the type underneath and could potentially introduce runtime issues?

dryajov: Current as in the `Future[T, E] = Future[T]`

dryajov: nvm, this doesn't have any runtime effect whatsoever.

zah: The proposed `Future[T] = FuturEx[T, (CatchableError,)]` definition wouldn't suffer from the problems I described indeed. If you managed to make it work, that's great.

Menduist:
Ok so we said goodbye to backward compatibility, now we must limit the breakage. We have a few options:

## Option 1
```nim
type
  FuturEx[T, E] = ref object of FutureBase
     [...]
  Future[T] = FuturEx[T, (CatchableError,)]
```

What it breaks:
```nim
proc test1: Future[void] {.async, raises: [].}
proc test2: Future[void] {.async.}

let
  a: Future[void] = test1() # Won't work, because FuturEx[void, (CatchableError,)] is not compatible with FuturEx[void, (,)]
  b: Future[void] = test2() # Works
```

## Option 2
(major props to @markspanbroek for finding this one)
```nim
  Future*[T] = ref object of FutureBase ## Typed future.
    [..]
  FuturEx*[T, E] = ref object of Future[T]
```

What it breaks:
```nim
type
  broken = proc: Future[void]
  works = proc: Future[void] {.async.} # modified chronos to make it work on proc types

proc test: Future[void] {.async.}
let
  a: broken = test # Proc type are not compatible, even though they should be
  b: works = test # works
```

## Option 3
```nim
  Future*[T] = ref object of FutureBase
    [..]

  Raising[T, E] = distinct Option[T] # or other
```

Haven't look to much at it yet, but it will break at least what option 1 & 2 does. Though, it will happen only when we start to modify the procedures, it will probably work as-is (just like option 1 does)

arnetheduck:
> Ok so we said goodbye to backward compatibility,

well, one strategy is to develop everything on the side to completion using other names (`async2` etc), so there's a fully working option to look towards, _then_ think about how to minimize breakage.,

Menduist:
We can leave backward compatibility (meaning return Future instead of FuturEx) when there is no raise specified.
Everything will continue to compile as is (minus the wrong raises discussed earlier), and we can fix the few incompatibilities (for reference, 5 in libp2p with option 2) while we add the raises through the codebase

arnetheduck:
> We can leave backward compatibility (meaning return Future instead of FuturEx) when there is no raise specified.

I suspect this might be underwhelming - it would mean that when `raises` is specified _on_ the `async` proc, it'll have to do rewriting of the return type from `Future[X]` to the other one, which is surprising (async, for better or worse, does not do return type rewriting) - _or_ the user has to type `FutureEx` instead of `raising` (ie `proc f(): FutureEx[T, E] {.async.}`) - one of the difficulties here is interop with "manual" code which is needed / useful in say .. 10% of the cases. This is why it's attractive to develop a complete solution on the side _then_ worry about backwards compat.

Menduist:
@zah
> Didn't we already establish that Option 2 cannot work?

I think option 2 works, it's not an alias but another type

zah:
> I think option 2 works, it's not an alias but another type

Yes, indeed. I was too quick to comment.

Menduist:
I've added tests, updated the readme & the original message of this PR.

PTAL

README.md
```diff
@@ -151,9 +151,9 @@ feet, in a certain section, is to not use `await` in it.
 
 ### Error handling
 
-Exceptions inheriting from `CatchableError` are caught by hidden `try` blocks
-and placed in the `Future.error` field, changing the future's status to
-`Failed`.
+Exceptions inheriting from `Exception` (or `CatchableError` when
```
arnetheduck, line 154: We should never catch non-`CatchableError` really - in particular, we must not catch `Defect` at all

zah (approved):

README.md
```diff
@@ -211,6 +211,43 @@ originating from tasks on the dispatcher queue. It is however possible that
 `Defect` that happen in tasks bubble up through `poll` as these are not caught
 by the transformation.
 
+#### Checked exceptions
+
+By specifying a `raises` list to an async procedure, you can check which
+exceptions can be thrown by it.
+```nim
+proc p1(): Future[void] {.async, raises: [IOError].} =
```
arnetheduck, line 219: btw, this will break existing code that has raises annotations - we have such code "out there"

tests/testmacro.nim
```diff
@@ -80,3 +80,48 @@ suite "Macro transformations test suite":
     check waitFor(testAwait()) == true
   test "`awaitne` command test":
     check waitFor(testAwaitne()) == true
+
+suite "Exceptions tracking":
+  test "Can raise valid exception":
```
zah, line 85: You need some tests with return values (all current tests use `Future[void]`).

README.md
arnetheduck: we might want to consider `asyncraises` instead - this would neatly work around compatibility issues, and would highlight the semantic differences compared to normal "raises"

tests/testmacro.nim
```diff
@@ -80,3 +80,48 @@ suite "Macro transformations test suite":
...
+    proc test3 {.async, raises: [IOError, ValueError].} =
+      if 1 == 2:
+        raise newException(ValueError, "hey")
+      else:
+        raise newException(IOError, "hey")
+
+    proc test4 {.async, raises: [].} = raise newException(Defect, "hey")
+
+  test "Cannot raise invalid exception":
+    check not (compiles do:
```
zah, line 97: I usually introduce a template called `reject` for test cases like this one:

```nim
template reject(body: untyped) =
  check(not compiles(body))

...

reject:
  proc test3 {.async, raises: [IOError].} = raise newException(ValueError, "hey")

```

chronos/asyncfutures2.nim
```diff
@@ -60,6 +60,10 @@ type
       closure*: iterator(f: Future[T]): FutureBase {.raises: [Defect, CatchableError, Exception], gcsafe.}
     value: T ## Stored value
 
+  # Future with a tuple of possible exception types
+  # eg FuturEx[void, (ValueError, OSError)]
+  FuturEx*[T, E] = ref object of Future[T]
```
zah, line 65: This spelling is a bit too clever perhaps. I anticipate that people will make less typos with the simpler `FutureEx`.

arnetheduck: I'd strive to make it non-public to begin with - once it spreads, we can't get rid of it if it's public - however, in particular for non-raising code, there's a great deal of simplification that can be done to `Future` in general.

README.md
Menduist: iirc to make nimbus compile, there was around 10 lines changed in the whole codebase, not sure it's worth it to duplicate the entire macro instead of just fixing the 10 lines

Especially since theses 10 lines are greatly error inducing

arnetheduck: > duplicate the entire macro

the current duplication was borne out of laziness, not necessity - ie one can factor the macro into better quality code to avoid this

chronos/asyncfutures2.nim
arnetheduck: `RaiseTrackingFuture`?

README.md
Menduist: @arnetheduck to be clear, you're thinking about
`{.async, asyncraises: [].}`
or just
`{.asyncraises: [].}`
?

arnetheduck: the former - new feature, new keyword (purely additive changes help maintain compatibility generally)

I'd actually also add the capability to print warnings when the asyncraises list is missing - this is something we're unable to do with "imissing" `push raises` for example

Menduist: Done

tests/testmacro.nim
Menduist: Neat, thanks! Could be a nice addition to unittest2

tests/testmacro.nim
Menduist: Neat, thanks! Could be a nice addition to unittest2

README.md
arnetheduck: wait, "the former" as in ` {.async, asyncraises: [].}` - the code now does the latter

chronos/asyncfutures2.nim
```diff
@@ -60,6 +60,10 @@ type
       closure*: iterator(f: Future[T]): FutureBase {.raises: [Defect, CatchableError, Exception], gcsafe.}
     value: T ## Stored value
 
+  # Future with a tuple of possible exception types
```
arnetheduck, line 63: document that this should not be used generally and may change in the future

chronos/asyncmacro2.nim
```diff
@@ -72,16 +72,39 @@ proc verifyReturnType(typeName: string) {.compileTime.} =
...
+  var
+    raisesTuple =
+      if raises.len > 0:
+        nnkTupleConstr.newTree()
+      else:
+        ident("void")
+    foundRaises = -1
+
+  for index, pragma in pragma(prc):
+    if pragma.kind == nnkExprColonExpr and pragma[0] == ident "raises":
```
arnetheduck, line 98: wrong raises here?

arnetheduck: ie it should match `asyncraises` for the list that goes in raisestuple

README.md
Menduist: Whoops, forgot my english dictionary at home :sweat_smile: 
Though it's a bit of a mouthful, maybe `{.async, asraises: [].}` or similar would be better?

arnetheduck: I dunno - I'm fine with the longer name: exceptions are supposed to be exceptional after all. 

The other thing I'd like to arrive at, at some point is a mode where no exceptions are allowed through, similar to `nothrows` in c++, which simply disables all these expensive and redundant code paths altogether (catching and re-raising exceptions is significantly slow, in addition to the "normal" slowness of creating an Exception object and in addition to the unpleasant control flow and cognitive overhead issues exceptions cause). 

The other thing is the `.error` pragma that @zah has been contemplating - to make that smooth (as in, coherent and efficient), we'd probably need to get some help for the language, but once that's been developed to completion, perhaps a single pragma with a single semantic can cover both sync and async. Until then, the longer name is fine and in the unlikely event it _really_ becomes a real bother, we can always add a shorter alias later.

arnetheduck: All that said, I'm not strongly tied to keeping this a separate keyword: `asyncraises` achieves the same core objective which is to leave existing code working as-is - it does mix two concepts however in a single keyword, which I generally think should be avoided unless there's exceptional (haha) reasons to do it.

Menduist: Should be easy enough to add a fast-path for RaiseTrackingFuture[_, void] that skips the catching-raising dance
And then add a shortcut for `.async, asyncraises: [].`, eg `.neasync.` (no exception async, terrible name, I'm bad at naming)

arnetheduck: can add that fast path right away, as far as I'm concerned - we can worry about lazy typists later :)

arnetheduck: in the longest run, what I'd really want is for `asyncraises: []` to be the default when you do `async` - ie that you have to enable and explicitly list all exceptions if you want them - but that's a compatibility nightmare.

chronos/asyncfutures2.nim
Menduist: Done

README.md
Menduist: Fast path is an optimization that can be done later without issues, changing the pragma name is a breaking change

Updated to `.async, asyncraises: [].`, and added `-d:chronosWarnMissingRaises`

arnetheduck (approved):

dryajov (approved):
LGTM!

arnetheduck:
btw, is this bumpable on nimbus-eth2? ie does it still compile?

Menduist:
nimbus-eth2 does compile, with a wall of Warnings because of async procedure using .raises.

arnetheduck:
> for non-async functions as well

this is not a blocker for merging btw, it might be nontrivial to implement

Menduist:
Nice! Would imply some trickery because there would be two rewriting macro that need to work in any order, but should be doable

We could also check the trackedfuture.fail(exc) at CT to protect against incompatible errors

The only missing piece would be how to create the future nicely
```nim
let retfut = newTrackedFuture[int]
```
That reads the result type?

I think it's important to improve the manual tracking since the beginning of the process will be to transform chronos which is mostly manual

Menduist:
Did a first version, has a lot of code duplication ATM, will be improved after https://github.com/status-im/nim-chronos/pull/261 is merged
It's just missing the `trackedfuture.fail` checker

Menduist:
Finished the manual handling part (though still waiting on #261 to cleanup a bit):
```nim
    proc test2: Future[int] {.asyncraises: [IOError, OSError].} =
      result = newRaiseTrackingFuture[int]()
      result.fail(newException(IOError, "fail"))
      result.fail(newException(OSError, "fail"))
      result.fail(newException(ValueError, "fail"))
```

Last line fails to compile with
```
test.nim(failline, 13) template/generic instantiation of `fail` from here
chronos/asyncfutures2.nim(316, 19) template/generic instantiation of `checkFailureType` from here
nim-#devel/lib/system/fatal.nim(53, 5) Error: unhandled exception: Can't fail with ValueError. Only [IOError, OSError] allowed [AssertionDefect]
```
(also works with inheritance)

chronos/asyncmacro2.nim
```diff
@@ -236,35 +285,87 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+
+  result = nnkIfExpr.newTree(
+    nnkElifExpr.newTree(
+      quote do: isNil(`f`.error),
+      quote do: discard
+    )
+  )
+
+  result.add nnkElifExpr.newTree(
+    quote do: `f`.error of type CancelledError,
```
arnetheduck, line 344: why does this need special casing? ie if a function uses a call that potentially can be cancelled and doesn't handle that cancellation, it should `asyncraises: [CancelledError]`

arnetheduck: conversely, a call that disarms cancellederror will no longer leak cancellations and should have this effect

chronos/asyncfutures2.nim
```diff
@@ -60,6 +61,11 @@ type
       closure*: iterator(f: Future[T]): FutureBase {.raises: [Defect, CatchableError, Exception], gcsafe.}
     value: T ## Stored value
 
+  ## Future with a tuple of possible exception types
+  ## eg RaiseTrackingFuture[void, (ValueError, OSError)]
+  ## Should generally not be used manually
```
arnetheduck, line 66: here, we should document that one should annotate with `asyncraises` instead of using "manually", for callback type definitions for example - also, afair shouldn't type comments come below the type, indented?

chronos/asyncmacro2.nim
Menduist: I think it would add more noise than anything else (especially since we can't `push` this)
ie:
```nim
proc a {.async, asyncraises: [CancelledError].} =
  await sleepAsync(10.minutes)

asyncSpawn a()
```
In this setup, `a` can't actually raise cancelled error, but it's still in the tracking. You would have to do
```nim
proc a {.async, asyncraises: [].} =
  try:
    await sleepAsync(10.minutes)
  except CancelledError as exc:
   discard

asyncSpawn a()
```

And that would be the only case where it would make sense to track it: 95% of procedures will just bubble up the cancellation, after possibly cleaning some stuff
So I see little gain in tracking it, but a lot of noise

Menduist:
> when hitting a violation, currently a rather unhelpful message is printed

Weird, I get correct error messages on nim 1.2 & devel:
```nim
import ./chronos/[asyncloop, asyncsync]

proc test1() {.async, asyncraises: [ValueError], gcsafe.} =
  raise newException(ValueError, "Value")

proc test2() {.async, asyncraises: [], gcsafe.} =
  await test1()
```

```
/home/tavon/Status/nim-chronos/testexc.nim(6, 16) template/generic instantiation of `async` from here
/home/tavon/Status/nim-chronos/testexc.nim(7, 14) Error: cast[ref ValueError](chronosInternalTmpFuture.error) can raise an unlisted exception: ref ValueError
```

Menduist:
Indeed, you still get the correct line in the:
```
nim-chronos/testexc4.nim(6, 17) template/generic instantiation of `async` from here
nim-chronos/testexc4.nim(7, 3) template/generic instantiation from here # HERE
nim-chronos/chronos/asyncmacro2.nim(389, 31) Error: internalCheckComplete(chronosInternalTmpFuture) can raise an unlisted exception: CatchableError
```
But that could be improved :)

chronos/asyncfutures2.nim
Menduist: Done

chronos/asyncmacro2.nim
arnetheduck: The main problem here is that `a` can and indeed does raise a `CancelledError`, depending on what the caller does - it's just that it so happens that this functionality is not used by the caller in this particular case. 

There's also another thing to consider: some tasks are inherently not cancellable:

```nim
proc f() {.async.} = discard
```

can indeed never raise `CancelledError` because it will always returns a completed future (and cancelling a completed future has no effect) - putting cancellederror in asyncraises documents / makes explicit which tasks indeed are cancellable. 

We *should* be able to track this through `await` calls as well - ie intuitively `proc g() {.async, ..}= await f()` should be able to determine that no cancellation is possible.

We might need language to define non-cancellable tasks, but that's an orthogonal issue.

Eventually, we also want to fix the `push` issue, perhaps with some help from the compiler.

I *think* the right thing to do here is to make it explicit, at least for now - if we add the magic later on, it should be harmless - ie we must always support `asyncraises: [CancelledError]` - it should not be "wrong" to spell it out.

Menduist: > makes explicit which tasks indeed are cancellable

I don't fully agree here.
```nim
proc dial(address: MultiAddress): Future[Connection] {.async, asyncraises: [DialError, UpgradeError, CancelledError].} =
  let conn = await whatever(address)
  try:
    await conn.upgrade()
  except UpgradeError as exc:
    await conn.close()
    raise exc
  return conn
```
This procedure is technically cancellable, but it's wrong: conn will leak if we get cancelled during the upgrade
So this is not a good signal that the CancelledError is handled properly


> We should be able to track this through await calls as well - ie intuitively proc g() {.async, ..}= await f() should be able to determine that no cancellation is possible.

```nim
proc g() {.async, ...} = await asyncSleep(10.minutes)

g().cancel()
```
`await` can't determine if it's cancellable or not, since it depends on the behavior of callers & callees

> I think the right thing to do here is to make it explicit, at least for now

Well, that would be the only way to change it in the future without backward compatibility break, indeed. But still not convinced it makes sense in the first place

arnetheduck: > This procedure is technically cancellable, but it's wrong:

So think about it this way: 
* `dial` is cancellable because `upgrade` is cancellable (because ultimately, it passes though `select` which goes through the event loop which introduces the possibility for cancellation)
* how does the compiler know `upgrade` is cancellable? because it has a `CancelledError` raises effect
* when you're writing `dial`, the first thing you type is `asyncraises: []` assuming nothing is cancellable and there are no errors (with `push` support, this is what we'll enforce in every module top-level, similar to `push raises: []`)
* the compiler complains: `upgrade` raises cancellederror that you need to handle (together with a bunch of other errors)
* you now write the EH for cancellederror and reraise it

the shorter version of the above is that as you're typing `dial... asyncraises: [CancelledError]`, you have a clear and visual indication that cancellationerror is possible, and that you need to handle it in your function. One  can imagine a warning for functions that have overly broad raises lists as well (though this is dubious: a broad exception list is one way to make API future-proof, by including the possibility to raise new errors, even if the code presently doesn't).

In the hidden version, we're exactly where we are today: there's no indication that cancellation errors may happen, so programmers don't deal with it, just like they don't deal with exceptions in general. It also leaves no way to indicate that a function does _not_ support cancellation.


> await can't determine if it's cancellable or not, since it depends on the behavior of callers & callees

You're confusing whether a function is cancellable with whether it gets cancelled. The former is determined by the implementation of the function: `sleepAsync` is cancellable because it interacts with the event loop, and therefore should have a `raises: [CancelledError]`.

Consider this:
```
proc f(b: bool) {. raises [ValueError].} =
  if b:
    raise (ref ValueError)()
```

In the above case, the _caller_ determines whether the function actually raises (dynamically) - the same situation applies to cancellation and its associated exception: the caller might call cancel or not, but in the implementation of `f`, you have to assume that it's a possibility and write the cancellation handling code.

arnetheduck: > t, but in the implementation of f, you have to assume that it's a possibility and write the cancellation handling code.

the exception to this is if you know that you call no cancellable functions: then, you can safely write `f` without considering `cancel` - and have a `raises: []`.

By extension, `await` can also know if the entire chain of functions is cancellable or not by examining the raises effect of each called function - interestingly, this also tells you if a function actually could be made `sync` :)

Menduist: > sleepAsync is cancellable because it interacts with the event loop

Yes, but in practice, every async procedure will go to the event loop, either directly, or through it's callees. Because otherwise, why is it async? Which is my original point: almost every async procedure will raise the CancelledError, because almost every async procedure uses await on something eventually going to the event loop
The only (sensible) async procedures that don't raise CancelledError are the one swallowing it (which is dubious in other ways)


Your first scenario could work well if we forced the user to remap the CancelledError to something else, eg:
```nim
proc f {.raises [CancelledError].} =
  try: await sleepAsync(xx) # will throw ChildCancelledError
  except ChildCancelledError as exc: bubbleup(exc) # or whatever, a remap to CancelledError
```

arnetheduck:
> Because otherwise, why is it async?

future-proofing, uniterruptible tasks, remapping CancelledError to a "regular" response

> one swallowing it (which is dubious in other ways)

well, one way to deal with cancellation is to give a non-cancellation error - ie change `CancelledError` to `0 bytes read`, breaking the cancellation chain: this gives functions that do go through the event loop a way to stay cancellation-free, isolating cancellation to specific parts of the code (just like you simplify other error codes generally: FileReadError becomes DbError becomes sometihng else as you move through layers)

chronos/asyncmacro2.nim
arnetheduck: > Because otherwise, why is it async?

future-proofing, uniterruptible tasks, remapping CancelledError to a "regular" response

> one swallowing it (which is dubious in other ways)

well, one way to deal with cancellation is to give a non-cancellation error - ie change `CancelledError` to `0 bytes read`, breaking the cancellation chain: this gives functions that do go through the event loop a way to stay cancellation-free, isolating cancellation to specific parts of the code (just like you simplify other error codes generally: FileReadError becomes DbError becomes sometihng else as you move through layers)

Menduist: That could become interesting, and even a possible solution to the "cancellation race issue", but that will imply some heavy changes to the cancellation scheme

But ok, I'm starting to see some interesting possible behaviors, so I'll err on the "can be changed in a non-breaking way" side :)

arnetheduck: > Your first scenario could work well if we forced the user to remap the CancelledError to something else, eg:

I'd actually prefer this, except that it's insanely verbose. We have a less verbose solution for that though, `Result` :) Once the exception tracking is in, I'd say an interesting area to explore is Future/Result integration at some level (ranging from making `?` compatible to 

All of this discussion goes back to what information should be visible / captured, and where: the "true" return type of a cancellable `async` function, if you consider the full gamut of control-flow options, can be expressed roughly as `Future[Option[Either[T, XxxError]]`: through experience, we know that hiding `XxxError` from that signature leads to code not considering it - you're saying we should hide the `Option` part as well: the same thing happens.

Once you put both cancellation and error information in the function signature (via `asyncraises: [Cancelled, Xxx]`), the remaining practical differences between `Result` and Exception are:

* `Result` enforces call-site annotations (its main feature) - Exceptions bubble up without any visual indication at the call site
* `Result` allows static typing / type analysis (and permits dynamic typing) - `Exception` forces dynamic typing

We'll see how far the balance goes, but a good starting point nonetheless is to make the function signature rich(er): each library can then decide about call-site annnotations, as we build out the capabilities of these.

Perhaps we'll end up with `asynccancellable: [A, B]` to shorten things, or a `read(v: Future): Opt[Result[T, E]]` with nice syntactic sugar to make cancellation and errors easy to work with.

Menduist: d5eed32

chronos/asyncfutures2.nim
```diff
@@ -571,6 +571,48 @@ proc internalRead*[T](fut: Future[T]): T {.inline.} =
   when T isnot void:
     return fut.value
 
+macro checkFutureExceptions*(f, typ: typed): untyped =
```
cheatfate, line 574: Whatever you are doing please leave all the macro related code in `asyncmacro2.nim`

arnetheduck:
an alternative syntax to `asyncraises` was discussed: `proc x() {.async(raises = []).}` - this would avoid the need for separate pragmas and perhaps simplify the macro itself - to be evaluated while this PR is rebased.

It's nice because it also offers a way forward for other extensions (ie making the strict exception mode per-function instead of global etc)

arnetheduck:
> an alternative syntax

unfortunately, doing it with the alternative synax does not cover the rewriting of `Future` into `RaiseTrackingFuture` - either `async(expand=false, raises=[])` or we need `asyncraises` _also_ if we want to hide RTF

Menduist:
Bumped this to be on top of #449

arnetheduck:
This is now looking ok:ish - since it was easy, we also no longer generate try/finally at all for `asyncraises: []` eliminating a significant amount of bloat :)

arnetheduck:
This version of the PR is sufficient for enabling experimentation with the feature, as the changes are additive (should not significantly change code that doesn't use it) - merging so as to simplify further work on the feature while testing it in real-life code

PR #254, opened 2022-01-14
Add onDone, onComplete, onUncomplete, onFail, onCancel

Menduist:
As we want user to be able to work without handling exceptions (with results for instance), we have an issue with cancellation.
Currently, the only way to handle it is by wrapping every `await` in a `try / except`, and handle the CancelledError

This PoC PR offers an alternative, by allowing users to do:
```
proc dial(ma: MultiAddress): Future[Connection] {.async.} =
  let conn = await dial()
  onCancel: conn.close()
  return await upgrade(conn)
```

I took the opportunity to add a few option:
`onDone`: will always be called
`onCompleted`: will be called if the future succeeded
`onUncompleted`: will be called if the future failed or was cancelled
`onFail`: will be called if the future failed
`onCancel`: will be called if the future was cancelled

This is similar, in principle, to a defer, with the key difference that it's bound to the future, not the scope

If we find better names, would you think this is worthy of pursuit?

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -294,6 +294,37 @@ template awaitne*[T](f: Future[T]): Future[T] =
   else:
     unsupported "awaitne is only available within {.async.}"
 
+template onDone*(code: untyped): untyped =
+  when not declared(chronosInternalRetFuture):
+    unsupported "onFinished is only available within {.async.}"
+
+  chronosInternalRetFuture.addCallback(proc (fut: pointer) {.gcsafe, raises: [Defect].} =
+    let future {.inject.} = cast[type(chronosInternalRetFuture)](fut)
+    GC_unref(future)
+    code
+  , cast[pointer](chronosInternalRetFuture))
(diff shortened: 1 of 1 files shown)


cheatfate:
This approach is incorrect and unsafe, there many issues present.
1. Because you are adding callback to `chronosInternalRetFuture` it is possible that you will be called not first, but last in sequence. In such case parent's callback will be scheduled first and parent procedure which awaits your procedure will proceed with result of your procedure before your `onXXX` callback will be called. It will happens if your code which adds callbacks to `chronosInternalRetFuture` will be placed after first `await` call.
2. When you got error or cancellation from awaited Future - exception will be raised CatchableError/CancelledError which at the end will be catched by main `try/except` branch and at this point Future will be marked as `failed/cancelled` and all callbacks will be scheduled. If your onCancel/onError will be placed after `await` - this code will never be executed, because addCallback will not add callback, But if setup your `onCancel/onError` before `await` your code will be scheduled after all the exception handlers will be passed, and be careful with situation which i have described at point 1.

PR #257, opened 2022-01-21
Refactor `preferredContentType` proc & add tests

EmilIvanichkovv:
Fix of the following bug:
In case of multiple accept headers with same preference
`preferredContentType` used to select the first match within content types
provided by the application. For example, if user specifies accept headers
`application/octet-stream, application/json` and application provides
`application/json, application/octet-stream`, `application/octet-stream`
will be returned, and that is a bug.

Now the procedure returns the most suitable match according
to the application preferences.


zah (commented):
We need a test for the problematic case I've described here:
https://github.com/status-im/nim-http-utils/pull/21/files#r790098655

PR #261, opened 2022-02-23
Cleanup return type open syms

Menduist:
Fixes https://github.com/status-im/nimbus-eth1/issues/963
For some reason, the code was generating this kind of AST for the return type:
```
    Call
      OpenSymChoice
        Sym "[]"
        Sym "[]"
        ...
      Sym "Future"
      Call
        OpenSymChoice
          Sym "[]"
          Sym "[]"
          ..
        Ident "seq"
        Ident "OpenObject"
```
And this "recursive" OpenSymChoice wasn't handled properly by the macro

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -72,6 +72,21 @@ proc verifyReturnType(typeName: string) {.compileTime.} =
 macro unsupported(s: static[string]): untyped =
   error s
 
+proc cleanupOpenSymChoice(node: NimNode): NimNode {.compileTime.} =
+  # Replace every Call -> OpenSymChoice by a Bracket expr
+  # ref https://github.com/nim-lang/Nim/issues/11091
+  if node.kind in nnkCallKinds and
+    node[0].kind == nnkOpenSymChoice and node[0].eqIdent("[]"):
+    result = newNimNode(nnkBracketExpr).add(
+      node[1],
+      node[2]
+    )
+  else:
+    result = node
+
+  for index, child in result:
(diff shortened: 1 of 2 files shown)


mratsim:
I'm unsure how often we stumble upon this but maybe we need a `toUntypedAST` proc in `stew` to cleanup all the situation where we end up receiving `typed` AST in an untyped macros?

For example, I'm sure we hit this in serialization libraries macros. I didn't test using taskpools in a generic proc or a template yet but it's very possible as well.

Menduist:
It would be a bit overkill for this fix, since we can only have OpenSyms AFAICT, but sure would be good to have as a general helper
+ this kind of bugs are hard to find

Menduist:
https://github.com/nim-lang/Nim/issues/11091#issuecomment-486171129
Apparently we could also put it in the stdlib

Araq:
The fix is acceptable but I would do it differently. I would use an `extractReturnType` helper instead. Your patch is conceptually quite brutal -- you change the AST to something else and then you extract what you're interested in.

Menduist:
Not sure I understand @Araq, is it because I modify the AST in place instead of copying it?
If so, last commit fixes that

mratsim:
In my experience working with types with macros is fraught with peril https://github.com/nim-lang/RFCs/issues/44 and worse, this might manipulate generics which tend to become a timesink https://github.com/status-im/nimbus-eth2/issues/2219

If we can avoid manipulating types in macros and just modify AST to make it canonical untyped, it will likely save hours of debugging afterwards at the cost of forcing the compiler to redo semcheck/sigmatch.

PR #263, opened 2022-03-14
Rate limiter

Menduist:
Adds a generic rate limiter:
```nim
var counter = RateCounter.init(cap=1.seconds)
echo counter.tryConsume(10.operationsPerSecond) # succeeds
await counter.consume(operationsPerSecond(0.1)) # Will block for ~9 seconds
```

### DIFF
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -19,8 +19,6 @@ type
     budgetCap: int64
     lastUpdate: Moment
 
-  RateCounterRef* = ref RateCounter
-
   RateCost* = distinct int64
 
 proc update(rc: var RateCounter) =
@@ -93,8 +91,8 @@ proc init*(T: typedesc[RateCounter], cap: Duration): T =
     lastUpdate: Moment.now()
   )
 
-proc new*(T: typedesc[RateCounterRef], cap: Duration): T =
-  T(
+proc new*(T: typedesc[RateCounter], cap: Duration): ref RateCounter =
+  (ref T)(
     budget: cap.nanoseconds,
     budgetCap: cap.nanoseconds,
     lastUpdate: Moment.now()
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -36,6 +36,18 @@ proc operationsPerSecond*(ops: float): RateCost =
 
   RateCost(budgetPerSecond / ops)
 
+proc tokensPerSecond*(tokens, budget: SomeNumber): RateCost =
+  ## Create a RateCost from a token budget per second
+  runnableExamples:
+    let cost = tokensPerSecond(
+      1518, # packet of 1518 bytes
+      1000000 # budget of 1mb
+    )
+  let
+    budgetUsed = tokens / budget
+    allowedPerSeconds = 1.float / budgetUsed
+  operationsPerSecond(allowedPerSeconds)
+
 proc timeBudgetPerSecond*(time, budget: Duration): RateCost =
   ## Create a RateCost from a time budget per second.
   runnableExamples:
@@ -49,9 +61,7 @@ proc timeBudgetPerSecond*(time, budget: Duration): RateCost =
   let
     timeAsUs = time.nanoseconds.float
     budgetAsUs = budget.nanoseconds.float
-    budgetUsed = timeAsUs / budgetAsUs
-    allowedPerSeconds = 1.float / budgetUsed
-  operationsPerSecond(allowedPerSeconds)
+  tokensPerSecond(timeAsUs, budgetAsUs)
 
 proc tryConsume*(rc: var RateCounter, cost: RateCost): bool =
   ## If there is still budget left, remove cost from the budget
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -1,0 +1,83 @@
+#               Chronos Rate Limiter
+#            (c) Copyright 2022-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+
+{.push raises: [Defect].}
+
+import ../chronos
+
+export timer
+
+type
+  RateCounter* = object
+    budget: int64
+    budgetCap: int64
+    lastUpdate: Moment
+
+  RateCounterRef* = ref RateCounter
+
+  RateCost* = distinct int
+
+proc update(rc: var RateCounter) =
+  let
+    currentTime = Moment.now()
+    timeDelta = nanoseconds(currentTime - rc.lastUpdate)
+
+  rc.lastUpdate = currentTime
+  rc.budget = min(rc.budgetCap, rc.budget + timeDelta)
+
+proc operationsPerSecond*(ops: float): RateCost =
+  ## Create a RateCost from an allowed number of 
+  ## operations per second
+  const budgetPerSecond = 1.seconds.nanoseconds.float
+
+  RateCost(budgetPerSecond / ops)
+
+proc tryConsume*(rc: var RateCounter, cost: RateCost): bool =
+  ## If there is still budget left, remove cost from the budget
+  ## Otherwise, return false
+  rc.update()
+  if rc.budget >= 0:
+    rc.budget.dec(int(cost))
+    true
+  else:
+    false
+
+proc consume*(rc: var RateCounter, cost: RateCost): Future[void] =
+  ## Wait until some budget is available, then substract cost
+  ## from it
+
+  # Manual async because of var argument
+  rc.update()
+  rc.budget.dec(int(cost))
+
+  if rc.budget >= 0:
+    result = newFuture[void]("RateCounter.consume")
+    result.complete()
+    return result
+  else:
+    return sleepAsync(nanoseconds(-rc.budget))
+
+proc init*(T: typedesc[RateCounter], cap: Duration): T =
+  ## Create a RateCounter.
+  ## The RateCounter will be smoothed out over the
+  ## duration of `cap`.
+  ##
+  ## Example: RateCounter.init(10.seconds)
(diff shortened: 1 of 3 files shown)


chronos/ratelimit.nim
```diff
@@ -0,0 +1,101 @@
...
+  ##
+  ## Example: RateCounter.init(10.seconds)
+  ## 10 seconds worth of budget could be used instantly
+  T(
+    budget: cap.nanoseconds,
+    budgetCap: cap.nanoseconds,
+    lastUpdate: Moment.now()
+  )
+
+proc new*(T: typedesc[RateCounterRef], cap: Duration): T =
```
arnetheduck, line 96:
```suggestion
proc new*(T: typedesc[RateCounter], cap: Duration): ref T =
```

`new` is for the non-ref type because it introduces `ref`, unlike init which gives you an instance of the type you asked for - in other parts of the codebase we've also stopped using `Ref` aliases - typing `ref X` is almost the same

chronos/ratelimit.nim
```diff
@@ -0,0 +1,101 @@
...
+  rc.budget -= int64(cost)
+
+  if rc.budget >= 0:
+    result = newFuture[void]("RateCounter.consume")
+    result.complete()
+    return result
+  else:
+    return sleepAsync(nanoseconds(-rc.budget))
+
+proc init*(T: typedesc[RateCounter], cap: Duration): T =
```
arnetheduck, line 83: Hm, I think we'll benefit from using a more standard setup of naming and initialization here - the [ones](https://github.com/osrg/openvswitch/blob/master/lib/token-bucket.c#L33) I've seen take a separate rate and cap (so that I can create a 100 tokens/sec bucket with a burst capacity of 1000, for example) - it's not obvious to me how to create such a bucket from these parameters.

Instead of "operations" I'd also go with "tokens" - there's a lot of [literature](http://intronetworks.cs.luc.edu/current/html/tokenbucket.html#token-bucket-definition) out there describing this particular algorithm so it's good if it's instantly recognizable - it should be documented with the name of the algo as well.

arnetheduck: In nimbus for example, we'll likely want to switch to having [two of these buckets](http://intronetworks.cs.luc.edu/current/html/tokenbucket.html#multiple-token-buckets): a smaller bursty one and a global more "even" one - when other clients ask for blocks, we want each peer to get an allocation but also put a global limit on how much bandwidth we  assign to providing blocks.

That said, what the "tokens" are is arbitrary - it's usually bytes, but sometimes packets or requests - there are arguments for each, depending on what's costly.

arnetheduck: Finally, I've been thinking about a manual replenish function as well: instead of a rate / sec, in some cases I might want to replenish manually: the loose idea I had was that we should not spend more on block uploads than we spend on gossip - this would mean that every time we send gossip, we'd replenish the block bucket up to its cap, so that max block bandwidth is expressed in terms of gossip bandwidth - it's a loose idea so far, but it does solve one problem which is how to "size" the buckets naturally.

mratsim:
> In nimbus for example, we'll likely want to switch to having [two of these buckets](http://intronetworks.cs.luc.edu/current/html/tokenbucket.html#multiple-token-buckets): a smaller bursty one and a global more "even" one - when other clients ask for blocks, we want each peer to get an allocation but also put a global limit on how much bandwidth we assign to providing blocks.
> 
> That said, what the "tokens" are is arbitrary - it's usually bytes, but sometimes packets or requests - there are arguments for each, depending on what's costly.

In terms of modules we have:
- discovery
- libp2p
- Nimbus eth2_network https://github.com/status-im/nimbus-eth2/blob/v1.7.0/beacon_chain/networking/eth2_network.nim
- RPC
  - Generic RPC, metrics 
  - Beacon Node, Validator Client link
  - Consensus <-> Execution link

I think for now we can focus on discovery/libp2p/Nimbus.
If we start adding bandwidth limits we need metrics on bandwidth and messages sent by each module so that it's easier understanding the steady state vs an under-attack scenario.

arnetheduck:
>  If we start adding bandwidth limits we need metrics on bandwidth and messages sent by each module so that it's easier understanding the steady state vs an under-attack scenario.

sure - a bandwidth limit though is nothing other than using `tokens == bytes` - I'm merely outlining the future use cases here so that the chronos support works for as many as possible.

chronos/ratelimit.nim
Menduist: I thought a bit about it, and couldn't find a use case wherever scaling the "costs" instead of the fill rate didn't work

Hence my idea here to simplify the api to just burst capacity / cost, which is I think easier to reason about

But indeed, to handle manual replenish, we might have to expose the fill rate 🤔

arnetheduck:
> await bucket.consume(tokensPerSecond(2000, 1000))

Well, this reads a bit wrong: you're not "consuming tokens per second" - you're consuming tokens once - the per-second part makes the API unintuitive to read.

> await bucket.consume(operationsPerSecond(10))

This one is even more problematic: operations can't be "consumed" really - it's the tokens that get consumed in order to let the operations proceed.

> share the bucket between multiple applications using different systems

why would you want to do that? these systems either would need to know about each other so as to use the same "scale" for their token consumption (which is arbitrary), in which case customizing the scale per-consumer is not needed - if they don't know about each other, their scales will be different, and then they can't meaningfully share a limiter afaict?

Menduist:
> if they don't know about each other, their scales will be different, and then they can't meaningfully share a limiter afaict?

I think they can, because you can always express a cost as a percent of a budget, so you can scale different costs similarly
My current wording is maybe not clear enough, maybe it's clearer this way:
```nim
let gossipSubCost = operationsPerSecond(10) # Allow 10 OPS for gossipsub
let pingCost = operationsPerSecond(30) # Allow 30 OPS for ping
proc syncCost(bytesUsed: int): auto = tokensPerSecond(bytesUsed, 1_000_000) # Allow 1_000_000 bytes / second
```

You can now share the same bucket:
```nim
let bucket = RateCounter.init(cap=1.seconds)

# we received a GS connection
bucket.consume(gossipSubCost)

# we received a ping
bucket.consume(pingCost)

# we received a sync request
bucket.consume(syncCost(response.len))
```

arnetheduck:
> cost as a percent of a budget

"percent" is a scale too :) the rate limiter could prescribe that rate budgets have to be expressed in percentages, but that doesn't make it any more meaningful or "shareable": a bytes-per-second rate and a cpu-ticks rate or an i/o disk rate have nothing in common, and expressing them as "fractions" only complicates usage of the rate limiter - I think this is generally why the classic literature on the topic expresses rate limiting as generic tokens being "produced" at a given rate, with the bucket being able to hold up to a certain amount and consumers being able to take tokens out of it according to those parameters - it's a useful analogy that intuitively covers both bursts and overall average rate making it easy to reason about, and it is this analogy that I think is valuable to preserve in the wording of the API usage.

The "burstiness" in particular is important to tune for something like eth2 - we want consumers to generally be able to grab a few epochs of data without constraints, but once they hit that limit we want to go back to fairness overall.

Also, I think it's common to run multiple overlapping rate limiters for a single resource, as far as use cases go: you typically want to have a per-connection limit as well as a group and/or global limit for bandwidth for example: we want to allow a total of 100 blocks / sec to be uploaded to peers (to limit total bandwidth consumption by the req/resp system) but we also want to limit individual peers such that one doesn't monopolise that pool.

I can't think of many use cases where I would want to share rate limiters and why this is a primary use case which should drive the API design - what would be common examples of such sharing?

The discussion is a little bit analogous to peer scoring: there doesn't exist a globally meaningful measure of "score" for peers - rather, the score encodes domain-specific knowledge and while it would be possible to constrain it to a range, doing so would make it less intuitive / expressive and harder to reason about while not providing much benefit.

tests/testall.nim
```diff
@@ -8,4 +8,4 @@
 import testmacro, testsync, testsoon, testtime, testfut, testsignal,
        testaddress, testdatagram, teststream, testserver, testbugs, testnet,
        testasyncstream, testhttpserver, testshttpserver, testhttpclient
-import testutils
+import testutils, testratelimit
```
cheatfate, line 11: `testutils` MUST BE last imported module for a reason, its why it is actually moved to different `import` statement, so please move `testratelimit` to previous `import` statement.

chronos/ratelimit.nim
```diff
@@ -0,0 +1,109 @@
...
+
+  rc.lastUpdate = currentTime
+  rc.budget = min(rc.budgetCap, rc.budget + timeDelta)
+
+proc operationsPerSecond*(ops: float): RateCost =
+  ## Create a RateCost from an allowed number of 
+  ## operations per second
+  const budgetPerSecond = 1.seconds.nanoseconds.float
+
+  RateCost(budgetPerSecond / ops)
```
cheatfate, line 37: What the reason to perform conversion to `float` perform `float` division and convert result back to integer? I think number of operations SHOULD not be float because i dont see how count could be rational number.

chronos/ratelimit.nim
```diff
@@ -0,0 +1,109 @@
...
+  ## Create a RateCost from a token budget per second
+  runnableExamples:
+    let cost = tokensPerSecond(
+      1518, # packet of 1518 bytes
+      1000000 # budget of 1mb
+    )
+  let
+    budgetUsed = tokens / budget
+    allowedPerSeconds = 1.float / budgetUsed
+  operationsPerSecond(allowedPerSeconds)
```
cheatfate, line 49: I dont see any reasons to use here `floats` operations too.

tests/testall.nim
arnetheduck: what happens if it's not? this would be good to have in a comment, it's not obvious

cheatfate: `testutils` runs tests/checks for pending Futures. So if there some pending futures happens in previous tests this last test will signal it. If you put other test after `testutils` this test could just bypass it.

Menduist:
Closed in favor of #279

PR #265, opened 2022-03-30
Restore HTTP auth test

Menduist:
This test is failing, I've skipped it on master to avoid blocking the CI (here and Nim)

The server used for this test (https://guest:guest@jigsaw.w3.org/HTTP/Basic/) is returning a TLS error 40 when we try to connect to it from chronos (works fine with `curl`)

Wireshark doesn't seem to dislike the Client Hello we send, so it may be a bug on the other side

For reference, the curl hello on the left, and the bearssl one on the right:
image


Menduist:
Update, I found another test server which was working correctly (https://httpbin.org/basic-auth/user/passwd), so switched to this one on master
We should probably investigate why the old one started to fail

cheatfate:
Fixed in #301

PR #269, opened 2022-04-17
nimble setup

zah:


arnetheduck:
hm, lock files for libraries remain problematic because there is no way to reason about what that means, and no way to signal that a library supports multiple versions of a dependency - it is the role of a "leaf" application to compose a compatible set of libraries into a reproducible product - instead, libraries should be describing supported version ranges in the .nimble file - committing these lock files sets a precedent that will turn into debt in the long term - if anything, they could live in the tests folder so as to serve as build instructions for the tests, but nothing more than that.

zah:
A lock file in a library is used to the library's CI. This solves the problems that Tanguy has been having in the nim-libp2p library where he invented his own flavour of lock files.

Indeed, the library signals what the supported version ranges of its dependencies are in the nimble file of the library. The lock file is placed at the root of the package because the existence of the tests folder in merely a convention. We cannot build standardized tooling around such conventions.

arnetheduck:
>  This solves the problems that Tanguy has been having in the nim-libp2p library where he invented his own flavour of lock files. 

there are two ways to solve that:
* version ranges in the nimble file - this information is critical over time because it's used to update the lock file in a compatible way when composing multiple libraries into a new application - it requires versioning in the projects it depends on
* treat tests as a separate leaf project

having a lock file for a library proper creates confusion as to what it means: what if nim-eth and nim-libp2p have different lock files? are they now incompatible? how do you compose them? how do you tell users your library does _not_ require the stuff in the lock file, but rather that this is a ci-only instruction? this will become a recurring topic of confusion in the general UX of working with nimble, because semantically, it doesn't make sense.

zah:
The educational cost required to ameliorate the potential confusion is much lower than the educational cost for teaching the users to create separate packages for the tests in the library. Libraries having lock files will be a familiar pattern for many users coming from other ecosystems. The claim that it doesn't make sense semantically is quite the exaggeration. Here is how yarn explains this to users for example:

https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/#libraries-vs-applications

arnetheduck:
> the existence of the tests folder in merely a convention. 

so is the `src` folder and similar structural conventions that nimble can or should impose - the better we are able exploit these conventions, the less "friction" there will be in the user experience for developers - ie they less they'll have to read manuals - simply because the structure enforces correctness. In this case, it's "odd" to talk about the lock file even as discussed in the yarn file - there, they make the case that everyone should be on latest version of everything always and in that world, "de facto" there is just one supported combination of a library, and therefore there is not much harm in committing the lock file.

On a related topic, I'd be very interested to see minimal versioning in nimble as a version selection strategy - ie choose the lowest compatible version instead of the highest, when making a bump, and optimising for not changing versions except when explicitly forced or asked to - this would, over time, lead to less version churn and more predictable behavior, but also promote quality in libraries overall, with respect to compatibility - when I bump libp2p, I don't want the tool to make a chronos bump in _as well_ in the same go. 

The other fun thing to see would be tooling that automates the discovery of mutually compatible versions by running the test suite on the matrix of possible dependency versions ... crazy? perhaps, but so was google maps when it launched.

PR #275, opened 2022-05-24
AsyncEventQueue (AsyncEventBus replacement).

cheatfate:
Because its impossible to achieve goals which was initially set for `AsyncEventBus`. So this new structure will become as replacement for `AsyncEventBus`.


arnetheduck (commented):
So, as I see it, the principal difference between AsyncEventBus and AsyncEventQueue is that the former gets a single piece of data to users whereas the latter gets many. 

Both are useful - ie it's often the case that one wants to notify some listeners about a change, but only the latest version of the change is interesting (ie a counter that changed and has to be displayed on a screen - if it changed many times before it's displayed, the latest value should be "read" in the event - is that the current semantic of `AsyncEventBus`? If yes, I'd think we should keep both primitives rather than this being a replacement.

In terms of this queue, we'll need the option to set bounds on it so that we can deal with slow subscribers (by kicking them out, more or less).

chronos/asyncsync.nim
```diff
@@ -610,3 +623,131 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
...
+      resetFuture = false
+      ab.readers[index].waiter = nil
+
+    let reader = ab.readers[index]
+    doAssert(isNil(reader.waiter),
+             "Concurrent waits on same key are not allowed!")
+
+    let length = len(ab.queue) + ab.offset
+    doAssert(length >= ab.readers[index].offset)
+    if length == ab.readers[index].offset:
```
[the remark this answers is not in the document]
cheatfate: Nope, `offset` indicates from which point in the queue you should start reading events.

cheatfate:
`AsyncEventBus` is not flowed only when producer emits event with `emitWait()`. But this means that in our use-case processing queue could be blocked until consumer will not receive event with `waitEvent()`.

`AsyncEventBus` issue happens when producer emits events with `emit()`. Inside `emit()` consumers list is notified with first event and this consumers list is got emptied. So until consumer will not be scheduled all calls to `emit()` just discarding events...

So as `AsyncEventBus` it does not provide the core functionality, and its why it should be deprecated/removed.

chronos/asyncsync.nim
```diff
@@ -610,3 +623,131 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
...
+      reader.waiter.complete()
+  # This could generate leak, shrink() is not yet implemented for sequences.
+  ab.readers.setLen(0)
+  ab.queue.clear()
+
+proc closeWait*[T](ab: AsyncEventQueue[T]): Future[void] =
+  var retFuture = newFuture[void]("AsyncEventQueue.closeWait()")
+  proc continuation(udata: pointer) {.gcsafe.} =
+    if not(retFuture.finished()):
+      retFuture.complete()
```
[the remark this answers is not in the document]
cheatfate: Why? to raise `CancelledError` from `closeWait()` procedure?
The idea here is to ensure that all event consumers will receive notification about that `AsyncEventQueue` is get closed, and because all this callbacks get schedule inside `ab.close()` call, it means that `continuation` procedure will be scheduled after all the consumer notification callbacks will be already processed.

cheatfate:
Because every consumer has its own `offset` field, which tracks its position in the shared queue, its unlikely to be happened in such case. But on other side there could be very slow event's emitter.

chronos/asyncsync.nim
```diff
@@ -610,3 +626,162 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
...
+      else:
+        # Because ab.readers is sequence sorted by `offset`, we will apply our
+        # limit to the most recent consumer.
+        if ab.limit + (ab.readers[^1].offset - ab.offset) < len(ab.queue):
+          false
+        else:
+          true
+
+    if couldEmit:
+      ab.queue.addLast(data)
```
arnetheduck, line 702: ok, so here, for bounded queues, I think we need to clean up readers that are out-of-bounds / slow / stopped reading and call `compact`, or the queue will grow as if it were unbounded - consider:

```
let eventQueue = newAsyncEventQueue[seq[int]]()
let
  fast = eventQueue.register()
  slow = eventQueue.register()
  slowF = eventQueue.waitEvents(slow)
for i in 0..<1000:
  let fastF = eventQueue.waitEvents(fast)
  eventQueue.emit(newSeq[int](100*1024*1024))
  discard waitFor fastF
  GC_fullCollect()

  debugEcho GC_getStatistics()
```

this program should never keep more than one entry in its data structure, but keeps allocating more and more memory because `waitEvents` will not run until the slow reader schedules itself (consider for example if it's stuck in a socket `write` which takes minutes to timeout)

arnetheduck, line 702: ok, so here, for bounded queues, I think we need to clean up readers that are out-of-bounds / slow / stopped reading and call `compact`, or the queue will grow as if it were unbounded - consider:

```
let eventQueue = newAsyncEventQueue[seq[int]]()
let
  fast = eventQueue.register()
  slow = eventQueue.register()
  slowF = eventQueue.waitEvents(slow)
for i in 0..<1000:
  let fastF = eventQueue.waitEvents(fast)
  eventQueue.emit(newSeq[int](100*1024*1024))
  discard waitFor fastF
  GC_fullCollect()

  debugEcho GC_getStatistics()
```

this program should never keep more than one entry in its data structure, but keeps allocating more and more memory because `waitEvents` will not run until the slow reader schedules itself (consider for example if it's stuck in a socket `write` which takes minutes to timeout)

chronos/asyncsync.nim
```diff
@@ -610,3 +626,162 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
...
+          else:
+            min(itemsInQueue, eventsCount - len(events))
+
+      for i in 0 ..< itemsCount:
+        events.add(ab.queue[itemsOffset + i])
+      ab.readers[index].offset += itemsCount
+
+      # Keep readers sequence sorted by `offset` field.
+      var slider = index
+      while (slider + 1 < len(ab.readers)) and
```
arnetheduck, line 777: does it ever happen that a reader is _not_ moved to the far back of this list?

chronos/asyncsync.nim
cheatfate: Your example do not use bounded queue, is it intentional?

cheatfate: I'm not sure i understand your example because only 2 items in this specific scenario will be added to the queue, all other items will be discarded.

chronos/asyncsync.nim
cheatfate: waitEvents(N) could be used to retrieve specific number of events, not all of them, in this case reader will not be moved to the end of list.

chronos/asyncsync.nim
arnetheduck: oops, old version of example - edited to adds bounds, memory usage still explodes

arnetheduck: according to the memory usage profile, it keeps growing - this makes sense I think: waitEvents is never run for `slow` meaning that `compact` is never called with the slow reader removed

cheatfate: 1. We can't block/stop/remove/cleanup queue on emitter side, because we do not control poll() callback scheduling, so we don't know if slow reader was able to do some processing or we just adding callbacks and slow reader is never scheduled yet. Your example exploits this knowledge, in your example slow reader never gets any events after first one. So slow reader `offset` is now `1` and fast reader offset is at `1000`. It means because slow reader doesn't yet know that its slow... To get this information it should call `waitEvents` and receive an exception. After receiving an exception slow reader must `unregister()` itself from the queue, and only after that 1000 events will be garbage collected.

chronos/asyncsync.nim
```diff
@@ -610,3 +627,193 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
...
+  if len(ab.readers) > 0:
+    # We enqueue `data` only if there active reader present.
+    var changesPresent = false
+    let couldEmit =
+      if ab.limit == 0:
+        true
+      else:
+        # Because ab.readers is sequence sorted by `offset`, we will apply our
+        # limit to the most recent consumer.
+        if ab.readerOverflow(ab.readers[^1]):
```
arnetheduck, line 715: what if `readerOverflow(reader) == true` but `reader.overflow == false`? shouldn't we set `reader.overflow = true` in that case?

arnetheduck (approved):

chronos/asyncsync.nim
```diff
@@ -610,3 +627,193 @@ template emitWait*[T](bus: AsyncEventBus, event: string,
   ## wait until all the subscribers/waiters will receive notification about
   ## event.
   emitWait(bus, event, data, getSrcLocation())
+
+proc `==`(a, b: EventQueueKey): bool {.borrow.}
+
+proc compact[T](ab: AsyncEventQueue[T]) =
```
arnetheduck, line 633: `[T]` is not strictly needed here 
```suggestion
proc compact(ab: AsyncEventQueue) =
```

chronos/asyncsync.nim
cheatfate: Fixed.

PR #276, opened 2022-05-25
Asyncproc

cheatfate:


chronos/asyncproc.nim
```diff
@@ -0,0 +1,1354 @@
...
+  let poptions = options + {AsyncProcessOption.EvalCommand}
+  let process = await startProcess(command, options = poptions)
+  let res =
+    try:
+      await process.waitForExit(timeout)
+    finally:
+      await process.closeWait()
+  return res
+
+proc execCommandEx*(command: string,
```
yyoncho, line 1324: can you add the ability to pass arguments to startProcess?

cheatfate: I'm not sure i understand your question what arguments you want to pass?

yyoncho: Like this: https://github.com/status-im/nim-chronos/commit/7bbc59f448b4da737a0905efe0b855d22deccf01

cheatfate: But `execCommandEx` will not support all the arguments for `startProcess`, so we need to create different subset of options for it.

cheatfate: Oh i see you want `arguments` not `options`, but execCommand is supposed to start shell commands, so you can combine your options to a command with `command & arguments.join(" ")`

arnetheduck: > arguments.join

this would introduce quoting / whitespace issues, no?

yyoncho: > But `execCommandEx` will not support all the arguments for `startProcess`, so we need to create different subset of options for it.

Is there any benefit from that? IMHO it looks like inconsistency in the API.

cheatfate: It doesnt look like inconsistency of the API, because `execCommand` is supposed to invoke shell and execute your command inside the shell, its why it is called `execCommand`. This API call has corresponding std function https://github.com/nim-lang/Nim/blob/devel/lib/pure/osproc.nim#L99

cheatfate: If you like to have arguments list and still execute your command via shell, you can use `startProcess(poEvalCommand)`.

cheatfate: > > arguments.join
> 
> this would introduce quoting / whitespace issues, no?

This should be reviewed and i left appropriate TODOs in the code.

yyoncho: > It doesnt look like inconsistency of the API, because `execCommand` is supposed to invoke shell and execute your command inside the shell, its why it is called `execCommand`. This API call has corresponding std function https://github.com/nim-lang/Nim/blob/devel/lib/pure/osproc.nim#L99

got it, thanks.

arnetheduck:
Superseded by #374 et al

PR #279, opened 2022-05-31
Token bucket

Menduist:
A regular token bucket

I added a manual replenish as @arnetheduck suggested

### DIFF
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -1,0 +1,73 @@
+#               Chronos Rate Limiter
+#            (c) Copyright 2022-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+
+{.push raises: [Defect].}
+
+import ../chronos
+import timer
+
+export timer
+
+type
+  TokenBucket* = object
+    budget: int
+    budgetCap: int
+    lastUpdate: Moment
+    fillPerMs: int
+
+proc update(bucket: var TokenBucket) =
+  let
+    currentTime = Moment.now()
+    timeDelta = milliseconds(currentTime - bucket.lastUpdate).int
+    replenished = bucket.fillPerMs * timeDelta
+
+  bucket.lastUpdate += timeDelta.milliseconds
+  bucket.budget = min(bucket.budgetCap, bucket.budget + replenished)
+
+proc tryConsume*(bucket: var TokenBucket, tokens: int): bool =
+  ## If `tokens` are available, consume them,
+  ## Otherwhise, return false.
+
+  bucket.update()
+
+  if bucket.budget >= tokens:
+    bucket.budget -= tokens
+    true
+  else:
+    false
+
+proc consume*(bucket: var TokenBucket, tokens: int): Future[void] =
+  ## Wait for `tokens` to be available, and consume them.
+  ##
+  ## `tokens` is not limited to the bucket cap
+  doAssert(bucket.fillPerMs > 0, "consume is only available on autofilled buckets")
+
+  if bucket.tryConsume(tokens):
+    result = newFuture[void]("TokenBucket.consume")
+    result.complete()
+    return result
+
+  bucket.budget -= tokens
+  let timeToZero = milliseconds((-bucket.budget) div bucket.fillPerMs)
+  return sleepAsync(timeToZero)
+
+proc replenish*(bucket: var TokenBucket, tokens: int) =
+  ## Add `tokens` to the budget (capped to the bucket capacity)
+  bucket.budget += tokens
+  bucket.update()
+
+proc init*(
+  T: type[TokenBucket],
+  budgetCap, fillPerMs: int): T =
+  ## Create a TokenBucket
+  T(
+    budget: budgetCap,
+    budgetCap: budgetCap,
+    fillPerMs: fillPerMs,
+    lastUpdate: Moment.now()
+  )
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -17,6 +17,7 @@ type
   BucketWaiter = object
     future: Future[void]
     value: int
+    alreadyConsumed: int
 
   TokenBucket* = ref object
     budget: int
@@ -53,17 +54,22 @@ proc worker(bucket: TokenBucket) {.async.} =
 
   while bucket.pendingRequests.len > 0:
     bucket.manuallyReplenished.clear()
-    let waiter = bucket.pendingRequests[0]
+    template waiter: untyped = bucket.pendingRequests[0]
 
     if bucket.tryConsume(waiter.value):
       waiter.future.complete()
       bucket.pendingRequests.delete(0)
       continue
 
+    waiter.value -= bucket.budget
+    waiter.alreadyConsumed.inc(bucket.budget)
+    bucket.budget = 0
+
     let eventWaiter = bucket.manuallyReplenished.wait()
     if bucket.fillPerMs > 0:
       let
-        timeToZero = ((waiter.value - bucket.budget) div bucket.fillPerMs) + 1
+        nextCycleValue = min(waiter.value, bucket.budgetCap)
+        timeToZero = (nextCycleValue div bucket.fillPerMs) + 1
         sleeper = sleepAsync(milliseconds(timeToZero))
       await sleeper or eventWaiter
       sleeper.cancel()
@@ -87,6 +93,7 @@ proc consume*(bucket: TokenBucket, tokens: int): Future[void] =
   proc cancellation(udata: pointer) =
     for index in 0..<bucket.pendingRequests.len:
       if bucket.pendingRequests[index].future == retFuture:
+        bucket.budget.inc(bucket.pendingRequests[index].alreadyConsumed)
         bucket.pendingRequests.delete(index)
         if index == 0:
           bucket.manuallyReplenished.fire()
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -23,18 +23,27 @@ type
     budget: int
     budgetCap: int
     lastUpdate: Moment
-    fillPerMs: int
+    fillDuration: Duration
     workFuture: Future[void]
     pendingRequests: seq[BucketWaiter]
     manuallyReplenished: AsyncEvent
 
 proc update(bucket: TokenBucket) =
+  if bucket.fillDuration.milliseconds == 0:
+    bucket.budget = min(bucket.budgetCap, bucket.budget)
+    return
+
   let
     currentTime = Moment.now()
-    timeDelta = milliseconds(currentTime - bucket.lastUpdate).int
-    replenished = bucket.fillPerMs * timeDelta
+    timeDelta = currentTime - bucket.lastUpdate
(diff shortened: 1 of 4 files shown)


Menduist:
Ok, for reference a similar implementation is already used in nimbus: https://github.com/status-im/nimbus-eth2/blob/f124f22f108f64416f890297560cae8e6c136d08/beacon_chain/networking/eth2_network.nim#L428

I'll switch here to a queue of waiting futures, which will allow a single timer per bucket

Menduist:
Pushed new version.

Should we cap the number of `pendingRequests`, and throw an exception on `consume` if that's too big?

chronos/ratelimit.nim
```diff
@@ -0,0 +1,120 @@
...
+  bucket.update()
+
+  if bucket.budget >= tokens:
+    bucket.budget -= tokens
+    true
+  else:
+    false
+
+proc worker(bucket: TokenBucket) {.async.} =
+  defer: bucket.workFuture = nil
```
cheatfate, line 53: Usage of `defer` and `while/continue` inside `{.async.}` procedure is unsafe.

chronos/ratelimit.nim
```diff
@@ -0,0 +1,119 @@
...
+
+proc replenish*(bucket: TokenBucket, tokens: int) =
+  ## Add `tokens` to the budget (capped to the bucket capacity)
+  bucket.budget += tokens
+  bucket.update()
+  bucket.manuallyReplenished.fire()
+
+proc new*(
+  T: type[TokenBucket],
+  budgetCap, fillPerMs: int): T =
```
arnetheduck, line 111: this fill rate should probably be per second (SI units), and as such, a float, in the public api at least

chronos/ratelimit.nim
```diff
@@ -0,0 +1,119 @@
...
+    pendingRequests: seq[BucketWaiter]
+    manuallyReplenished: AsyncEvent
+
+proc update(bucket: TokenBucket) =
+  let
+    currentTime = Moment.now()
+    timeDelta = milliseconds(currentTime - bucket.lastUpdate).int
+    replenished = bucket.fillPerMs * timeDelta
+
+  bucket.lastUpdate += timeDelta.milliseconds
```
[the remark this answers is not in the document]
arnetheduck: or .. right, if it's rounded to mills above, needs rounding here too .. hm.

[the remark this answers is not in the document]
arnetheduck: worth a comment

chronos/ratelimit.nim
```diff
@@ -0,0 +1,119 @@
...
+  while bucket.pendingRequests.len > 0:
+    bucket.manuallyReplenished.clear()
+    template waiter: untyped = bucket.pendingRequests[0]
+
+    if bucket.tryConsume(waiter.value):
+      waiter.future.complete()
+      bucket.pendingRequests.delete(0)
+    else:
+      waiter.value -= bucket.budget
+      waiter.alreadyConsumed.inc(bucket.budget)
```
arnetheduck, line 62:
```suggestion
      waiter.alreadyConsumed += bucket.budget
```
same style?

chronos/ratelimit.nim
Menduist: If we expose a float and use a int per ms internally, it may be surprising to user since there will be rounding errors (ie, anything less than 1 / ms will be rounded to 0, so can't use 10 per second)
And can't remember why, but a design goal was to avoid floats internally

Menduist: So we either switch to float internally, (would need to remember why that wasn't done in the first place), or we switch to seconds internally, which would make the bucket update only once per second

arnetheduck: > rounded to 0

only as implemented presently

Menduist: Or `assert fillPerSecond mod 1000 == 0` :D

c-blake: It might be that some operating systems (like Linux) can (potentially) do faster context switches (no save & restore of FP context) until you first dirty floating point registers.  Once you do that the OS sets a flag and always does it.

This may be less relevant in the modern era of SIMD-optimized string libraries since almost all code does (string OR float) stuff.  OTOH, maybe on x86 it is only the FP stack (in|ex)cluded, and that might be a smaller % of state to save&restore anyway with all the (SSE/AVX)* registers.  So, it could also be less of an optimization these days.

It would be good to point to some document somewhere on the inet elaborating on this (and trying, hah!, to measure it).  Whether to commit to float in otherwise context switch heavy code like networking certainly comes up a lot.

arnetheduck: anyway, we could even stick with `int` and still use seconds in init - this is after all just a scaling factor.. internally the library can keep using `millis` or even `micros` for its computations as long as the "external" api is sane .. this is not really a real-time API - things like rounding and granularity are somewhat expected and as long as they're dealt with with reasonable attention to drift/bias, it seems somewhat arbitrary - so if I was using a fillPerMS of 1000 before, I'd use `1` now and if I want `0.5` I'd simply scale my entire token allocation by 10x and use `5`.

arnetheduck: alternative: turn this parameter into a `replenishTime: Duration`

Menduist: Switched to float internally, and all tests fail, ie
```
budget = 1000
tryConsume(800) == true
tryConsume(200) == false # rounding error
```
So I'll fix them by adding more margins, but still something to have in mind

> anyway, we could even stick with int and still use seconds in init

The issue is that currently, we can "tick" the bucket every millisecond, so we must be able to update its budget every millisecond. If you have fill rate per second = 1, what is the budget increase of 1ms supposed to be? 0.001, which requires a float budget

Which is why the two options in my eyes are: 1) use float internally 2) change the "tick" to second 3) keep fill in milliseconds
1. may indeed be the lesser evil, as long as we are careful not to add up errors over time

Menduist: Oh, maybe we could just scale everything to 1000* indeed

Menduist: Oh, maybe we could just scale everything to 1000* indeed

arnetheduck: `replenishTime` fixes rounding and allows arbitrary granularity internally (per-second granularity is probably not good enough ... millis or micros at least)

Menduist: Oh, github played games and hid your comment before I replied
Sounds good!

Menduist: Done, still requires some floatery, but it's at least contained & cannot drift

chronos/ratelimit.nim
```diff
@@ -0,0 +1,132 @@
...
+      (replenished.float / bucket.budgetCap.float))
+
+  bucket.lastUpdate += milliseconds(deltaFromReplenished)
+  bucket.budget = min(bucket.budgetCap, bucket.budget + replenished)
+
+proc tryConsume*(bucket: TokenBucket, tokens: int): bool =
+  ## If `tokens` are available, consume them,
+  ## Otherwhise, return false.
+
+  bucket.update()
```
arnetheduck, line 53: we can skip the update if there are still tokens

arnetheduck: (in the budget)

chronos/ratelimit.nim
```diff
@@ -0,0 +1,136 @@
...
+
+proc update(bucket: TokenBucket) =
+  if bucket.fillDuration.milliseconds == 0:
+    bucket.budget = min(bucket.budgetCap, bucket.budget)
+    return
+
+  let
+    currentTime = Moment.now()
+    timeDelta = currentTime - bucket.lastUpdate
+    fillPercent = timeDelta.milliseconds.float / bucket.fillDuration.milliseconds.float
```
arnetheduck, line 39: if we use nanosecond precision here, we can simply assign the lastUpdate - there should be enough precision in the float, allowing `deltaFromReplenished` to be removed

chronos/ratelimit.nim
```diff
@@ -0,0 +1,136 @@
...
+proc replenish*(bucket: TokenBucket, tokens: int) =
+  ## Add `tokens` to the budget (capped to the bucket capacity)
+  bucket.budget += tokens
+  bucket.update()
+  bucket.manuallyReplenished.fire()
+
+proc new*(
+  T: type[TokenBucket],
+  budgetCap: int,
+  fillDuration: Duration): T =
```
arnetheduck, line 126:
```suggestion
  fillDuration: Duration = 1.second): T =
```

chronos/ratelimit.nim
```diff
@@ -0,0 +1,136 @@
...
+      waiter.alreadyConsumed += bucket.budget
+      bucket.budget = 0
+
+      let eventWaiter = bucket.manuallyReplenished.wait()
+      if bucket.fillDuration.milliseconds > 0:
+        let
+          nextCycleValue = float(min(waiter.value, bucket.budgetCap))
+          budgetRatio = (nextCycleValue.float / bucket.budgetCap.float)
+          timeToZero = int(budgetRatio / bucket.fillDuration.milliseconds.float) + 1
+          sleeper = sleepAsync(milliseconds(timeToZero))
```
arnetheduck, line 84: let's document here in a TODO that creating many timers may negatively affect performance

chronos/ratelimit.nim
```diff
@@ -0,0 +1,136 @@
...
+  ## Add `tokens` to the budget (capped to the bucket capacity)
+  bucket.budget += tokens
+  bucket.update()
+  bucket.manuallyReplenished.fire()
+
+proc new*(
+  T: type[TokenBucket],
+  budgetCap: int,
+  fillDuration: Duration): T =
+  assert fillDuration.nanoseconds mod 1e6.int == 0
```
arnetheduck, line 127: this is not needed really

chronos/ratelimit.nim
```diff
@@ -0,0 +1,136 @@
...
+    budget: int
+    budgetCap: int
+    lastUpdate: Moment
+    fillDuration: Duration
+    workFuture: Future[void]
+    pendingRequests: seq[BucketWaiter]
+    manuallyReplenished: AsyncEvent
+
+proc update(bucket: TokenBucket) =
+  if bucket.fillDuration.milliseconds == 0:
```
arnetheduck, line 32: == default(Duration)

chronos/ratelimit.nim
Menduist: Not really, see the "Very long replenish" test
To fill by 1, we must wait 240ms, so if `update` is called before 240ms, it should not update `lastUpdate`, otherwise it may not refill properly
We could just `if replenished == 0: return`, but then it would be very imprecise for slow replenish / small buckets

arnetheduck: aaah, good point

Menduist:
Did a simple benchmark to see if we have low hanging fruits:
```nim
  import std/random
  proc zzz {.async.} =
    let tokenBucket = TokenBucket.new(1000)
    while true:
      let toConsume = rand(1000)
      await tokenBucket.consume(toConsume)

  for _ in 0..10000:
    asyncSpawn zzz()

  waitFor(sleepAsync(20.seconds))
```
callgrind:
image

The vast majority of the time is in the GC, with `sleepAsync` (and thus the timer sorting) taking only 1.27% of the time, that's seems really low, maybe I did something wrong

PR #281, opened 2022-06-02
Auto inject "except CancelledError"

Menduist:
I don't remember if this was already discussed, but I think this behavior is quite nice.
```nim
    proc example {.async.} =
      while true:
        try:
          await sleepAsync(1000.minutes)
        # will be injected automatically here:
        # except CancelledError as exc:
        #   raise exc
        except CatchableError:
          discard

      while true:
        try:
          await sleepAsync(1000.minutes)
        # Already handled, nothing to do here
        except CancelledError as exc:
          # could cleanup or whatever
          raise exc
```

For reference, in libp2p, about 90% of CancelledError handling is just bubbling it up the stack

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -41,6 +41,40 @@ proc processBody(node, retFutureSym: NimNode,
   of RoutineNodes-{nnkTemplateDef}:
     # skip all the nested procedure definitions
     return node
+  of nnkTryStmt:
+    var cancelledHandled = false
+    for tryBranch in 1 ..< node.len:
+      for child in node[tryBranch]:
+        case child.kind:
+        of nnkStmtList: break
+        of nnkInfix:
+          if child[1] == ident"CancelledError":
+            cancelledHandled = true
+            break
+        of nnkIdent:
+          if child == ident"CancelledError":
+            cancelledHandled = true
+            break
+        else: discard
+    
+    if not cancelledHandled:
+      result = newNimNode(nnkTryStmt, node)
+      result.add(node[0])
+      result.add(nnkExceptBranch.newTree(
+        nnkInfix.newTree(
+          ident"as",
+          ident"CancelledError",
+          ident"exc"
+        ),
+        nnkStmtList.newTree(
+          nnkRaiseStmt.newTree(
+            ident"exc"
+          )
+        )
+      ))
+      for tryBranch in 1 ..< node.len:
+        result.add(node[tryBranch])
+
   else: discard
 
   for i in 0 ..< result.len:
--- a/tests/testmacro.nim
+++ b/tests/testmacro.nim
@@ -104,3 +104,26 @@ suite "Macro transformations test suite":
 
     macroAsync2(testMacro2, seq, Opt, Result, OpenObject, cstring)
     check waitFor(testMacro2()).len == 0
+
+  test "Insert CatchableError re-raise":
+    proc swallower {.async.} =
+      while true:
+        try:
+          await sleepAsync(1000.minutes)
+        except CatchableError: discard
+        except Defect as exc: discard
+        except: discard
+        
+
+    var tester = false
+    proc handler {.async.} =
+      while true:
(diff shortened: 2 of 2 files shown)


cheatfate:
If your pattern of handling exceptions is `except CatchableError` then of course maybe you need to inject it, but you can do it in your own project. Because in 90% of cases you don't need to handle `CancelledError`. You only need to handle it if you allocating resources which could not be freed automatically in case of cancellation. So this "automatic" approach will just introduce more issues.

Menduist:
It's not "my pattern to handle exceptions"

https://github.com/status-im/nim-chronos/blob/1233f8fb4b6eeb332febdda6c1862360e23d00f2/chronos/streams/asyncstream.nim
This file alone contains 16
```nim
    except CancelledError as exc:
      raise exc
```

Menduist:
Most of the time when we handle exceptions with chronos, we'll cast a wide net (with CatchableError for instance), since without exception tracking, we don't know what will come of it.
But since with CatchableError, you'll need to re-raise manually the CancelledError, it becomes a very common pattern

If you forget to do it, you'll swallow cancellation, which is not good

This PR will:
A) allow us to remove a lot of boilerplate present all around our code base
B) make sure that no one swallows cancellations inadvertedly

arnetheduck:
> Most of the time when we handle exceptions with chronos, we'll cast a wide net (with CatchableError for instance), since without exception tracking, we don't know what will come of it.

since the idea is to move forwards with raises tracking, this argument doesn't really apply - ie it doesn't represent a state that we want to be in, in the future

> If you forget to do it,

better would be a warning during compilation that you forgot it - this has several positive outcomes:

* it rewards those that catch more specific exceptions
* it rewards those that use the new raises tracking to make the former point safe
* it highlights where the handlers are missing
* it doesn't introduce magic surprising behavior

Menduist:
> ie it doesn't represent a state that we want to be in, in the future

Sure, but realistically, migrating the entire code base to raise tracking will take months / years. I think that improving this in the meantime will not be lost

> it doesn't introduce magic surprising behavior

I would argue that the CancelledError is already a magic surprising behavior: you're never raising it yourself, and yet it pops up
This PR would make it a bit more hidden (you would only have to handle it when you want to do specific things on cancellation, which is pretty rare). And with something like https://github.com/status-im/nim-chronos/pull/254, you would actually never have to handle it at all, which would make it an implementation detail, hidden from the user

But I get that it could be "too magical", maybe a warning would make more sense

arnetheduck:
> Sure, but realistically, migrating the entire code base to raise tracking will take months / years. I think that improving this in the meantime will not be lost

it's equally messy to migrate to this new magic behavior, specially considering that it's a behavior that doesn't have that much value once the other things are in place. also, it'll take exactly as long as we want it to, based on how valuable it is.

> yet it pops up

hiding it further is not the way to go though - quite the opposite: it should be visible and explicit, except where it's safe to hide it. hiding it out of general laziness or the desire not to think about cancellation has the same end result as not thinking about error handling cancellation is something that all code inherently must consider when using chronos, at all times - specially when writing libraries, it's easy to forget because the library itself does not initiate the cancellation.

The core reason for keeping things visible here is an asymmetry in outcomes: when you have to write an extra `catch/raise exc`, that's a small aesthetic  issue - when you forget to write it, it's a bug that is incredibly difficult to track down (being resource leaks, table entries that shouldn't be there etc), specially in the presence of catch-all exception handlers that simply drop it.

Menduist:
So I opened a branch which does the warning, but it's very noisy: https://github.com/status-im/nim-chronos/tree/warnmissingcatch
On libp2p, I get about 30 false positives (stuff that's going to get asyncSpawned, stuff where handling the cancellation is similar to a regular CatchableError, or stuff where await is not called in the try body)

Still useful to have around, caught a few missing handlers in libp2p: https://github.com/status-im/nim-libp2p/pull/733

But probably to noisy to keep around all the time, maybe I'll open a PR with the warning behind a flag

PR #286, opened 2022-06-16
asyncTest

cheatfate:


chronos/unittest2/asynctools.nim
```diff
@@ -0,0 +1,19 @@
...
+#             (c) Copyright 2022-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+import unittest2
+import ".."/asyncloop
+
+export unittest2
```
arnetheduck, line 12: we can export chronos from here also

arnetheduck (approved):

PR #296, opened 2022-07-19
make it more Nim 1.4+ compatible

narimiran:
A continuation of https://github.com/status-im/nimbus-eth2/pull/3888

Rationale (from that link):

> Currently (the latest commit in unstable branch), the CI log for the 1.6 branch has **107.000 lines**.
Removing every hint about cannot raise 'Defect' (including the ones in the vendor directory) leaves **16.000 lines**. 85% off.

### DIFF
--- a/chronos/apps/http/httptable.nim
+++ b/chronos/apps/http/httptable.nim
@@ -10,7 +10,10 @@
 import std/[tables, strutils]
 import stew/base10
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
 
 type
   HttpTable* = object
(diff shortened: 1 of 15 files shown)


narimiran:
Anything stopping this from being merged?

Menduist (changes requested):
Could you also fix
https://github.com/status-im/nim-chronos/blob/master/chronos/asyncmacro2.nim#L184-L200
(have already been fixed [here](https://github.com/status-im/nim-chronos/pull/251/files#diff-c1e1958e958f530246209b2176ae811a267b4c8d9ee8ffeac0fadf4a6abab460R242-R252), you can do something similar)

& asyncfutures2

Currently, every async procedure is causing 2 "XCannotRaiseY" warnings, which is not very viable

narimiran:
> Could you also fix

Can that be part of a follow-up PR?
Maybe together with some fixes for `UnusedImport`, `CStringConv`, etc.?

Menduist (approved):
sure

PR #297, opened 2022-07-22
`results.?` compatibity

arnetheduck:
This PR uses the facilities provided in
https://github.com/status-im/nim-stew/pull/134 to support `?` in `async`
proc's.

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -156,6 +156,13 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
 
       procBody = prc.body.processBody(castFutureSym, baseType)
 
+      # Support `let x = ? await ...`
+      assignResultSym = ident "assignResult?"
+      assignResult = quote do:
+        when declared(Result):
+          template `assignResultSym`(v: Result) {.used.} =
+            complete(`castFutureSym`, v)
+
     # don't do anything with forward bodies (empty)
     if procBody.kind != nnkEmpty:
       let
@@ -185,7 +192,7 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
 
         completeDecl = completeWithNode(castFutureSym, baseType, procBodyBlck)
 
-        closureBody = newStmtList(resultDecl, completeDecl)
+        closureBody = newStmtList(assignResult, resultDecl, completeDecl)
 
         internalFutureParameter = nnkIdentDefs.newTree(
           internalFutureSym, newIdentNode("FutureBase"), newEmptyNode())
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -8,7 +8,7 @@ license       = "MIT or Apache License 2.0"
 skipDirs      = @["tests"]
 
 requires "nim >= 1.2.0",
-         "stew",
+         "stew#assign-result",
          "bearssl",
          "httputils",
          "unittest2"
--- a/tests/testmacro.nim
+++ b/tests/testmacro.nim
@@ -94,6 +94,20 @@ proc testAwaitne(): Future[bool] {.async.} =
 
   return true
 
+proc resultOk(): Future[Result[int, string]] {.async.} =
+  return ok(42)
+
+proc resultErr(): Future[Result[int, string]] {.async.} =
+  return err("string")
+
+proc testResult: Future[Result[int, string]] {.async.} =
+  let
+    x = ? await resultOk()
+
+  if x == 42:
+    let _ = ? await resultErr()
+  return err("not this one")
+
 suite "Macro transformations test suite":
   test "`await` command test":
     check waitFor(testAwait()) == true
@@ -226,3 +240,18 @@ suite "Closure iterator's exception transformation issues":
 
     waitFor(x())
 
+suite "Result integration":
+  test "question mark":
+    # generics are tricky and buggy, test them more!
+    proc someFunc2[T](v: T): Future[Result[T, string]] {.async.} =
+      return ok(v)
+    proc someFunc[T](v: T): Future[Result[T, string]] {.async.} =
+      let tmp = ? await someFunc2(v)
+      return ok(tmp + 2)
+
+    proc caller(v: int): Future[Result[int, string]] {.async.} =
+      return ok(3 + ? await someFunc(v))
+
+    check waitFor(testResult()).error() == "string"
+
+    check waitFor(caller(42))[] == 42 + 2 + 3


chronos/asyncmacro2.nim
```diff
@@ -316,5 +325,5 @@ macro async*(prc: untyped): untyped =
       result.add asyncSingleProc(oneProc)
   else:
     result = asyncSingleProc(prc)
-  when defined(nimDumpAsync):
+  when true or defined(nimDumpAsync):
```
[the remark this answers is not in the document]
arnetheduck: updated

chronos/asyncmacro2.nim
```diff
@@ -137,6 +140,12 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
     # fix #13899, `defer` should not escape its original scope
     procBody = newStmtList(newTree(nnkBlockStmt, newEmptyNode(), procBody))
 
+    if returnTypeIsResult:
+      # Support `let x = ? await ...`, but only if `Result` exists as a symbol
+      procBody.insert 0, quote do:
+        template assignResult(v: Result) =
```
zah, line 146: Instead of enabling this only in procs that return `Result`, I think it would be better to enable it always, but to use a name that is less likely to clash with user-defined symbols. You can use an identifier such as `#asignResult` for example.

zah: This would allow third party libraries to plug into the same mechanism.

Furthermore, it enables the creation of templates that create control-flow abstractions in async procs. `?` is just one example.

zah: Another alternative is to call it just `asyncReturn` and to include the `return` statement in it. This would be the most convenient option for authoring templates like the ones that I mentioned. Many input validation patterns in the REST API can be abstracted away with such templates for example.

arnetheduck: > Many input validation patterns in the REST API can be abstracted away with such templates for example.

this sounds like overloading the mechanism for an entirely different purpose, and as such would probably require a dedicated template - `?` works only because it's part of the "generally known control flow vocabulary" like `break` - having random custom control flow hidden in templates sounds like something that might overobfuscate the code instead (something we've avoided pretty much everywhere).. can you provide a more specific code example of how this would look?

zah: There are plenty of reasonable use-cases where you might want to hide a control-flow statement behind a template. Consider this authorization check that must be repeated at the beginning of every request in the Keymanager API:

https://github.com/status-im/nimbus-eth2/blob/eb6b7affee4264e4c4b642a153289ce9f439c65c/beacon_chain/rpc/rest_key_management_api.nim#L117-L120

Extracting it in a template called `checkAuthorization` or `return40xIfNotAuthorized` is perfectly reasonable way to improve the factoring of the code.

Here is another example, featuring input validation logic for the `ValidatorIndex` values which is duplicated in many of the REST handlers:

https://github.com/status-im/nimbus-eth2/blob/eb6b7affee4264e4c4b642a153289ce9f439c65c/beacon_chain/rpc/rest_beacon_api.nim#L252-L259

arnetheduck: > Extracting it in a template called checkAuthorization or 

`? checkAuthorization` would be the canonical way to do that, indeed to avoid hidden control flow in templates - with `Result`, we separate the concerns of "is this an error or not" then use that separation to create logic based on this separation - this way, when reading the code, it's clear that there is a control-flow point on that line - it's the same argument as why we don't want to use exceptions: they hide control flow inside the callee, so when reading code it's hidden.

> input validation logic for the `ValidatorIndex`

```
                  let vres = item.index.toValidatorIndex().valueOr:
                    return mapValidatorIndex(error)
````

Comment (zah) [reply chronos/asyncmacro2.nim:146]: `? checkAuthorization` is not possible, because REST API errors are mapped to error responses, not to error values of the `Result` type.

Chronos should be a general-purpose library in my opinion and its design should lean towards enabling people to build abstractions over it rather than guiding them into specific usage patterns. Many developers would find it sufficient to name their templates appropriately to reveal the hidden control flow inside them (e.g. `returnIfX`). Others would be perfectly happy with coding styles that resemble the semantics of exceptions-based error handling.

Comment (arnetheduck) [reply chronos/asyncmacro2.nim:146]: Well, that's a design issue in presto above all - as such, the use-cases you highlight sound more like they should have a dedicated mechanism in presto - also because if `?` hijacks this mechanism in chronos, there's no "trivial" way to chain "handlers" or converters at this level - one template cannot call the "next" template easily, to install a chain of handlers.

It would be reasonable that presto itself uses a mechanism like this for some dedicated purpose, but I'm not sure what that would be.

I'm not strongly opposed to making this not tied to Result, but still, the use cases presented give little guidance as to how it would work or how to balance the feature because they overload the mechanism for a completely different purpose at a much higher abstraction level, and doing this in general leads to a poorly defined low-level feature.

Comment (arnetheduck) [reply chronos/asyncmacro2.nim:146]: ie another way to look at this: we can do a more constrained mechanism now - `Result` is a fully working piece of code that needs this mechanism today - when we have another strong use case for ii, we can relax it, but without a concrete use where it actually solves a real problem, the increased generality is more likely to miss the target.

Comment (zah) [reply chronos/asyncmacro2.nim:146]: I think I've seen this argument in the past. Here, you've argued that the `either` utility from the error handling proposal is not useful:
https://github.com/status-im/nim-stew/pull/26#discussion_r409364304

Many months later, you've refactored hundreds of call sites to use `valueOr` (which practically the same thing).

I think the current situation is quite simple - the `async` pragma brakes a basic feature of Nim's templates. Everybody has to learn the hard way that `return` doesn't work in templates when they are used in async procs. Introducing `asyncReturn` is way to gain back part of what was lost. It's really as simple as that.

**Comment by Menduist** [comment]
Besides the alternative I proposed [here](https://github.com/status-im/nim-chronos/issues/217#issuecomment-1009898048) and will test out when I find some time, a cooler solution would be to open an RFC on the Nim specs to allow:
```nim
proc hey: string =
  template `return`(i: int) = return $i
  return 5
```
to work. It could replace the whole "rewritting return" sheningan, fix the issue with stew in a nice generic way, and doesn't seem stupid to me

Do you think it would make sense to open such RFC? (Obviously, it would take a long time, so we'll need a temporary solution anyway)

chronos/asyncmacro2.nim
arnetheduck: > There are plenty of reasonable use-cases

I've considered these points a bit more and:

* I would wait to design the feature until more clarity on such use cases exist - ie a concrete example to build off, so that we don't build ourselves into a corner with something we don't want to support down the line, considering generic sandwitch problems etc - the `Result` use case is already well understood from other uses of `?`
* The use cases presented could be considered `chronos` extension points too - ie instead of assigning to the future, `chronos` could offer a similar extension point to `?` for such cases where someone wants to extend chronos to support _their_ extension point.
  * Result is already a dependency of `chronos` so putting this statement here does not create additional dependency constraints or hierarchy - it's natural that chronos "supports" result extension points this way

zah: Well, I have already provided few very specific motivating examples here:
https://github.com/status-im/nim-chronos/pull/297#discussion_r927587294

What would constitute "more clarity"? 

Can you motivate in any way the inconsistency in the language that is being created here - using a `return` statement within templates is allowed in general, but it's not possible to use inside `async` procs. You seem to be arguing that Nim should not allow `return` in templates at all, but this ship has already sailed.

arnetheduck: > What would constitute "more clarity"?

code.

arnetheduck: >  it's not possible to use inside async procs. 

this is perhaps the core of the problem: to enable this / make it work, we're reaching for more and more convoluted constructs whose ability to scale in a composable way is .. dubious.

ie your own surprise that `if compiles(templateHere())` works is a hint that what we're doing to achieve this goal is .. dubious and perhaps short-sighted - that it doesn't work in 2.0 is another hint that it's perhaps ill advised - I'm concerned that this will turn into non-composability along an axis if we overload the mechanism, so before venturing further in this direction at all, it stands to reason to examine it at smaller scale.

arnetheduck: > here: https://github.com/status-im/nim-chronos/pull/297#discussion_r927587294

the question is: is increasing the complexity of interaction between components the right solution here - or is decreasing the complexity of presto etc better? the success in combining a particular set of tools lies in using the tools efficiently together: some tools simply aren't good matches for each other - the same way with the authorization example: if you're working against the offered framework, you run into friction and you might be able to force your way a bit further but sometimes, it's the framework that needs changing..

zah: > ie your own surprise that if compiles(templateHere()) works is a hint that what we're doing to achieve this goal is .. dubious and perhaps short-sighted - that it doesn't work in 2.0 is another hint that it's perhaps ill advised - I'm concerned that this will turn into non-composability along an axis if we overload the mechanism, so before venturing further in this direction at all, it stands to reason to examine it at smaller scale.

The more general solution seems to me equal in complexity with the current solution (i.e. it would exercise the same sub-systems within the compiler).

arnetheduck:
ci fails are due to nimble, not the changes

arnetheduck:
https://github.com/arnetheduck/nim-results/pull/37 is one more entry in the `?` compatibility line which works out-of-the-box now that #449 is merged

arnetheduck:
obsoleted by https://github.com/arnetheduck/nim-results/pull/37

PR #298, opened 2022-07-22
Raise tracking support for chronos procedures

Menduist:
Follow up on #251


chronos/asyncfutures2.nim
```diff
@@ -613,18 +613,23 @@ macro checkFutureExceptions*(f, typ: typed): untyped =
     quote do: raiseAssert("Unhandled future exception: " & `f`.error.msg)
   )
 
-proc read*[T](future: Future[T] ): T {.
-     raises: [Defect, CatchableError].} =
+# we have to rely on inference here
+{.pop.}
+proc read*(future: Future | RaiseTrackingFuture ): auto =
```
arnetheduck, line 618: just make a separate read function for RTF? the `auto` return type here is unnecessarily wide as well, there's too many compromises for too little benefit

chronos/asyncfutures2.nim
```diff
@@ -931,13 +946,13 @@ proc oneValue*[T](futs: varargs[Future[T]]): Future[T] {.
 
   return retFuture
 
-proc cancelAndWait*[T](fut: Future[T]): Future[void] =
+proc cancelAndWait*[T](fut: Future[T]): RaiseTrackingFuture[void, (CancelledError,)] =
```
arnetheduck, line 949:
```suggestion
proc cancelAndWait*[T](fut: Future[T]): Future[void] {.asyncraises:[(CancelledError,)] =
```
the rewrite to RTF should be macro-handled, no?

chronos/asyncloop.nim
```diff
@@ -1100,6 +1102,15 @@ proc wait*[T](fut: Future[T], timeout = InfiniteDuration): Future[T] =
 
   return retFuture
 
+proc wait*[T, E](fut: RaiseTrackingFuture[T, E], timeout = InfiniteDuration): auto =
```
arnetheduck, line 1105: what's the return type here? in public functions, we ideally want to avoid auto both for documentation reasons and because it serves as a type inference guide for the compiler.

chronos/asyncfutures2.nim
Menduist: We don't have access to macros here (asyncmacro2 requires asyncfutures2)

chronos/asyncfutures2.nim
Menduist: Indeed, we can avoid it here by making a specialized read for RTF

chronos/asyncloop.nim
Menduist: It's RTF[T, E + CanceledError] (which I just realized is wrong, should be AsyncTimeoutError)

I don't know a better solution than auto for this kind of procedures

chronos/asyncfutures2.nim
arnetheduck: maybe we should move this one elsewhere then - ie `cancelAndWait` is a composite "utility" operation much like `allFutures` etc - then we keep "primitives" in asyncfuture2 and maybe one fine day we can get rid of include altogether

arnetheduck:
superseded by #454

PR #299, opened 2022-07-29
Optimized implementation for preferredContentType() 

cheatfate:
This is more optimized version for preferredContentType() which deprecates helpers usage and sorting of Accept header values array.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -467,9 +467,10 @@ proc preferredContentMediaType*(acceptHeader: string): MediaType =
       MediaType.init("*", "*")
 
 proc preferredContentType*(acceptHeader: string,
-                           types: varargs[MediaType]): Result[MediaType, cstring] =
-  ## Match or obtain preferred content-type using ``Accept`` header specified by
-  ## string ``acceptHeader``.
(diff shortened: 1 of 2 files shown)


chronos/apps/http/httpserver.nim
```diff
@@ -506,15 +521,55 @@ proc preferredContentType*(acceptHeader: string,
...
+          let
+            preferredIndex = types.find(itemType.mediaType)
+            preferredWeight =
+              if preferredIndex == -1:
+                0.0
+              else:
+                # Calculate weight which depends on position in ``types``
+                # array. Because ``preferredIndex`` is always less than
+                # ``len(types)`` this weight could not be ``0.0``.
+                float(len(types) - preferredIndex) / float(100_000)
```
zah, line 547: These float divisions seems unnecessary to me. You can just make the comparison below that checks for a new best element to work as if it compares tuples: `(qvalue, preferredIndex)`. (you can either use a tuple directly, or you can code the logical equivalent of this tuple comparison).

cheatfate: Clients must not violate the spec and if client will supply something like `q=0.0001` it will be rejected at earlier phase.

PR #302, opened 2022-08-02
Use new Content-Type header value parser.

cheatfate:


chronos/apps/http/httpcommon.nim
```diff
@@ -193,16 +193,18 @@ func getContentEncoding*(ch: openArray[string]): HttpResult[
...
     ok(res)
 
-func getContentType*(ch: openArray[string]): HttpResult[string]  {.
+func getContentType*(ch: openArray[string]): HttpResult[ContentTypeData]  {.
      raises: [Defect].} =
   ## Check and prepare value of ``Content-Type`` header.
   if len(ch) == 0:
     err("No Content-Type values found")
   elif len(ch) > 1:
     err("Multiple Content-Type values found")
```
etan-status, line 202: Does the case where multiple header values are combined into a single header as a comma-separated list (e.g., by a proxy) need special treatment as well, or are these already split into multiple entries by the `ch` list creation logic?

RFC 7230:
>    A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.  The order
   in which header fields with the same field name are received is
   therefore significant to the interpretation of the combined field
   value; a proxy MUST NOT change the order of these field values when
   forwarding a message.

(with exception for `Set-Cookie` that cannot be combined)

etan-status (approved):

zah (approved):

chronos/apps/http/httpcommon.nim
cheatfate: How to decode document which has
```
Content-Type: application/json
Content-Type: application/octet-stream
```
?

etan-status: The reject is alright. I just wonder if this alternate syntax is also rejected:

```
Content-Type: application/json, application/octet-stream
```
(proxy may combine the two header lines into one comma-separated one except for `Set-Cookie`)

PR #303, opened 2022-08-07
getPreferredContentType() deprecate float math

cheatfate:


chronos/apps/http/httpserver.nim
```diff
@@ -470,6 +470,25 @@ proc preferredContentMediaType*(acceptHeader: string): MediaType =
     else:
       MediaType.init("*", "*")
 
+proc cmp*(a, b: tuple[qvalue: float, index: int]): int =
```
zah, line 473: I think `cmp` should already be defined for tuples in `system.nim`

PR #304, opened 2022-08-17
Add more FutureBase helpers.

cheatfate:


chronos/asyncfutures2.nim
```diff
@@ -845,7 +849,10 @@ proc cancelAndWait*[T](fut: Future[T]): Future[void] =
     fut.cancel()
   return retFuture
 
-proc allFutures*[T](futs: varargs[Future[T]]): Future[void] =
+proc cancelAndWait*[T](fut: Future[T]): Future[void] =
```
arnetheduck, line 852: why do we need this overload? the `FutureBase` overload should be used "automatically"

Issue #242, opened 2021-11-22
index error in readStreamLoop of transports/stream


michaelsbradleyjr:
I'm experiencing runtime crashes from unhandled exceptions re: nim-chronos, with [L1436](https://github.com/status-im/nim-chronos/blob/master/chronos/transports/stream.nim#L1436-L1437) in `transports/stream.nim` reported as the culprit:

```
/Users/michael/repos/nim-dagger/experiments/localstore.nim(117) localstore
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/asyncloop.nim(1091) waitFor
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/asyncloop.nim(279) poll
/Users/michael/repos/nim-dagger/experiments/localstore.nim(114) main
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/asyncloop.nim(279) poll
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/transports/stream.nim(1436) readStreamLoop
/Users/michael/.choosenim/toolchains/nim-1.6.0/lib/system/fatal.nim(53) sysFatal
[[reraised from:
/Users/michael/repos/nim-dagger/experiments/localstore.nim(117) localstore
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/asyncloop.nim(1091) waitFor
/Users/michael/.nimble/pkgs/chronos-#export-selector-field/chronos/asyncloop.nim(279) poll
]]
Error: unhandled exception: index 4096 not in 0 .. 4095 [IndexDefect]
```

I've tried with Nim v1.2.14, v1.4.8, v1.6.0 — same thing regardless of version.

When I (naively) change L1435 from:
```nim
while true:
```
to:
```nim
while transp.offset < transp.buffer.len:
```
I no longer experience that runtime crash.

**Context**

I'm doing some early stage experiments combining use of [`task_runner` impl_beta_2 branch](https://github.com/status-im/nim-task-runner/tree/impl_beta2_nimble/)'s threadpool with chronos' http server and async pipes:

https://github.com/michaelsbradleyjr/nim-dagger/blob/localstore/experiments/localstore.nim#L77-L78

Re: those two lines in particular, I only experience the runtime crash (i.e. without change mentioned above) if L77 is enabled and L78 is disabled, i.e. only when the Future returned by `writer.write` is *not* discarded.

---

[...]

PR #309, opened 2022-09-05
Fix nested waitFor (IndexError defect crash) bug.

### LINKED ISSUE #308: `poll` reentrancy may trigger network callbacks multiple times
During `poll`, expired timers and pending network events are collected into `loop.callbacks`, callback functions waiting for the events are scheduled, and then also executed (`processCallbacks()` in `asyncloop.nim`). When such a callback then proceeds to call `waitFor`, it re-enters `poll` while `processCallbacks()` is already ongoing.

If such a reentrancy condition occurs, `loop.callbacks` may still have items left. If they correspond to pending network events that have not yet been processed, the recursive `poll` call will re-discover that those events are still pending (`loop.selector.selectInto`). This then leads to network event callbacks being scheduled multiple times.

Certain network event callbacks do not operate properly when the underlying network event reflects a condition that is no longer accurate. For example, in `transports/stream.nim`, the function `readStreamLoop` expects to be called only when the underlying `transp.buffer` has space available, and crashes with an `IndexError` when it is called while that buffer is already full (because `addr transp.buffer[transp.offset]` on `transp.offset == transp.buffer.len` is invalid).

```
vendor/nim-chronos/chronos/asyncloop.nim(1124) waitFor
vendor/nim-chronos/chronos/asyncloop.nim(286) poll
vendor/nim-chronos/chronos/transports/stream.nim(1376) readStreamLoop
vendor/nimbus-build-system/vendor/Nim/lib/system/chcks.nim(23) raiseIndexError2
vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(50) sysFatal
Unhandled exception: index 4096 not in 0 .. 4095 [IndexError]
```

This condition can be experienced, for example using `asyncTest` in network heavy tests that operate on large amounts of data. It occurs from time to time on CI for `nimbus-eth2`.

macOS is affected for sure (probably Unix in general). I have not reviewed the Windows implementation.

cheatfate:
* Add GetQueuedCompletionStatusEx() support for Windows, which allows to capture more then one event in single `poll()` call.

This PR will fix
#308 
#242

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -352,9 +352,9 @@ when defined(windows):
 
     AsyncFD* = distinct int
 
-  proc getModuleHandle(lpModuleName: WideCString): HANDLE {.
+  proc getModuleHandle(lpModuleName: WideCString): Handle {.
        stdcall, dynlib: "kernel32", importc: "GetModuleHandleW", sideEffect.}
-  proc getProcAddress(hModule: HANDLE, lpProcName: cstring): pointer {.
+  proc getProcAddress(hModule: Handle, lpProcName: cstring): pointer {.
        stdcall, dynlib: "kernel32", importc: "GetProcAddress", sideEffect.}
   proc rtlNtStatusToDosError(code: uint64): ULONG {.
        stdcall, dynlib: "ntdll", importc: "RtlNtStatusToDosError", sideEffect.}
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -211,7 +211,6 @@ type
     callbacks*: Deque[AsyncCallback]
     idlers*: Deque[AsyncCallback]
     trackers*: Table[string, TrackerBase]
-    inPoll: uint64
 
 proc sentinelCallbackImpl(arg: pointer) {.gcsafe, raises: [Defect].} =
   raiseAssert "Sentinel callback MUST not be scheduled"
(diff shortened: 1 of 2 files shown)


chronos/asyncloop.nim
```diff
@@ -425,59 +459,91 @@ when defined(windows):
...
+            raiseOSError(errCode)
+          0
+        else:
+          eventsReceived
+
+    for i in 0 ..< networkEventsCount:
+      var customOverlapped = events[i].lpOverlapped
+      customOverlapped.data.errCode =
+        block:
+          let res = cast[uint64](customOverlapped.internal)
```
etan-status, line 522: Does regular `getQueuedCompletionStatus` also update `internal`/`status`? 
Previously, the return value was used to decide on putting -1 or `osLastError()` instead.

tests/testbugs.nim
```diff
@@ -98,6 +98,40 @@ suite "Asynchronous issues test suite":
...
     result = r1 and r2
 
+  proc createBigMessage(size: int): seq[byte] =
+    var message = "MESSAGE"
+    var res = newSeq[byte](size)
+    for i in 0 ..< len(result):
+      res[i] = byte(message[i mod len(message)])
+    res
+
+  proc testIndexError(): Future[bool] {.async.} =
```
etan-status, line 108: Thanks for finding and constructing a reproducible test case for this elusive bug!!

etan-status (approved):

chronos/asyncloop.nim
cheatfate: Of course. OVERLAPPED mechanism is not part of IOCP actually. So it doesn't matter what function we are using to query IOCP port. And yes previous we have used `GetLastError()`, but after my checks i found that both sources returns identical information when OVERLAPPED structure pointer is not `NULL`.

PR #310, opened 2022-09-05
Fix some warnings.

cheatfate:


chronos/transports/osnet.nim
```diff
@@ -749,10 +749,10 @@ when defined(linux):
         res.maclen = plen
       elif attr.rta_type == IFLA_MTU:
         var p = cast[ptr uint32](RTA_DATA(attr))
-        res.mtu = cast[int](p[])
+        res.mtu = int(p[])
```
arnetheduck, line 752: no risk of defect on 32-bit?

PR #311, opened 2022-09-06
Documentation

Menduist:
Generate documentation on each push to master: https://status-im.github.io/nim-chronos/docs/chronos.html

And some touch ups on existing documentation


README.md
```diff
@@ -243,6 +242,56 @@ In the strict mode, `async` functions are checked such that they only raise
...
+    await c3()
+
+waitFor(c4().wait(10.seconds))
+```
+
+In this example, `c3` swallows cancellations, so `c4` cannot be cancelled
+properly, and the `wait` timeout will not work
+
+
 ### Multiple async backend support
```
cheatfate, line 295: This paragraph should be removed.
Because there is no multiple backends, we trying to make the only one backend and if some other packages support `asyncdispatch` it their own way. Chronos is not going to support or be compatible with `asyncdispatch`. Also i do not see any reasons to have links in here.

Menduist: cc @arnetheduck who added it here

arnetheduck: chronos does not maintain compatibility with ad but there exist libraries that support both ad and chronos by providing an abstraction layer that removes the difference - they need a way to pick which backend to use, and it's better that there exists a standard way of doing so than not - this is important also when frameworks want to introduce support for chronos without breaking their existing ad support, for example to provide a transition period - as such, this is documented both here and in the standard library

.github/workflows/doc.yml
```diff
@@ -0,0 +1,62 @@
...
+    continue-on-error: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - uses: jiro4989/setup-nim-action@v1
+        with:
+          nim-version: 'stable'
```
arnetheduck, line 24: should pin a nim version: `doc` will have different outputs depending on what bugs are present and how the rendering changes - just because it's called "stable" doesn't mean it is

cheatfate:
Why we should keep references to `libp2p` and other projects if this projects do not keep references to `chronos`?

README.md
```diff
@@ -243,6 +242,56 @@ In the strict mode, `async` functions are checked such that they only raise
 effects on forward declarations, callbacks and methods using
 `{.raises: [CatchableError].}` (or more strict) annotations.
 
+### Cancellation support
+
```
cheatfate, line 246: This is very confusing and incorrect of what cancellation is. There is absolutely no information why cancellation is needed and where it is used, and why it can't be avoided. But there a lot of information how to swallow, bypass and ignore cancellation.

README.md
```diff
@@ -243,6 +242,56 @@ In the strict mode, `async` functions are checked such that they only raise
...
+proc c4 {.async.} =
+  while true:
+    await c3()
+
+waitFor(c4().wait(10.seconds))
+```
+
+In this example, `c3` swallows cancellations, so `c4` cannot be cancelled
+properly, and the `wait` timeout will not work
```
cheatfate, line 292: 'с3' swallows cancellations, but why `wait` timeout will not work? This code will exit from `waitFor` but because cancellation was swallowed, `c4` Future will never be completed.

README.md
```diff
@@ -243,6 +242,56 @@ In the strict mode, `async` functions are checked such that they only raise
...
+  try:
+    await sleepAsync(10.minutes)
+    echo "After sleep" # never reached
+  except CancelledError as exc:
+    echo "We got cancelled!"
+    raise exc
+
+proc c2 {.async.} =
+  await c1()
+  echo "Never reached, since the CancelledError bubble up"
```
cheatfate, line 261: This `echo` will never be printed just because of `c1`, but if `c1`, but this code depends on `c1` code. If `c1` code will not have any continuations this `echo` will be printed.

README.md
```diff
@@ -243,6 +242,56 @@ In the strict mode, `async` functions are checked such that they only raise
...
+
+let work = c2()
+waitFor(work.cancelAndWait())
+```
+
+The cancellation will travel down the call stack, and throw a `CancelledError`
```
cheatfate, line 267: There is no "stack". There is sequence of Futures, but it is definitely not a call stack.

README.md
Menduist:
```nim
import chronos
proc c3 {.async.} =
  try:
    await sleepAsync(100.minutes)
  except CatchableError as exc:
    echo "Something went wrong: ", exc.msg

proc c4 {.async.} =
  while true:
    await c3()

waitFor(c4().wait(5.seconds))
echo "Never reached"
```

README.md
Menduist: Just like the "call stack" is not actually a stack of calls
It's an implementation details whether it's a stack, a sequence or a linked list

README.md
Menduist: Added more simple examples

Menduist:
> Why we should keep references to other projects if this projects do not keep references to chronos?

libp2p mentions chronos as part of the first tutorial on how to use libp2p: https://status-im.github.io/nim-libp2p/docs/tutorial_1_connect/#a-simple-ping-application
We can add it to the libp2p readme if you want

cheatfate:
> > Why we should keep references to other projects
> 
> these references provide examples for how to use chronos - since the project itself lacks docs / examples / tutorials, looking at other code provides a substitute ("how do I download a file with http?")

Looks like we do not have any examples on how to download file via http in provided links. Also because this PR named as documentation we could move this links away from main README deeper inside documentation `Examples` section. Like projects do.

Menduist:
@arnetheduck @cheatfate other remarks on this PR? I would like to have at least the website up, and then we can work on better documentation content

PR #319, opened 2022-09-29
remove sentinelCallbackImpl

bung87:
refs: https://github.com/nim-lang/Nim/pull/20408  

const closure not usefull here, so I consider can saftly remove.

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -214,11 +214,8 @@ type
     idlers*: Deque[AsyncCallback]
     trackers*: Table[string, TrackerBase]
 
-proc sentinelCallbackImpl(arg: pointer) {.gcsafe, raises: [Defect].} =
-  raiseAssert "Sentinel callback MUST not be scheduled"
-
 const
-  SentinelCallback = AsyncCallback(function: sentinelCallbackImpl,
+  SentinelCallback = AsyncCallback(function: nil,
                                    udata: nil)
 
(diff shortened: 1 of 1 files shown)


arnetheduck:
> not usefull here

the closure is useful / deliberate here: it provides an assert when misused / there's a bug

bung87:
I see `isSentinel` is used, so it may not be mis used internal, does users can access it ?

arnetheduck:
> does users can access it ?

indirectly, when calling functions within the library that use it - directly, no - but that doesn't really matter - what we want to avoid is:

* global var/let - these cause `gcsafe` violations
* new instance every time: unnecessary heap allocations

bung87:
> > does users can access it ?
> 
> indirectly, when calling functions within the library that use it - directly, no - but that doesn't really matter - what we want to avoid is:
> 
> * global var/let - these cause `gcsafe` violations
> * new instance every time: unnecessary heap allocations

I understand that, so I cant find easier way to patch this. in the end I change it to nil.

arnetheduck:
>  so I cant find easier way to patch this

The capability to make const proc shouldn't be removed from Nim - potentially, it could be limited so that proc's assigned to consts don't do const violations

bung87:
I may make it partially support or wait for real const clsoure implementation.

arnetheduck:
> wait for real const clsoure implementation.

this is the way to go - ie Nim shouldn't break existing valid use cases unless there's a strong reason to, and this doesn't strike me as one

PR #321, opened 2022-10-10
NewConnectionAlways leaks connections fix.

cheatfate:


chronos/apps/http/httpclient.nim
```diff
@@ -634,24 +634,30 @@ proc releaseConnection(session: HttpSessionRef,
...
-    of HttpClientConnectionState.ResponseHeadersReceived:
-      if (HttpClientConnectionFlag.NoBody in connection.flags) and
-         (HttpClientConnectionFlag.KeepAlive in connection.flags):
-        # HTTP response headers received with an empty response body and
-        # "Connection: keep-alive" is present in response headers.
-        false
       else:
-        # HTTP response body is not received or "Connection: keep-alive" is not
-        # present or not supported.
+        if HttpClientConnectionFlag.KeepAlive in connection.flags:
```
tersec, line 640: This whole thing is equivalent to just `HttpClientConnectionFlag.KeepAlive notin connection.flags`, but maybe having this more explicit is useful to explain its functionality/purpose better.

tersec: In general, the entire thing is a way to get a `bool` out of a combination of a bunch of other `bool` conditions, so the comment in general refers to the whole `of HttpClientConnectionState.ResponseBodyReceived:` branch of the `case`.

It presumably doesn't make a difference given reasonable compile/codegen for the result, so it's a question of what most clearly communicates intent.

etan-status (approved):

zah:
Replaced by https://github.com/status-im/nim-chronos/pull/323

PR #322, opened 2022-10-11
AsyncEventQueue: add resetRegistration() to continue using the same r…

dbrignoli:
…egistration key after AsyncEventQueueFullError

With _EventBus_ being deprecated I am checking how to move exiting code to _AsyncEventQueue_ and was wondering if the addition of `resetRegistration()` (tentative proc name) would be acceptable. It could be used when tasks catch `AsyncEventQueueFullError` and handle/log the occurence but then continue receiving events using the same key instead of  getting a new one.

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -722,6 +722,15 @@ proc unregister*(ab: AsyncEventQueue, key: EventQueueKey) {.
     ab.readers.delete(index)
     ab.compact()
 
+proc resetRegistration*(ab: AsyncEventQueue, key: EventQueueKey) {.raises: [Defect].} =
+  let index = ab.getReaderIndex(key)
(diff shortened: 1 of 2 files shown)


cheatfate:
why do you need to reset? I'm not sure i understand why you need to perform unregister/register sequence, if you can just continue working/handling events. Because your procedure does not do anything helpful actually - pending events still going to be scheduled, so what the reason?

dbrignoli:
@cheatfate, my understanding is that once `reader.overflow` is set to true it will keep raising `AsyncEventQueueFullError` so there is no way to "reset" a reader after an overflow. The only place where `reader.overflow` is set to `false` is upon registration.

The reset would allow to keep using the same reader (using the same key) by resetting it after handling `AsyncEventQueueFullError`.

PR #323, opened 2022-10-13
HTTP client: Allow request connection management.

cheatfate:
This is proper fix for issue and replaces https://github.com/status-im/nim-chronos/pull/321.
With this fix
1. `NewConnectionAlways` global session flag which will create new connection for every HTTP request and close this connection after request's response received.
2. Added `HttpClientRequestFlag.DedicatedConnection` flag which will force to create new HTTP connection for this specific request. Note: Connection is not going to be closed.
3. `HttpClientRequestFlag.CloseConnection` flag will force HTTP client to send send `Connection: close` HTTP header with specific HTTP request and it also closes connection after response will be received.

So now, instead of using `NewConnectionAlways` you can specify both `DedicatedConnection` and `CloseConnection` to get equal behavior.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -67,7 +67,8 @@ type
     Custom                    ## None of the above
 
   HttpClientRequestFlag* {.pure.} = enum
-    CloseConnection,          ## Send `Connection: close` in request
+    DedicatedConnection,      ## Create new HTTP connection for request
+    CloseConnection           ## Send `Connection: close` in request
 
   HttpClientConnectionFlag* {.pure.} = enum
     Request,                  ## Connection has pending request
@@ -110,6 +111,7 @@ type
     headersTimeout*: Duration
     connectionBufferSize*: int
     maxConnections*: int
+    connectionsCount*: int
     flags*: HttpClientFlags
 
   HttpAddress* = object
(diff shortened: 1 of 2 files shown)


chronos/apps/http/httpclient.nim
```diff
@@ -620,13 +626,23 @@ proc acquireConnection(session: HttpSessionRef,
...
+      var res = false
+      session.connections.withValue(conn.remoteHostname, connections):
+        connections[].keepItIf(it != conn)
+        if len(connections[]) == 0:
+          res = true
+      res
+  if removeHost:
+    session.connections.del(conn.remoteHostname)
   await conn.closeWait()
+  dec(session.connectionsCount)
```
etan-status, line 645: Seems a bit tricky re cancellation of the await above; not sure if there are situations where cancellation can be triggered though.

`connectionsCount` seems to be only used in test code which is a controlled environment, so it's probably alright.

etan-status (approved):
I presume the code that actually closes the connection is triggered implicitly with presence of HTTP/1.0 or Connection: Close? i.e., if a buggy server sends "Connection: Close" but then leaves the connection open, we would still close the connection after receiving the response / timeout?

From tests it looks good :) Thanks for the adjustments!

PR #324, opened 2022-10-18
Add idle connection timeouts for HTTP client's connections pool.

cheatfate:
Add timestamps and duration for both HTTP client requests/responses.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -857,14 +857,26 @@ proc prepareResponse(request: HttpClientRequestRef, data: openArray[byte]
   let connectionFlag =
     block:
       case resp.version
-      of HttpVersion11, HttpVersion20:
+      of HttpVersion11:
+        # Keeping a connection open is the default on HTTP/1.1 requests.
+        # https://www.rfc-editor.org/rfc/rfc2068.html#section-19.7.1
         let header = toLowerAscii(headers.getString(ConnectionHeader))
         if header == "close":
           false
         else:
           true
-      else:
+      of HttpVersion10:
+        # This is the default on HTTP/1.0 requests.
         false
+      else:
+        # HTTP/2 does not use the Connection header field (Section 7.6.1 of
+        # [HTTP]) to indicate connection-specific header fields.
+        # https://httpwg.org/specs/rfc9113.html#rfc.section.8.2.2
+        #
+        # HTTP/3 does not use the Connection header field to indicate
+        # connection-specific fields;
+        # https://httpwg.org/specs/rfc9114.html#rfc.section.4.2
+        true
 
   let contentType =
     block:
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -22,9 +22,12 @@ const
     ## Timeout for connecting to host (12 sec)
   HttpHeadersTimeout* = 120.seconds
     ## Timeout for receiving response headers (120 sec)
-  HttpConnectionIdleTimeout* = 120.seconds
+  HttpConnectionIdleTimeout* = 60.seconds
     ## Time after which idle connections are removed from the HttpSession's
     ## connections pool (120 sec)
+  HttpConnectionCheckPeriod* = 10.seconds
+    ## Period of time between idle connections checks in HttpSession's
+    ## connection pool (10 sec)
   HttpMaxRedirections* = 10
     ## Maximum number of Location redirections.
   HttpClientConnectionTrackerName* = "httpclient.connection"
@@ -113,7 +116,9 @@ type
     maxRedirections*: int
     connectTimeout*: Duration
     headersTimeout*: Duration
-    idleTimeout*: Duration
+    idleTimeout: Duration
+    idlePeriod: Duration
+    watcherFut: Future[void]
     connectionBufferSize*: int
     maxConnections*: int
     connectionsCount*: int
@@ -319,6 +324,18 @@ template setDuration(
     reqresp.duration = timestamp - reqresp.timestamp
     reqresp.connection.setTimestamp(timestamp)
 
+template isReady(conn: HttpClientConnectionRef): bool =
+  (conn.state == HttpClientConnectionState.Ready) and
+  (HttpClientConnectionFlag.KeepAlive in conn.flags) and
+  (HttpClientConnectionFlag.Request notin conn.flags) and
+  (HttpClientConnectionFlag.Response notin conn.flags)
+
+template isIdle(conn: HttpClientConnectionRef, timestamp: Moment,
+                timeout: Duration): bool =
+  (timestamp - conn.timestamp) >= timeout
+
+proc sessionWatcher(session: HttpSessionRef) {.async.}
+
 proc new*(t: typedesc[HttpSessionRef],
           flags: HttpClientFlags = {},
           maxRedirections = HttpMaxRedirections,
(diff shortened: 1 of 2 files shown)


chronos/apps/http/httpclient.nim
```diff
@@ -604,18 +666,35 @@ proc acquireConnection(
...
+        var idleConnections: seq[HttpClientConnectionRef]
+        let
+          timestamp = Moment.now()
+          res =
+            block:
+              var cres: HttpClientConnectionRef = nil
+              session.connections.withValue(ha.id, connections):
+                for connection in connections[]:
+                  if connection.isReady():
+                    if connection.isIdle(timestamp, session.idleTimeout):
```
arnetheduck, line 678: shouldn't this happen on a timer? ie if nobody is requesting new connections, they'll stay forever?

chronos/apps/http/httpclient.nim
```diff
@@ -604,18 +666,35 @@ proc acquireConnection(
...
+          # We mark connection as acquired to avoid race, because we need to
+          # prune idle connections.
+          res.state = HttpClientConnectionState.Acquired
+        if len(idleConnections) > 0:
+          # Closing connections which was idle for`session.idleTimeout`
+          # time.
+          var pending: seq[Future[void]]
+          for connection in idleConnections:
+            pending.add(session.removeConnection(connection))
+          await allFutures(pending)
```
arnetheduck, line 694: this looks risky: if I'm trying to acquire a connection presumably I have some use for it which shouldn't be delayed by the closing of unrelated connections

chronos/apps/http/httpclient.nim
```diff
@@ -284,29 +298,75 @@ template checkClosed(reqresp: untyped): untyped =
...
 proc new*(t: typedesc[HttpSessionRef],
           flags: HttpClientFlags = {},
           maxRedirections = HttpMaxRedirections,
           connectTimeout = HttpConnectTimeout,
           headersTimeout = HttpHeadersTimeout,
           connectionBufferSize = DefaultStreamBufferSize,
-          maxConnections = -1): HttpSessionRef {.
+          maxConnections = -1,
+          idleTimeout = HttpConnectionIdleTimeout,
+          idlePeriod = HttpConnectionCheckPeriod): HttpSessionRef {.
```
arnetheduck, line 347: this is imo an implementation detail that we shouldn't expose in public API - it doesn't matter greatly since we can always make it a no-op, but it feels like the kind of thing that doesn't greatly matter for the user of the API

chronos/apps/http/httpclient.nim
```diff
@@ -836,22 +959,27 @@ proc prepareResponse(request: HttpClientRequestRef, data: openArray[byte]
...
-      raiseHttpReadError("Could not read response headers")
+  let timestamp = Moment.now()
+  req.connection.setTimestamp(timestamp)
+  let
+    bytesRead =
+      try:
+        await req.connection.reader.readUntil(addr buffer[0],
+                                              len(buffer), HeadersMark).wait(
+                                              req.session.headersTimeout)
+      except CancelledError as exc:
```
arnetheduck, line 970: no need to catch really since the other exceptions are independent

chronos/apps/http/httpclient.nim
```diff
@@ -707,11 +770,59 @@ proc closeWait*(session: HttpSessionRef) {.async.} =
...
+            toClose.add(conn)
+          else:
+            # Active connection (currently used) or not idle connection
+            toKeep.add(conn)
+        session.connections[peer] = toKeep
+        idleConnections.add(toClose)
+
+    if len(idleConnections) > 0:
+      dec(session.connectionsCount, len(idleConnections))
+      var pending = newSeqOfCap[Future[void]](len(idleConnections))
```
arnetheduck, line 811: today is functional programming day :)

```nim
let pending = idleConnections.mapIt(it.closeWait)
```

arnetheduck (approved):

chronos/apps/http/httpclient.nim
```diff
@@ -1079,13 +1211,16 @@ proc open*(request: HttpClientRequestRef): Future[HttpBodyWriter] {.
   try:
     let headers = request.prepareRequest()
     request.connection.state = HttpClientConnectionState.RequestHeadersSending
+    request.setTimestamp()
```
arnetheduck, line 1214: shouldn't this be part of `prepareRequest`?

chronos/apps/http/httpclient.nim
```diff
@@ -284,29 +298,75 @@ template checkClosed(reqresp: untyped): untyped =
...
+
+template isReady(conn: HttpClientConnectionRef): bool =
+  (conn.state == HttpClientConnectionState.Ready) and
+  (HttpClientConnectionFlag.KeepAlive in conn.flags) and
+  (HttpClientConnectionFlag.Request notin conn.flags) and
+  (HttpClientConnectionFlag.Response notin conn.flags)
+
+template isIdle(conn: HttpClientConnectionRef, timestamp: Moment,
+                timeout: Duration): bool =
+  (timestamp - conn.timestamp) >= timeout
```
arnetheduck, line 335: could it happen that some really long send (several mb?) causes the connection to be seen as idle?

chronos/apps/http/httpclient.nim
```diff
@@ -583,66 +643,67 @@ proc connect(session: HttpSessionRef,
...
+
+    if not(isNil(conn)): return conn
+
+    conn =
+      try:
+        await session.connect(ha).wait(session.connectTimeout)
+      except AsyncTimeoutError:
+        raiseHttpConnectionError("Connection timed out")
+    conn.state = HttpClientConnectionState.Acquired
+    session.connections.mgetOrPut(ha.id, default).add(conn)
```
arnetheduck, line 704:
```suggestion
    session.connections.mgetOrPut(ha.id, default(seq[HttpClientConnectionRef])).add(conn)
```

chronos/apps/http/httpclient.nim
```diff
@@ -583,66 +643,67 @@ proc connect(session: HttpSessionRef,
...
+        var cres: HttpClientConnectionRef = nil
+        let
+          timestamp = Moment.now()
+          connections = session.connections.getOrDefault(ha.id, default)
+        # We looking for non-idle connection, all idle connections will be
+        # freed by sessionWatcher().
+        for connection in connections:
+          if connection.isReady() and
+             not(connection.isIdle(timestamp, session.idleTimeout)):
+            cres = connection
```
arnetheduck, line 691: why not `return connection` in the loop?

chronos/apps/http/httpclient.nim
```diff
@@ -583,66 +643,67 @@ proc connect(session: HttpSessionRef,
...
-          res = true
-      res
-  if removeHost:
-    session.connections.del(conn.remoteHostname)
-  dec(session.connectionsCount)
-  await conn.closeWait()
+        var cres: HttpClientConnectionRef = nil
+        let
+          timestamp = Moment.now()
+          connections = session.connections.getOrDefault(ha.id, default)
```
arnetheduck, line 685:
```suggestion
          connections = session.connections.getOrDefault(ha.id)
```

chronos/apps/http/httpclient.nim
cheatfate: Of course its possible, but we can't control it, if people willing to send gigabytes of data from memory to a socket, how we can check if send was actually happened or not?

chronos/apps/http/httpclient.nim
cheatfate: `prepareRequest` is not related to actual data sending process, so why we need to start from there?

chronos/apps/http/httpclient.nim
cheatfate: I think some of the users could decrease `idlePeriod` to `1.seconds` just to have more precise control.

chronos/apps/http/httpclient.nim
cheatfate:
```
proc mgetOrPut*[A, B](t: var Table[A, B], key: A, val: B): var B =
```
https://github.com/nim-lang/Nim/blob/v1.6.12/lib/pure/collections/tables.nim#L441

chronos/apps/http/httpclient.nim
```diff
@@ -707,11 +770,59 @@ proc closeWait*(session: HttpSessionRef) {.async.} =
...
+      except CancelledError:
+        true
+
+    if firstBreak:
+      break
+
+    var idleConnections: seq[HttpClientConnectionRef]
+    let timestamp = Moment.now()
+    for peer, connections in session.connections.pairs():
+      if len(connections) > 0:
```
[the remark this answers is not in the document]
cheatfate: https://github.com/status-im/nim-chronos/pull/324/commits/bac8aea0741c28592feb3942870ea8142e26ad9a

PR #328, opened 2022-11-10
Fix Future[void] async procedure in generic

Menduist:
```nim
  test "Future void in generic bug":
    proc gen(T: typedesc[int]): T =
      proc testproc(): Future[void] {.async.} = discard
      T(5)
    let test = int.gen()
```
Currently fails

Upstream has a nicer fix: https://github.com/nim-lang/Nim/pull/17633 which also supports aliasing, but it's more invasive, so not sure if we should also integrate it

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -236,6 +236,10 @@ template complete*(future: Future[void]) =
   ## Completes a void ``future``.
   complete(future, getSrcLocation())
 
+template complete*(future: Future[void], val: untyped) =
+  ## Completes ``future``
+  complete(future, getSrcLocation())
+
 proc fail[T](future: Future[T], error: ref CatchableError, loc: ptr SrcLoc) =
   if not(future.cancelled()):
     checkFinished(FutureBase(future), loc)
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -15,8 +15,7 @@ proc skipUntilStmtList(node: NimNode): NimNode {.compileTime.} =
   if node[0].kind == nnkStmtList:
     result = skipUntilStmtList(node[0])
 
-proc processBody(node, retFutureSym: NimNode,
-                 subTypeIsVoid: bool): NimNode {.compileTime.} =
+proc processBody(node, retFutureSym: NimNode): NimNode {.compileTime.} =
   #echo(node.treeRepr)
   result = node
   case node.kind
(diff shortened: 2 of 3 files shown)


Menduist:
Tried a fix similar to upstream, but it changes the `result` behavior in `Future[void]`:
```nim
proc x {.async.} =
  echo result # Fails with Error: type mismatch: got <void> [..] but expression 'raiseAssert "You should not reference the `result` variable inside a void async proc"' is of type: void
  result = 12 # Fails with Error: expression 'raiseAssert "You should not reference the `result` variable inside a void async proc"' has no type (or is ambiguous)
  result # Fails AT RUNTIME with `Unhandled exception: You should not reference the `result` variable inside a void async proc`
```

Unfortunately, I don't see a simple fix to keep the current behavior of keeping the failure at compile time for every case, but maybe someone will have better luck

chronos/asyncmacro2.nim
```diff
@@ -124,41 +116,28 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
-        newNimNode(nnkExprColonExpr).add(newNimNode(nnkBracketExpr).add(
-          newIdentNode("warning"), newIdentNode("resultshadowed")),
-        newIdentNode("off")))) # -> {.push warning[resultshadowed]: off.}
-
-      procBody.insert(1, newNimNode(nnkVarSection, prc.body).add(
-        newIdentDefs(newIdentNode("result"), baseType))) # -> var result: T
-
-      procBody.insert(2, newNimNode(nnkPragma).add(
-        newIdentNode("pop"))) # -> {.pop.})
+    procBody.insert(0): quote do:
```
arnetheduck, line 125: using quote do loses original source location from `newNimNode` unfortunately

Menduist: I think in this case its fine, since this one doesn't contain any newNimNode?

arnetheduck:
the idea LGTM but we should stay with `newNimNode` for the result variable itself so that assignments to it point to the right code - can you rebase?

chronos/asyncfutures2.nim
```diff
@@ -236,6 +236,10 @@ template complete*(future: Future[void]) =
   ## Completes a void ``future``.
   complete(future, getSrcLocation())
 
+template complete*(future: Future[void], val: untyped) =
```
arnetheduck, line 239: hm, can we avoid this one somehow? or at least prevent it from being called with an actual value - ie like this, it looks like `complete(fut, 42)` will compile: it needs a check to ensure that val isn't a .. value

chronos/asyncmacro2.nim
arnetheduck:
```
      procBody.insert(1, newNimNode(nnkVarSection, prc.body).add(
        newIdentDefs(newIdentNode("result"), baseType))) # -> var result: T
```

`var result` in `quote do`?

chronos/asyncfutures2.nim
Menduist: Besides a beautiful `static: assert type(x) is void`, don't see a better way right now. But maybe this PR needs a new approach altogether

chronos/asyncmacro2.nim
Menduist: Ohh ok, you meant in the previous code. That is true, but I don't see how this bit could fail (and thus show the invalid lines)

Menduist: Ohh ok, you meant in the previous code. That is true, but I don't see how this bit could fail (and thus show the invalid lines)

Menduist: Ohh ok, you meant in the previous code. That is true, but I don't see how this bit could fail (and thus show the invalid lines)

chronos/asyncfutures2.nim
arnetheduck: I *think* #401 nails it but can you take a look?

Menduist: Yup, a lot better
Played a bit with it, couldn't make it fail horribly

PR #335, opened 2022-12-08
fix int set construction

bung87:


Menduist (approved):

Menduist:
For future readers being confused as I was, `ECONNRESET` is a `cint`, which is 32 bits, which is too big for a set

PR #339, opened 2022-12-16
Add 'firstCompleted' and 'firstCompetedFuture' to 'asyncfutures2'

zah:
These are useful alternatives to the `race` helper that make it easy to implement "first successful response" strategies.

Currently, the Nimbus VC is trying to implement such a strategy, but it has the issue that if some of the servers responds too quickly with an error, the successful responses that might complete later will be ignored.

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -608,7 +608,7 @@ proc `and`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] {.
   ## Returns a future which will complete once both ``fut1`` and ``fut2``
   ## complete.
   ##
-  ## If cancelled, ``fut1`` and ``fut2`` futures WILL NOT BE cancelled.
+  ## On cancellation, ``fut1`` and ``fut2`` futures WILL NOT BE cancelled.
   var retFuture = newFuture[void]("chronos.`and`")
   proc cb(data: pointer) =
     if not(retFuture.finished()):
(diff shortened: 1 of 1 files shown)


arnetheduck:
this looks pretty specialistic - ie the typical pattern for race is to do so in a loop until the desired result is achieved among the futures that were passed to it, removing the ones that were dissatisfactory - it feels like we keep adding special cases here instead of having a more limited amount of fundamental primitives that compose

arnetheduck:
iff we're going to add this feature, it should indeed be implemented as such, ie a while loop over a race-like primitive so we don't have to have this wall of logic, and to highlight that this is syntactic sugar for a special case

zah:
A loop over race will be significantly more expensive (it will feature unnecessary cancellations, O(n^2) in memory copying operations, etc). It's not clear what is gained exactly in return. The code size argument seems week - it's not even clear whether the loop woudn't be similar in size and complexity.

arnetheduck:
well, nonetheless this adds to a jungle of special functions, none of which quite seems to do the right thing - I'd consider it a UX degradation of the futures library to add this for this reason, ie the gain is mostly insignificant - o(n2) doesn't matter, we're not talking about tens of thousands of futures (chronos is not able to do that effeiciently for a number of other reasons).

What is gained is that you can begin to understand the difference between the various pre-composed special cases, ie what they actually do - the differences between them are so subtle and nuanced that it's hard to pick one.

arnetheduck:
This pr would be a lot more palatable if we didn't have race and 20 other future list combinators already - the fact that we have so many speaks of this needing a review with _removals_ rather than more and more additions

zah:
I would further argue that this helper is present in most async run-times. Here is the JavaScript version for example: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

arnetheduck:
Then perhaps we should deprecate `race`?

zah:
Here is `race` in JavaScript:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

arnetheduck:
`race` and `one` have since been joined to a single implementation, and some other combinators are deprecated / removed, so one could argue that the jungle has shrunk and left room for some nuance.

A remaining problem here is the error handling - ie if all of them fail, raising a catchable is not in line with current error handling that has since become explicit in its raising - potential options include:

* a dedicated error
* `ValueError` - similar to when no futures are given
* no error - just return the first future

in terms of naming, `oneCompleted` or `raceCompleted` probably make more sense.

Closing for now since it's been a while and a new PR would make more sense.

PR #343, opened 2023-01-10
Add possibility of binding to local address

diegomrsantos:
In order to do hole punching, a peer must be able to bind to the same port as it is listening to during the dialling synchronization.


Menduist:
#362 got merged instead

PR #346, opened 2023-01-15
`async` `proc` types

arnetheduck:
Allow `type = proc(...) {.async.}` for pre-declaring async callback types - in this case, the macro applies the correct transformation to the return type and adds the appropriate exception / gcsafe annotations

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -88,13 +88,12 @@ proc cleanupOpenSymChoice(node: NimNode): NimNode {.compileTime.} =
 proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
   ## This macro transforms a single procedure into a closure iterator.
   ## The ``async`` macro supports a stmtList holding multiple async procedures.
-  if prc.kind notin {nnkProcDef, nnkLambda, nnkMethodDef, nnkDo}:
-      error("Cannot transform this node kind into an async proc." &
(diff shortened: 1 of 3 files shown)


chronos/asyncmacro2.nim
```diff
@@ -109,135 +108,157 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+          newCall(newIdentNode("futureContinue"), retFutureSym)
+      )
+
+      # -> return resultFuture
+      outerProcBody.add newNimNode(nnkReturnStmt, prc.body[^1]).add(retFutureSym)
+      outerProcBody
+    else:
+      nil
+  else:
+    nil
```
Menduist, line 245: Personally, I prefer to do
```nim
if not condition:
  nil
else:
  [long piece of code]
```
than what is being done here, because by the time we get to the `else`, we forgot the context of the original `if`. With the `if xx: nil else:` we have the full context in sight

Also fyi, https://github.com/status-im/nim-chronos/pull/251/files also added the ability to have async proc types

arnetheduck: removed the nils completely instead

> also added

it might be worth extracting any other bugfixes and refactoring from that branch so that we reduce the surface of merging it (and thus make it easier to revert should the need arise)

arnetheduck: (also, this PR fixes some incorrect raises, ie both Exception and CatchableError were being added redundantly)

cheatfate (approved):

PR #350, opened 2023-01-20
Introduce `chronos/futures`, `chronosv4`, with associated refactoring

arnetheduck:
In prerparing chronos for a v4 release with raise tracking and other mildly incompatible changes, it's a good opportunity to revisit some of the core types and definitions - this PR starts that journey by introducing `chronos/futures` which contains the `Future` type along with those usage helpers that don't require event loop / `callSoon` support.

This subset allows breaking cyclic dependencies in applications using `chronos` and makes it easier to integrate `chronos` into applications in general, by allowing them to use the `Future` type independently of the rest of the chronos library, for example to expose their API expectations in modules that don't import all of chronos (and thus keep symbol pollution down).

The full migration is expected to affect backwards compatiblity more significantly, thus a `chronosv4` define can be used to "preview" the breaking changes - this set is expected to grow up to the v4 release, but the aim is to maintain a more or less complete v3 subset such that applications can target both v3 and v4 with relative ease for a transition period (similar to how `chronosStrictExceptions` was introduced).

This is the the first of a series of PR:s that gradually will introduce the new features.

* introduce `chronos/config` that collects and documents compile-time configuration flags
* initial code moves that introduce `chronos/futures`
* simplify transformation code
* prevent modification of internal `Future` fields by making most of them private (TODO currently breaking even in v3)
* `FutureState.Finished` -> `FutureState.Completed` to stay consistent with function names
* introduce low-level `Future` constructors that create "finished" futures (TODO naming - convenience of reusing `completed` vs template lookup conflicts due to overloading)


arnetheduck:
With the fields made public again while in v3 mode, this PR should be good to go

chronos.nimble
```diff
@@ -32,10 +32,11 @@ proc run(args, path: string) =
...
 task test, "Run all tests":
   for args in [
-      "-d:useSysAssert -d:useGcAssert",
-      "-d:chronosStackTrace -d:chronosStrictException",
+      "-d:debug",
+      "-d:debug -d:chronosv4",
       "-d:release",
-      "-d:release -d:chronosFutureTracking",
+      "-d:release -d:chronosv4",
+      "-d:release -d:useSysAssert -d:useGcAssert",
```
cheatfate, line 39: why `chronosFutureTracking` was removed?

chronos/asyncloop.nim
```diff
@@ -1139,7 +1133,7 @@ proc wait*[T](fut: Future[T], timeout = InfiniteDuration): Future[T] =
             clearTimer(timer)
 
           if fut.failed():
-            retFuture.fail(fut.error)
+            retFuture.fail(internalError(fut))
```
cheatfate, line 1136: I'm confused about what `internalError` means...

chronos/config.nim
```diff
@@ -0,0 +1,50 @@
...
+    chronosStackTrace* {.booldefine.}: bool = defined(debug)
+      ## Include stack traces in futures for creation and completion points
+
+    chronosFutureId* {.booldefine.}: bool = defined(debug)
+      ## Generate a unique `id` for every future - when disabled, the address of
+      ## the future will be used instead
+
+    chronosFutureTracking* {.booldefine.}: bool = defined(debug)
+      ## Generate a linked list of all relevant in-flight futures to allow
+      ## detection of hung tasks - works best with stack traces enabled
```
cheatfate, line 26: > - works best with stack traces enabled - totally incorrect statement

It doesn't depend on `stack traces` at all and uses nim's compiler source line information.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+  type StackTrace = string
+
+type
+  LocationKind* {.pure.} = enum
+    Create
+    Complete
+
+  CallbackFunc* = proc (arg: pointer) {.gcsafe, raises: [Defect].}
+
+  # TODO v4: consider removing this type and just use plain `CallbackFunc`
```
cheatfate, line 34: This TODO could be removed because this task is almost impossible to achieve.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+  AsyncCallback* = object
+    function*: CallbackFunc
+    udata*: pointer
+
+  FutureState* {.pure.} = enum
+    Pending, Completed, Cancelled, Failed
+
+  FutureBase* = ref object of RootObj ## Untyped future.
+    when defined(chronosv4):
```
cheatfate, line 43: Personally i do not like name `chronosv4`, but even if this name is going to be accepted it should be style compatible with all previous config settings so it should be `chronosV4`.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+      error: ref CatchableError ## Stored exception
+      mustCancel: bool
+    else:
+      location*: array[LocationKind, ptr SrcLoc]
+      callbacks*: seq[AsyncCallback]
+      cancelcb*: CallbackFunc
+      child*: FutureBase
+      state*: FutureState
+      error*: ref CatchableError ## Stored exception
+      mustCancel*: bool
```
cheatfate, line 58: There is no difference between types for `chronosv4` and old one, why this distinction is needed?

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+    cause*: FutureBase
+
+  FutureError* = object of CatchableError
+
+  CancelledError* = object of FutureError
+
+when defined(chronosV4):
+  type
+    NoValueError* = object of FutureError
+      ## Exception raised when trying to access the value of a future when
```
cheatfate, line 95: I dont see any reasons to distinct this errors, it should be just a single error `FutureNotYetCompleted`, and it doesn't matter is it get accessed by `error` or by `read` procedures.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+      ($entry.filename, $entry.procname)
+
+  proc getHint(entry: StackTraceEntry): string =
+    ## We try to provide some hints about stack trace entries that the user
+    ## may not be familiar with, in particular calls inside the stdlib.
+
+    let (filename, procname) = getFilenameProcname(entry)
+
+    if procname == "processPendingCallbacks":
+      if cmpIgnoreStyle(filename, "asyncdispatch.nim") == 0:
```
cheatfate, line 237: asyncdispatch.nim - really?

chronos/asyncloop.nim
arnetheduck: `internal*` provide public access to the otherwise private fields - since currently, all the fields are public it leads to people being able to set the fields "manually" outside of chronos' control which is difficult to provide backwards-compatibility for.

chronos/futures.nim
arnetheduck: what are current blockers?

chronos/futures.nim
arnetheduck: the old ones are public (`*`) - the new ones are all private and accessed only via `internal*` - unfortunately, there's quite a lot of existing code that does `if fut.failed: echo fut.error` - however, it's currently also possible to do `echo fut.error` on a pending future which seems off

chronos/futures.nim
arnetheduck: hey, this is straight from https://github.com/status-im/nim-chronos/blob/40143f8798e49e036179fbe066b0b32138ecfd85/chronos/asyncfutures2.nim#L420 - no idea what it does, but indeed, it looks out of date

chronos/futures.nim
arnetheduck: changed to `FuturePendingError` - `Completed` would mean that it has a value (and not an error/cancel)

chronos/futures.nim
arnetheduck: renamed to `chronosPreviewV4` - once all of V4 is in place, the intent would be to remove the `when` along with the v3 support - this is just a way to allow the two versions to coexist temporarily during development - the alternative would be a branch

chronos.nimble
arnetheduck: it's now enabled by `chronosPreviewV4` (previously `chronosv4`)

cheatfate: But it should not be enabled with by default.

chronos/futures.nim
cheatfate: `udata` is get used everywhere

chronos/futures.nim
cheatfate: But ```fut.failed(): echo fut.error()``` no?

chronos/futures.nim
cheatfate: Whatever Future has error or value it become completed, because it changes its state to state where it can't be changed anymore. So error should indicate the moment of time when attempt to get value or error from Future should fail with single type of Exception.

chronos/futures.nim
cheatfate: Yeah but you declared that you perform new `futures` so why should we keep this outdated/legacy stuff? Because i have also ideas how to improve different parts of code and so some of my comments in this review will be about improvements which you are not planned.

chronos/futures.nim
arnetheduck: currently, chronos is not precise when it comes to the names `Completed` and `Finished` - sometimes, `completed` means it has a `value` and sometimes it means its not pending (ie has value _or_ error) - in this PR, `Completed` only means "has value" and `Finished` means `not pending` but we can change that so that `Completed` means "has value".

chronos/futures.nim
arnetheduck: removed

chronos/futures.nim
arnetheduck: right - I see what you mean - so we have two possible error conditions:
* caller accesses `value` or `error` when future is pending
* caller accesses `error` when future completed successfully (so only the `value` field is valid - error contains `nil`

Do we want separate exceptions for the above two cases? In both cases, currently accessing `error` directly returns `nil`.

arnetheduck: 667beafd104c1a68af5b8a2009cc5d4da1292905 pushed to show the problem / various states

cheatfate: This problem is not only works for error, because you can try to call `Future.read()` on `Future[void]` too, it also represents situation when Future is completed but missing value/error. So i think we could introduce 2 exceptions
1. Signals when Future is not completed yet.
2. Signals that Future missing requested data, e.g. when `read()` called on `Future[void]` or `error()` called when there no error.

chronos.nimble
arnetheduck: ah, sorry, I misread which feature you meant - so if you look at `config.nim`, the future tracking would now be enabled by default `-d:debug` - you can also separately enable or disable it with `-d:chronosFutureTracking=true|false` which overrides `debug` - is this too aggressive?

chronos/futures.nim
arnetheduck: latest version does exactly this for `read` and `readError` - `error()` is still unchanged because that's a larger breaking change - it did not raise an exception before, meaning this needs some careful thinking

cheatfate: `readError()` which returns `nil` is impossible to use with `raise`, so even if this behavior is missing right now it will cause runtime problems, same idea with `read` on `void`.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+    newMsg.add("Exception message: " & exceptionMsg & "\n")
+
+    # # For debugging purposes
+    # newMsg.add("Exception type:")
+    # for entry in getStackTraceEntries(future.error):
+    #   newMsg.add "\n" & $entry
+    future.error.msg = newMsg
+
+# Internal utilities - these are not part of the stable API
+proc internalCheckComplete*(fut: FutureBase) {.
```
cheatfate, line 309: This thing should not be `procedure` it should be `template` otherwise it will always fall into stack traces and sometimes it hides original procedure name, so even for further work on async traces this should become `template`.

chronos/futures.nim
```diff
@@ -0,0 +1,369 @@
...
+  func error*(fut: FutureBase): ref CatchableError {.
+      deprecated: "Use `readError` to access error of future".} = fut.error
+
+  func `[]`*(loc: array[LocationKind, ptr SrcLoc], v: int): ptr SrcLoc =
+    case v
+    of 0: loc[LocationKind.Create]
+    of 1: loc[LocationKind.Complete]
+    else: raiseAssert("Unknown source location " & $v)
+
+proc read*[T](future: Future[T] ): T {.
```
cheatfate, line 340: This procedure should also be `template`.

chronos/futures.nim
arnetheduck: they're actually not used all that frequently - at least not productively - we create a dynamic environment in many cases for other reasons so often whatever is in udata can be included in that environment instead - it's something I've thought about revisiting at some point but never had the time

chronos/futures.nim
arnetheduck: it might be better to keep this as a function - it doesn't make any difference performance-wise really (compared to all allocations going on in general), but turning it into a function means we'll have a copy of the code in every instantiation (instead of a single one for `FutureBase`) - chronos is already one of the biggest code size contributors, so I think it might make sense to keep this as a non-generic proc

chronos/futures.nim
arnetheduck: hm, a template would have to protect against multiple evaluations of `future` - I think the difference is minimal - once we fully move to 1.6, we can return `lent` instead which will remove any practical performance difference vs a template, wdyt?

chronos/futures.nim
cheatfate: Both `proc` and `func` do not solve the problem of forming stack/async traces. And getting proper stack/async trace is much more important... Right now we don't have ability to form this stack/async trace, but when this work will be started all this "helper" procedures/funcs will hide place in actual async procedure code. So this "helpers" should become hidden and "helpers" become hidden when they become templates.

chronos/futures.nim
cheatfate: Ditto.

chronos/futures.nim
cheatfate: This `udata` is commonly used in Windows handling code, Future[T] combinators and in different waiting procedures. I wish to avoid create closures all the time i need to pass some integer/boolean value through callback.

arnetheduck: renamed to `InternalAsyncCallback` so as to avoid confusion about whether user code should use this type (it should not)

chronos/futures.nim
arnetheduck: `await` uses `internalValue` directly, without calling `read`

chronos/futures.nim
arnetheduck: made it a template for now - I think I found a better fix for stack code locations but I want to merge this PR first

chronos/futures.nim
arnetheduck: ok - I've made `error()` return `nil` if the future is not completed or completed with value for now - this is consistent with what happens today - we can return to this later before releasing v4.

`readError` instead raises an exception, so callers can choose if they want `nil` or an exception. I'm not too happy about the `readError` name.

Finally, in v4, `value()` now raises a defect if the future is in the wrong state - previously it returned the default value. Possibly, we should make `error()` raise a defect as well so as to "balance" the API in terms of defects.

chronos/futures.nim
```diff
@@ -0,0 +1,409 @@
...
+template Finished* {.deprecated: "Use Completed instead".} = Completed
+template Finished*(T: type FutureState): FutureState {.deprecated: "Use FutureState.Completed instead".} = FutureState.Completed
+
+const
+  LocCreateIndex* = LocationKind.Create
+  LocCompleteIndex* = LocationKind.Complete
+
+when chronosFutureId:
+  var currentID* {.threadvar.}: uint
+  currentID = 0'u
```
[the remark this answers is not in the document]
arnetheduck: Addressed in #371

arnetheduck:
replaced by #373

PR #353, opened 2023-01-25
TCP write cancellation support

Menduist:
This PR allows to cancel TCP write, if they are not already begun
Seems to work on every platform, surprisingly

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -414,6 +414,10 @@ elif defined(windows):
           if transp.kind == TransportKind.Socket:
             let sock = SocketHandle(transp.fd)
             var vector = transp.queue.popFirst()
+
+            if vector.buflen == 0:
+              # send cancelled
+              continue
             if vector.kind == VectorKind.DataBuffer:
               transp.wovl.zeroOvelappedOffset()
               transp.setWriterWSABuffer(vector)
@@ -1316,6 +1320,10 @@ else:
           failPendingWriteQueue(transp.queue, error)
 
       var vector = transp.queue.popFirst()
+
+      if vector.buflen == 0:
+        # send cancelled
+        continue
       case vector.kind
       of VectorKind.DataBuffer:
         let res =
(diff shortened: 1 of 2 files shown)


chronos/transports/stream.nim
```diff
@@ -2112,6 +2126,34 @@ template fastWrite(transp: auto, pbytes: var ptr byte, rbytes: var int,
             retFuture.fail(error)
             return retFuture
 
+proc addToWriteQueue(transp: StreamTransport, vector: StreamVector) =
+  transp.queue.addLast(vector)
+
+  # to catch cancellation
+  let childFut = newFuture[void]("stream.transport.write.child")
```
Menduist, line 2133: This new future is a bit unfortunate, but this the current Futures system, this is the only way to catch a cancel request, and be able to swallow it (which will happen here if we can't cancel)
I think it's useful to let higher level know if their message did get sent or not, but if we don't care about that, we can just remove this

cheatfate:
1. What the reason to cancel write request?
2. Additional Future[T] for every write request is pretty heavy price for it.
3. Why PIPE test is missing? Maybe its not working?

Menduist:
> What the reason to cancel write request?

The main use-case right now is the upcoming "big blocks" for ethereum. While sending a big block to a peer (which can take a long time on slow connections), we may realize that this was actually useless since the peer already know the block

If we have the opportunity to cancel it, that would save bandwidth

> Additional Future[T] for every write request is pretty heavy price for it.

I agree, do you think we should change the Future system to let "callback futures" to swallow cancel requests, or just not swallow the cancel?

> Why PIPE test is missing? Maybe its not working?

I didn't think about PIPE, I'll add tests

cheatfate:
> > What the reason to cancel write request?
> 
> The main use-case right now is the upcoming "big blocks" for ethereum. While sending a big block to a peer (which can take a long time on slow connections), we may realize that this was actually useless since the peer already know the block
> 

This is totally insane, just because you talking about high application level in this model. For example
1. `nimbus` sends "big block" using `libp2p` stream
2. `libp2p` stream uses `libp2p muxer` to insert this stream into `libp2 connection` (And there a lot of other libp2p streams could reuse this single `libp2p connection`).
3. `libp2p connection` provides encryption and keeps its own counters/states and writes to `libp2p transport`.
4. `libp2p transport` uses `chronos transport` to write data to TCP socket.

And you provide cancellation support to drop packets which should be sent to remote side... So at any moment if you drop some of the packets written you already break `3`, `2` and who knows what will happens with `1`. Also because `2` would be definitely broken (some of the data removed from the chronos writing queue could be related to other streams), it become impossible to recover and reset `libp2p muxer`. So the only way for you to recover the connection is to drop TCP connection and establish a new one.

Menduist:
You are right, I forgot that the encryption layer was stateful and thus we couldn't just remove messages from the queue, without having to re-encrypt the remaining queue

So any cancellation should happen at the muxer level, not in chronos. I'll close this PR since I don't need it anymore

PR #354, opened 2023-01-27
IpNet refactoring

cheatfate:
1. Remove `result` usage.
2. Switch from `cast[T]` usage to `endians2` usage.
3. Add more IPv4 and IPv6 helpers for TransportAddress checks and tests for it.
4. Add `isGlobal()` procedure to check IPv4 and IPv6 addresses are globally reachable (e.g. public). Add proper tests.
5. `isUnspecified()` check now checks for `0.0.0.0` and become alias for `isZero()` and `isNone()` check now checks for `AdressFamily.None`.

This PR superseeds https://github.com/status-im/nim-chronos/pull/352

### DIFF
--- a/chronos/transports/common.nim
+++ b/chronos/transports/common.nim
@@ -143,13 +143,10 @@ proc `==`*(lhs, rhs: TransportAddress): bool =
     true
   of AddressFamily.IPv4:
     if lhs.port != rhs.port: return false
-    uint32.fromBytes(lhs.address_v4) == uint32.fromBytes(rhs.address_v4)
+    lhs.address_v4 == rhs.address_v4
   of AddressFamily.IPv6:
     if lhs.port != rhs.port: return false
-    (uint64.fromBytes(lhs.address_v6.toOpenArray(0, 7)) ==
-      uint64.fromBytes(rhs.address_v6.toOpenArray(0, 7))) and
-    (uint64.fromBytes(lhs.address_v6.toOpenArray(8, 15)) ==
-      uint64.fromBytes(rhs.address_v6.toOpenArray(8, 15)))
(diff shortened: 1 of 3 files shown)


chronos/transports/common.nim
```diff
@@ -137,19 +137,19 @@ var
...
+  if (lhs.family != rhs.family): return false
   case lhs.family
   of AddressFamily.None:
     true
   of AddressFamily.IPv4:
-    equalMem(unsafeAddr lhs.address_v4[0],
-             unsafeAddr rhs.address_v4[0], sizeof(lhs.address_v4)) and
-      (lhs.port == rhs.port)
+    if lhs.port != rhs.port: return false
+    uint32.fromBytes(lhs.address_v4) == uint32.fromBytes(rhs.address_v4)
```
arnetheduck, line 146:
```suggestion
    lhs.address_v4 == rhs.address_v4
```

arnetheduck: same for the other arrays? not sure about unix though

arnetheduck (approved):
lgtm - I think some / many of the integer conversions are somewhat unnecessary (we could compare byte arrays instead), but it doesn't materially impact the code and it's a really nice improvement over the existing code that does all kinds of tricks

chronos/transports/ipnet.nim
```diff
@@ -36,73 +36,85 @@ type
 proc toNetworkOrder(mask: IpMask): IpMask {.inline.} =
```
arnetheduck, line 36: ideally, we'd remove this function and only ever create `IpMask` instances in one order (network?) - this way, we have a way of knowing which order the data is in by type, without exceptions

PR #357, opened 2023-02-07
Add facility to emulate signals on Windows.

cheatfate:
This PR enabled catching CTRL+C event in Windows console application using `addSignal()` and `waitSignal()` procedures. Also it enables implementation of signals in GUI or Service applications.


chronos/asyncloop.nim
```diff
@@ -587,6 +592,89 @@ when defined(windows):
...
       raise newException(ValueError, osErrorMsg(res.error()))
 
+  {.push stackTrace: off.}
+  proc consoleCtrlEventHandler(dwCtrlType: DWORD): WINBOOL {.
+       stdcall, raises: [Defect].} =
+    ## This procedure will be executed in different thread, so it MUST not use
+    ## any GC related features (strings, seqs, echo etc.).
+    case dwCtrlType
+    of CTRL_C_EVENT:
+      let res = raiseSignal(SIGINT).valueOr: return FALSE
```
arnetheduck, line 602:
```suggestion
      return if raiseSignal(SIGINT).valueOr(false): TRUE else: FALSE
```

arnetheduck (approved):

chronos/asyncloop.nim
Menduist: Or just
```nim
return if raiseSignal(SIGINT).get(false): TRUE else: FALSE
```
?

PR #361, opened 2023-02-20
OS definitions and exceptions changes. (AsyncProc part 1).

cheatfate:


chronos/asyncloop.nim
```diff
@@ -665,8 +691,60 @@ elif unixPlatform:
...
+    try:
+      loop.selector.updateHandle(int(fd), newEvents)
+    except CatchableError:
+      return err(osLastError())
+    ok()
+
+  proc register*(fd: AsyncFD) {.raises: [Defect, OSError].} =
+    ## Register file descriptor ``fd`` in thread's dispatcher.
+    let res = register2(fd)
+    if res.isErr(): raiseOSError(res.error())
```
arnetheduck, line 705: it's possible to create a `toException` function like so:
```nim
func toException(v: OsErrorCode): ref OsError = newOSError(v)
proc register*(fd: AsyncFD) {.raises: [Defect, OSError].} =
 register2(fd).tryGet()
```

it's not something that's widely used though, but maybe this is a good place

chronos/osdefs.nim
```diff
@@ -0,0 +1,1562 @@
...
+  import std/[posix]
+  export posix
+
+  when not defined(android) and defined(amd64):
+    const IP_MULTICAST_TTL*: cint = 33
+  else:
+    var IP_MULTICAST_TTL* {.importc: "IP_MULTICAST_TTL",
+                            header: "<netinet/in.h>".}: cint
+  const
+    EPOLLIN* = 0x00000001
```
arnetheduck, line 885: many of these constants are already (correctly) defined in upstream - why recreate them here? ie this one is part of `posix/epoll` for example ..

chronos/asyncloop.nim
```diff
@@ -270,38 +268,31 @@ template processCallbacks(loop: untyped) =
...
-                                            gcsafe, stdcall, raises: [].}
-
     CompletionKey = ULONG_PTR
 
     CompletionData* = object
       cb*: CallbackFunc
       errCode*: OSErrorCode
-      bytesCount*: int32
+      bytesCount*: uint32
+      cell*: ForeignCell # we need this `cell` to protect our `cb` environment,
```
arnetheduck, line 293: this is not used in this PR, so it should likely be introduced elsewhere?

arnetheduck (approved):

chronos/osdefs.nim
cheatfate: There actually 3 issues.
1. Constants in `posix/epoll` and `posix/kqueue` are not get updated very often.
2. Continuous issues `Error: ambiguous identifier`.
3. We should monitor all the changes inside `epoll` and `kqueue` to avoid issues.

arnetheduck: 1. well, if they are correct in upstream, this would be a benefit (they won't change them either)
2. should be solved with targeted imports/exports - ie the problem with ambiguous identifiers doesn't happen if we don't declare our own copy - this is exactly why it's better to use the common ones when possible, because code using chronos *and* upstream constants will work better (`import posix, chronos` breaks for example when both expose the same symbol, but doesn't break if chronos uses the one from posix)
3. chronos gets tested as part of nim testing, and we test chronos with multiple releases, so this is solved.

I think this has been more of a problem historically than it is now - we can merge this PR as-is but I think it's better to look for opportunities to use the upstream definitions when doing so is not outright wrong. One way to do it is basically to `from posix/epoll import all_correct_symbols` - this guards against future edits to `epoll` adding conflicting symbols while at the same time avoiding `ambiguous identifier` issues in downstream code.

arnetheduck: > I definitely do not want to use upstream for many reasons, there was already long history of updates which made broken every related package

This is not a sustainable option - but above all, it is also one that we have addressed, and having done that investment in upstream stability, we should now recognize the benefits and start profiting from them rather than holding on to an inefficient way of working around issues that no longer exist - ie we've addressed those problems together with upstream, and it holds specially true for "trivial" code like these constants - selectors is different because it introduces opinions whereas these are just trivial definitions. 

Copying things here perhaps makes it easier for chronos to implement itself maybe, but does a disservice to _users_ of chronos which now have to deal with ambiguous symbols, longer compile times and less interoperability.

PR #362, opened 2023-02-21
Add `localAddress` support to `stream.connect`

Menduist:
Supersedes #344

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -50,7 +50,7 @@ type
     # get stuck on transport `close()`.
     # Please use this flag only if you are making both client and server in
     # the same thread.
-    TcpNoDelay
+    TcpNoDelay # deprecated: Use ServerFlags.TcpNoDelay
 
 
   StreamTransportTracker* = ref object of TrackerBase
@@ -699,7 +699,9 @@ when defined(windows):
   proc connect*(address: TransportAddress,
                 bufferSize = DefaultStreamBufferSize,
                 child: StreamTransport = nil,
-                flags: set[TransportFlags] = {}): Future[StreamTransport] =
+                flags: set[TransportFlags] = {},
+                localAddress: TransportAddress = AnyAddress,
+                serverFlags: set[ServerFlags] = {}): Future[StreamTransport] =
     ## Open new connection to remote peer with address ``address`` and create
     ## new transport object ``StreamTransport`` for established connection.
     ## ``bufferSize`` is size of internal buffer for transport.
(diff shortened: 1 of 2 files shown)


chronos/transports/stream.nim
```diff
@@ -699,7 +699,9 @@ when defined(windows):
   proc connect*(address: TransportAddress,
                 bufferSize = DefaultStreamBufferSize,
                 child: StreamTransport = nil,
-                flags: set[TransportFlags] = {}): Future[StreamTransport] =
+                flags: set[TransportFlags] = {},
+                localAddress: TransportAddress = AnyAddress,
```
cheatfate, line 703: I think its better to use `TransporAddress()` as default here.

chronos/transports/stream.nim
```diff
@@ -699,7 +699,9 @@ when defined(windows):
...
+                serverFlags: set[ServerFlags] = {}): Future[StreamTransport] =
```
cheatfate, line 704: Why we need `ServerFlags` here?

Menduist: To support ReuseAddr & ReusePort.
We could also add ReuseAddr to TransportFlags, but then we get
```
nim-chronos/tests/teststream.nim(349, 61) Error: ambiguous identifier: 'ReuseAddr' -- use one of the following:
  ServerFlags.ReuseAddr: ServerFlags
  TransportFlags.ReuseAddr: TransportFlags
```

so don't see a way to make it backward compatible

cheatfate: Its ok, people could use `TransportFlags.ReuseAddr`.

Menduist: Sure, but that breaks everywhere people just use `{ReuseAddr}` instead of `{ServerFlags.ReuseAddr}` (like in these test)
Or we could add `TransportFlags.ClientReuseAddr`

cheatfate: So the idea is to avoid double flags in connect() call. So personally i would prefer to introduce `SocketFlags` for both `server` and `client`.

Menduist: Sure, but then I would also need to add `WinServerPipe` & `WinNoPipeFlash` to `SocketFlags`, would that be ok?

cheatfate: yep

chronos/transports/stream.nim
```diff
@@ -50,7 +50,12 @@ type
     # get stuck on transport `close()`.
     # Please use this flag only if you are making both client and server in
     # the same thread.
-    TcpNoDelay
+    TcpNoDelay # deprecated: Use ClientFlags.TcpNoDelay
+
+  ClientFlags* {.pure.} = enum
```
cheatfate, line 55: The only name which greatly explains all the flags inside is `SocketFlags`, because all the flags are related to sockets only.

Menduist: Done

PR #366, opened 2023-02-24
IOSelectors refactoring to properly support signals and processes. (AsyncProc 2)

cheatfate:
1. Eliminate and deprecate exception usage in core ioselectors components.
2. Add ability to switch engines.
3. Remove Windows select() module, because it is not needed and not used in `chronos`.
4. Revisited all operations to properl handle `EINTR`.
5. Removed legacy `Selector.count` which was introduced by Nim developers and did not perform correctly.


chronos/selectors2.nim
```diff
@@ -285,78 +281,45 @@ else:
...
-      skey.ident = pident
-      skey.events = pevents
-      skey.param = pparam
-      skey.data = data
-
-  when ioselSupportedPlatform:
-    template blockSignals(newmask: var Sigset, oldmask: var Sigset) =
+  when defined(macosx) or defined(freebsd) or defined(netbsd) or
+       defined(openbsd) or defined(dragonfly) or defined(linux):
+    proc blockSignals(newmask: var Sigset,
```
arnetheduck, line 286: `newmask` should not be `var` here, it's not modified

chronos/selectors2.nim
```diff
@@ -285,78 +281,45 @@ else:
...
-                events: set[Event]) {.deprecated: "use updateHandle instead".} =
-  ## Update file/socket descriptor ``fd``, registered in selector
-  ## ``s`` with new events set ``event``.
-  ##
-  ## **Deprecated since v0.18.0:** Use ``updateHandle`` instead.
-  s.updateHandle()
+    doAssert((timeout >= min) and (timeout <= max),
+             "Cannot select with incorrect timeout value, got " & $timeout)
+
+when asyncEventEngine == "epoll":
```
arnetheduck, line 318: shouldn't `iocp` be an engine also, which replaces `when windows`?

chronos/selectors2.nim
arnetheduck: if changing is tricky, document at least

chronos/selectors2.nim
cheatfate: It could be, but there no other engines available on Windows. For example linux could use `epoll` and `poll`, BSD/MacOS could also use `linux` and `poll`, but Windows support is different and already bundled inside asyncloop, it do not use selectors2 module.

arnetheduck: https://github.com/piscisaureus/wepoll :)

In general though, it feels a bit more clean if we use a single mechanism for selecting "backend" throughout the source code - it makes the code more easy to understand because there's only one "concept" involved - who knows, maybe future windows versions will support sockets in I/O ring or something similar as well?

I'd go for it, since we're here already - potentially it would then make sense to rename `ioselects` to `backends`

arnetheduck (commented):
LGTM in general - though I would seriously consider a Table-free design for sockets in particular where the various `udata` fields are used to point back to a (`ref SelectorKey`) that is appropriately protected by `GC_ref` - the table lookup, although O(1), still introduces some complexity and overhead and has the same safety properties overall - ie adding and removing items from the table is equivalent to GC_ref / GC_unref and failures lead to the same kind of bugs.

The other point to consider before merging this is the mixing of different integer sizes and types - there's an opportunity to clean this up by introducing a public api that only uses distinct types, then have the "low-level" code only use "low-level" types like `cint` - right now, there's no clear line.

chronos/ioselects/ioselectors_epoll.nim
```diff
@@ -8,520 +8,775 @@
...
+proc new*(t: typedesc[Selector], T: typedesc): SelectResult[Selector[T]] =
+  var nmask: Sigset
+  if sigemptyset(nmask) < 0:
+    return err(osLastError())
+  let epollFd = epoll_create(asyncEventsCount)
+  if epollFd < 0:
+    return err(osLastError())
+  let selector = Selector[T](
+    epollFd: epollFd,
+    fds: initTable[uint32, SelectorKey[T]](asyncInitialSize),
```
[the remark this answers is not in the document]
arnetheduck: worth keeping in mind that `initTable` is not generally needed any more

chronos/selectors2.nim
cheatfate: `WSAPoll` and `select` (implementation which i have removed from the repository) support only sockets on Windows. Pipes and all other things are not supported. There is no reason to keep implementation which do not support all the `chronos` features. Even with `ioselectors_poll` and Posix system with some effort its possible to support all the chronos features (yeah we will need to create threads and all other things, but its possible). But whatever we going to do its impossible to achieve with `WsaPoll` and `select`. So i do not see any reasons to support this mechanisms.

PR #371, opened 2023-03-16
Compile-time configuration

arnetheduck:
This PR moves all compile-time configuration to a single module, simplifying documentation and access to these features.

Upcomfing features may be enabled either individually, or through a new `chronosPreviewV4` catch-all designed to allow code to be prepared for increased strictness in future chronos releases.

`-d:chronosDebug` may be used to enable the existing debugging helpers together.

This PR was originally part of #350.


chronos/asyncfutures2.nim
```diff
@@ -8,9 +8,9 @@
 #    Apache License, version 2.0, (LICENSE-APACHEv2)
 #                MIT license (LICENSE-MIT)
 
-import std/sequtils
+import std/[sequtils, strutils]
```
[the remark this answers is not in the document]
arnetheduck: right, this slipped in from the futures PR in which it becomes necessary

tests/testaddress.nim
```diff
@@ -8,7 +8,7 @@
 import unittest2
 import ../chronos
 
-when defined(nimHasUsed): {.used.}
+{.used.}
```
cheatfate, line 11: I wish to leave here old variant which is more safe and maybe it helps our users to use `chronos` with older versions of Nim.

chronos/asyncfutures2.nim
arnetheduck, line 11:
```suggestion
import std/sequtils
```

chronos.nimble
```diff
@@ -32,10 +32,11 @@ proc run(args, path: string) =
 
 task test, "Run all tests":
   for args in [
-      "-d:useSysAssert -d:useGcAssert",
-      "-d:chronosStackTrace -d:chronosStrictException",
+      "-d:debug -d:chronosDebug",
+      "-d:debug -d:chronosPreviewV4",
```
cheatfate, line 36: -d:debug -d:useSysAssert -d:useGcAssert is also important, because i'm not sure if its available for `release` mode.

tests/testaddress.nim
arnetheduck: `used` exists - we support 1.2+ only, you won't be able to compile chronos with earlier versions

chronos.nimble
arnetheduck: bbd706446e4ed19c39353f728ef989b1e5bc5422

arnetheduck: bbd706446e4ed19c39353f728ef989b1e5bc5422

PR #374, opened 2023-04-03
Asyncproc (Part 3/3)

cheatfate:


chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+
+proc init*(t: typedesc[ProcessStreamHandle],
+           writer: AsyncStreamWriter): ProcessStreamHandle =
+  ProcessStreamHandle(kind: ProcessStreamHandleKind.StreamWriter,
+                      writer: writer)
+
+proc isEmpty*(handle: ProcessStreamHandle): bool =
+  handle.kind == ProcessStreamHandleKind.None
+
+proc suspend*(p: AsyncProcessRef): AsyncProcessResult[void] {.gcsafe.}
```
zah, line 254: procs with forward declarations will be assumed to `raises: [Exception]`, so consider adding explicit `raises:` annotation here and elsewhere.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+          let sres = setCurrentDirectory(workingDir)
+          if sres.isErr():
+            raiseAsyncProcessError("Unable to change current directory",
+                                   sres.error())
+          cres.get()
+        else:
+          ""
+
+      if AsyncProcessOption.EchoCommand in options:
+        echo getFullCommand(commandLine, commandArguments)
```
zah, line 881: This feature is quite unusual.
If the user needs it, it would be very easy to implement it as a wrapper of `startProcess`.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+        return err(OSErrorCode(res))
+    ok()
+
+  proc envToCStringArray(t: StringTableRef): cstringArray =
+    let itemsCount = len(t)
+    var
+      res = cast[cstringArray](alloc0((itemsCount + 1) * sizeof(cstring)))
+      i = 0
+    for key, value in pairs(t):
+      var x = key & "=" & value
```
zah, line 727: You can avoid the memory allocation here by directly copying the key and value in the target `cstring`.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+    let itemsCount =
+      block:
+        var res = 0
+        for key, value in envPairs(): inc(res)
+        res
+    var
+      res = cast[cstringArray](alloc0((itemsCount + 1) * sizeof(cstring)))
+      i = 0
+    for key, value in envPairs():
+      var x = key & "=" & value
```
zah, line 743: Ditto

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+    ok()
+
+  proc envToCStringArray(t: StringTableRef): cstringArray =
+    let itemsCount = len(t)
+    var
+      res = cast[cstringArray](alloc0((itemsCount + 1) * sizeof(cstring)))
+      i = 0
+    for key, value in pairs(t):
+      var x = key & "=" & value
+      res[i] = cast[cstring](alloc0(len(x) + 1))
```
zah, line 728: The zero-initialization of `alloc0` is unnecessary here. All bytes will be overwritten.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+        var res = 0
+        while not(isNil(arguments[res])): inc(res)
+        res
+    var res = @[command]
+    if length > 0:
+      for index in 1 ..< length:
+        res.add($arguments[index])
+    res.join(" ")
+
+  template exitStatusLikeShell(status: int): int =
```
zah, line 786: This template is a bit dangerous because it perform double evaluation of the `status` parameter.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+
+  proc getFullCommand(command: string, arguments: cstringArray): string =
+    let length =
+      if isNil(arguments):
+        0
+      else:
+        var res = 0
+        while not(isNil(arguments[res])): inc(res)
+        res
+    var res = @[command]
```
zah, line 780: Using `newSeqOfCap` here is an easy way to save some potential re-allocations in the loop below.

Technically, it's also possible to avoid many string and sequence allocations by directly copying the `cstrings` into a pre-allocated result string.

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+          value
+      mask =
+        block:
+          var res: Sigset
+          discard sigemptyset(res)
+          res
+
+    doCheck(posixSpawnAttrSetSigMask(attrs, mask))
+    if AsyncProcessOption.ProcessGroup in options:
+      doCheck(posixSpawnAttrSetPgroup(attrs, 0))
```
zah, line 643: What is this option trying to do? It sets the `pgid` of the created process to zero? Thus, allowing only a single process group to exist in the application?

I think the right API is to create a `ProcessGroup` object before spawning the process and to pass it as an input to `startProcess`. This would allow me to create separate process groups which I can wait for completion or terminate at will (as groups).

On Windows, I expect the same API to be mapped to [Job Objects](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects?redirectedfrom=MSDN).

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+  return res
+
+proc pid*(p: AsyncProcessRef): int =
+  ## Returns process ``p`` identifier.
+  when defined(windows):
+    cast[int](p.processId)
+  else:
+    int(p.processId)
+
+template processId*(p: AsyncProcessRef): int = pid()
```
zah, line 1323:
```suggestion
template processId*(p: AsyncProcessRef): int = pid(p)
```

chronos/asyncloop.nim
```diff
@@ -426,6 +441,159 @@ when defined(windows):
...
+    ## Result of this procedure call ``WaitableHandle`` should be closed using
+    ## closeWaitable() call.
+    ##
+    ## NOTE: This is private procedure, not supposed to be publicly available,
+    ## please use ``waitForSingleObject()``.
+    let
+      loop = getThreadDispatcher()
+      cell =
+        try:
+          system.protect(rawEnv(cb))
```
arnetheduck, line 482: is this interface documented somewhere? why is it needed, ie what does it do that `GC_ref` doesn't?

arnetheduck: also, looking at its interface, we would need to call `dispose` as well: https://github.com/nim-lang/Nim/blob/version-1-6/lib/system/gc_common.nim#L96

that said, this does like some cross-thread GC magic, which we're not stressing here yet, from what I can tell?

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
...
+#         Chronos' asynchronous process management
+#
+#  (c) Copyright 2022-Present Status Research & Development GmbH
+#
+#                Licensed under either of
+#    Apache License, version 2.0, (LICENSE-APACHEv2)
+#                MIT license (LICENSE-MIT)
+{.push raises: [Defect].}
+
+import std/[options, strtabs]
```
arnetheduck, line 11: we can use `Opt` instead, avoiding the need for this import (and the global symbols like `some` that it adds)

chronos/asyncproc.nim
```diff
@@ -0,0 +1,1323 @@
+#
+#         Chronos' asynchronous process management
+#
+#  (c) Copyright 2022-Present Status Research & Development GmbH
+#
+#                Licensed under either of
+#    Apache License, version 2.0, (LICENSE-APACHEv2)
+#                MIT license (LICENSE-MIT)
+{.push raises: [Defect].}
```
arnetheduck, line 9:
```suggestion
when (NimMajor, NimMinor) < (1, 4):
  {.push raises: [Defect].}
else:
  {.push raises: [].}
```

chronos/config.nim
```diff
@@ -32,6 +32,17 @@ when (NimMajor, NimMinor) >= (1, 4):
...
+
+    chronosProcShell* {.strdefine.}: string =
+      when defined(windows):
+        "cmd.exe"
+      else:
+        when defined(android):
+          "/system/bin/sh"
+        else:
+          "/bin/sh"
+      ## Default shell binary path.
```
arnetheduck, line 44: good to document how / where this gets used, it's kind of .. unusual .. to have to have a shell

chronos/asyncproc.nim
zah: I think the most useful aspect of this "Process Group" API is that it can help you kill the entire subtree of processes launched by a particular program that you are starting. By default, such indirect child processes will be orphaned on Posix, which is painful for the users.

The job objects on Windows offer the same functionality, with even stronger guarantees.

zah: A related common problem is this one:
https://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits

... but I'll grant you that there aren't very good cross-platform solutions that have precisely the same semantics everywhere.

chronos/asyncloop.nim
arnetheduck: ok, found the calls to dispose, hm

chronos/asyncloop.nim
```diff
@@ -294,6 +294,9 @@ when defined(windows):
       cb*: CallbackFunc
       errCode*: OSErrorCode
       bytesCount*: uint32
+      cell*: ForeignCell # we need this `cell` to protect our `cb` environment,
```
arnetheduck, line 297: so, from what I can tell this should not be needed after all, ie a simple `GC_ref` should be sufficient:

although the callback that is registered with `RegisterWaitForSingleObject` might be called from a different thread, the actual callback that we implement does not perform any memory allocations or other interactions with the garbage collector - all it does is re-post the same notification to the thread that ultimately owns the cb that this cell aims to protect.

chronos/asyncproc.nim
cheatfate: Fixed.

chronos/asyncproc.nim
cheatfate: Fixed.

chronos/asyncproc.nim
cheatfate: done

chronos/asyncproc.nim
cheatfate: done

chronos/asyncproc.nim
cheatfate: done

chronos/asyncproc.nim
cheatfate: Fixed.

chronos/asyncproc.nim
cheatfate: done.

chronos/asyncloop.nim
cheatfate: system.protect/dispose removed from the code.

chronos/asyncproc.nim
cheatfate: done

chronos/asyncproc.nim
cheatfate: Fixed.

chronos/config.nim
cheatfate: done.

chronos/asyncloop.nim
cheatfate: Fixed.

chronos/asyncloop.nim
```diff
@@ -433,6 +447,141 @@ when defined(windows):
...
+    ## Callback ``cb`` will be scheduled with ``udata`` parameter when
+    ## ``handle`` become signaled.
+    ##
+    ## Result of this procedure call ``WaitableHandle`` should be closed using
+    ## closeWaitable() call.
+    ##
+    ## NOTE: This is private procedure, not supposed to be publicly available,
+    ## please use ``waitForSingleObject()``.
+    let loop = getThreadDispatcher()
+    var ovl = RefCustomOverlapped(data: CompletionData(cb: cb))
```
arnetheduck, line 487:
```suggestion
    let ovl = RefCustomOverlapped(data: CompletionData(cb: cb))
```

chronos/asyncloop.nim
```diff
@@ -433,6 +447,141 @@ when defined(windows):
...
+
+    var whandle = (ref PostCallbackData)(
```
arnetheduck, line 489:
```suggestion
    let whandle = (ref PostCallbackData)(
```

arnetheduck (approved):

arnetheduck:
Nice!

PR #378, opened 2023-04-13
drop support for Nim 1.2, 1.4, add Nim 2.0 to CI testing matrix

tersec:


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             cpu: amd64
           #- os: windows
             #cpu: i386
-        branch: [version-1-2, version-1-4, version-1-6, devel]
```
arnetheduck, line 29: we could wait with the 1.2/1.4 deprecations until the (moderately breaking) 4.0 release

tersec: Why though?

arnetheduck: because right now, there exists no compelling and active reason to drop it - ie 1.2 is not blocking any particular feature that we have planned for the near future - however, there exists a body of existing work that potentially could start breaking things (eh tracking etc), so the plan would be to tag an "official" release for those that want to use 1.2 and then drop support.

This would set the stage for a "regular" way to communicate and document nim version compatibility in libraries in general, ie "chronos 3.x supports nim 1.2, chronos 4.x supports nim 1.6+" etc

it also creates an environment in which for example there's a straight way to handle security patching: we would in such a scenario be able to create a 3.x branch with further fixes.

All this work is in the interest of attracting long-term investments into chronos (and librarieS), ie fulfill the requirement that any larger project has, namely that there exists a viable and documented maintenance / deprecation roadmap (similar to lts releases) that doesn't include arbitrary rug-pulls (a hobbyist will not mind upgrading stuff incessantly - a production project however is a different story where upgrades are costs that need to be managed).

tersec: One concern is that often, there's been some way to work around Nim 1.2/1.4 compatibility concerns, even if it's just `when (NimMajor, NimMinor) >= (1, Foo)`, but that every time these mechanisms are used, it increases complexity, technical debt, and decreasingly tested codepaths, while increasing maintenance burden. Without specific use cases where, say, 1.2 or 1.4 (unclear which is worth removing first, if it's not simultaneous -- 1.4 was never a long-term supported release, for example) cited/listed, or a specific timeline of removal, these end up looking like rhetorical techniques to indefinitely delay removal of expensive contingencies for which I haven't seen specific evidence of utility.

So this proposed criterion of, well, "there exists no compelling and active reason to drop it" can be perpetually argued to be true as long as one is willing to continue to add essentially unnecessary codebase complexity, with no specific threshold where, oops, no, this is too much ever occurs. Since I haven't seen evidence from the keep-1.2/1.4-indefinitely faction that this threshold exists, I don't yet believe it does, and won't act on that basis. Therefore, I'll continue to advocate for removing 1.2/1.4 support as soon as feasible.

arnetheduck: > Since I haven't seen evidence from the keep-1.2/1.4-indefinitely faction that this threshold exists,

Uh, 4.0 (which is imminent), as noted several times here, would be that threshold *for chronos* - it is based on real, actual changes that would benefit from the dropped support.

The aim is exactly to make the threshold explicit, documented and predictable, as well as provide a clear path for future decisions of this kind so that we don't have to rehash it, and so that it isn't arbitrary.

Essentially, what this policy says is that dropping support for a nim version is a "breaking change" as far as API/ABI compatibility goes and we release a new "major" version for it - unlike to "I fear unspecified maintenance, therefore I will remove support arbitrarily".

The benefit of this documented policy is that it also provides a predictable path for third-parties to bake into their own maintenance schedules - they don't have to worry about the potential for critical security patches (at that point, there exists a well-known mechanism to distribute them to past "versions"). More than anything else, this becomes a communication tool (just like LTS releases usually are joined by criteria for when that support ends, rather than "when I feel like it").

"as soon as feasible" is subjective based your limited perception of what "feasible" means - it's a not a good asynchronous communication mechanism outside of a tight-knit group.

there's a critical difference here as well: chronos has reached "maturity" in the sense that we maintain it as a post-1.0 library - before 1.0, it's "expected" that code comes with no stability or versioning guarantees - post-1.0, it's common to provide a measure of such guarantees precisely so that people asynchronously via well-established common practice can decide.

cheatfate: I'm not hiding anything in here, i just dont understand your politic, because in DM you saying that there no reason to support 1.2 and 1.4 and on public you have absolutely different opinion.

arnetheduck: In DM I mentioned we should consider it - but let's do it in an orderly and well-documented way, that's all

cheatfate: So, we should consider it here and now, because there is no need to increase CI time at all, we should either remove `1.2 and 1.4` or add `2.0`. I dont see any reasons to run so many targets right now.

arnetheduck: well, my suggestion remains: maintain 1.2 compat for the duration of the 3.x series, merge the outstanding compatible pr:s (asyncproc mainly at this point), branch 3.x into a "stable" series in which we _may_ publish security fixes in the future then focus on 4.0 as a 1.6+-only branch with new features.

I don't have a strong opinion on whether 3.x should be 2.0-compatible or not, but probably not worth it

arnetheduck:
so, how about we just make this PR do what it set out to do: add 2.0.

We can move the 1.2 discussion elsewhere, to a separate PR which removes past versions (which should get merged soon enough as well)

PR #380, opened 2023-04-25
disallow reentrancy

arnetheduck:
Reentrancy causes event reordering and stack explosions, in addition to leading to hard-to-debug scenarios.

Since none of chronos is actually tested under reentrant conditions (ie that all features such as exceptions, cancellations, buffer operations, timers etc work reliably when the loop is reentered), this is hard to support over time and prevents useful optimizations - this PR simply detects and disallows the behaviour to remove the uncertainty, simplifying reasoning about the event loop in general.

### DIFF
--- a/chronos/internal/asyncengine.nim
+++ b/chronos/internal/asyncengine.nim
@@ -65,19 +65,22 @@ type
     ticks*: Deque[AsyncCallback]
     trackers*: Table[string, TrackerBase]
     counters*: Table[string, TrackerCounter]
-
-proc sentinelCallbackImpl(arg: pointer) {.gcsafe, noreturn.} =
-  raiseAssert "Sentinel callback MUST not be scheduled"
-
-const
-  SentinelCallback = AsyncCallback(function: sentinelCallbackImpl,
-                                   udata: nil)
-
-proc isSentinel(acb: AsyncCallback): bool =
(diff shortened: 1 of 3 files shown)


arnetheduck:
Marked draft to merge into 4.0

chronos/asyncloop.nim
```diff
@@ -220,6 +212,17 @@ func getAsyncTimestamp*(a: Duration): auto {.inline.} =
...
+template preparePoll(loop: PDispatcherBase) =
+  # If you hit this assert, you've called `poll`, `runForever` or `waitFor`
+  # from within an async function which is not supported due to the difficulty
+  # to control stack depth and event ordering
+  # If you're using `waitFor`, switch to `await` and / or propagate the
+  # up the call stack.
+  doAssert not loop.polling, "The event loop and chronos functions in general are not reentrant"
+
+  loop.polling = true
+  defer: loop.polling = false
```
zah, line 224: I believe Chronos already needs to wrap every call to user-land in try/except, so why is it not possible to just set this flag there, just before entering user-land?

arnetheduck: the check applies to async callbacks too which don't go through the async transformation where said finally appears :/

arnetheduck:
superseded by #689

PR #381, opened 2023-04-26
Fix cast[pointer] issues on nim-devel.

cheatfate:
This PR introduces fixes for latest `devel` changes, one of the fixes requires to have Posix error codes as constants.
* Introduced new file oserrno.nim which declares error constants for FreeBSD, OpenBSD, NetBSD, DragonflyBSD, MacOS and Linux.
* Fixed cast[pointer] warnings and errors.
* Fix usage of error constants.
* Deprecate `winlean` usage.
* Fix MacOS osnet.net issue.

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -780,8 +780,8 @@ when defined(windows):
         var ovl = cast[RefCustomOverlapped](udata)
         if not(retFuture.finished()):
           if ovl.data.errCode == OSErrorCode(-1):
-            if setsockopt(SocketHandle(sock), cint(osdefs.SOL_SOCKET),
-                          cint(osdefs.SO_UPDATE_CONNECT_CONTEXT), nil,
+            if setsockopt(SocketHandle(sock), cint(SOL_SOCKET),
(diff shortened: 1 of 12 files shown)


chronos/transports/stream.nim
```diff
@@ -322,10 +322,12 @@ when defined(windows):
     (t).wwsabuf.buf = cast[cstring](v.buf)
     (t).wwsabuf.len = cast[ULONG](v.buflen)
 
-  proc isConnResetError(err: OSErrorCode): bool {.inline.} =
-    result = (err == OSErrorCode(osdefs.WSAECONNRESET)) or
-             (err == OSErrorCode(osdefs.WSAECONNABORTED)) or
-             (err == OSErrorCode(osdefs.ERROR_PIPE_NOT_CONNECTED))
+  func isConnResetError(err: OSErrorCode): bool {.inline.} =
+    case err
```
arnetheduck, line 326: also `err in [WSAECONNRESET, WSAECONNABORTED, ERROR_PIPE_NOT_CONNECTED]`

PR #390, opened 2023-05-18
fix missing sigprocmask import

bung87:
the devel default enable threads, so it will not occurs on devel


bung87:
please also bump a new version as it's a bug fix for nim stable version.

PR #393, opened 2023-05-24
Future cleanup

arnetheduck:
* FutureState.Finished -> FutureState.Completed (to avoid name clash with `proc finished` which means not-pending)
* deprecate `done` - to avoid additional confusion over completed vs finished
* remove ad leftovers in stack trace formatting

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -26,7 +26,9 @@ when chronosStackTrace:
 const
   LocCreateIndex* = 0
   LocFinishIndex* = 1
-  LocCompleteIndex* = LocFinishIndex  ## Deprecated
+
+template LocCompleteIndex*: untyped {.deprecated: "LocFinishIndex".} =
+  LocFinishIndex
 
 type
   FutureState* {.pure.} = enum
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -2,7 +2,7 @@
 #                     Chronos
 #
 #  (c) Copyright 2015 Dominik Picheta
-#  (c) Copyright 2018-2021 Status Research & Development GmbH
+#  (c) Copyright 2018-2023 Status Research & Development GmbH
 #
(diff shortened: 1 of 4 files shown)


arnetheduck:
This PR introduces some of the refactoring done in #373 to simplify discussion around it

cheatfate:
If you want to be so strict about terms `finished` and `completed` - you should correct all the places where this terms are equal:
For example:
1. `LogCompleteIndex` should become `LogFinishIndex` because it records when Future become `Finished` and not only `Completed`.
2. All the documentation comments we have in there are also using "finished == completed" term. For example
https://github.com/status-im/nim-chronos/blob/2d82a4319ea34303096c428f50dd0c0d83c861a4/chronos/asyncfutures2.nim#L156-L158 or this one https://github.com/status-im/nim-chronos/blob/2d82a4319ea34303096c428f50dd0c0d83c861a4/chronos/asyncfutures2.nim#L170 Because after we merge this PR sentence "completed without error" will mean "completed without error without error".

chronos/asyncfutures2.nim
```diff
@@ -55,10 +60,15 @@ type
...
+        closure*: iterator(f: Future[T]): FutureBase {.raises: [CatchableError], gcsafe.}
     else:
-      closure*: iterator(f: Future[T]): FutureBase {.raises: [Defect, CatchableError, Exception], gcsafe.}
-    value: T ## Stored value
+      closure*: iterator(f: Future[T]): FutureBase {.raises: [Exception], gcsafe.}
+
+    when T isnot void:
+      value*: T ## Stored value
 
   FutureStr*[T] = ref object of Future[T]
```
cheatfate, line 73: This could be deprecated.

arnetheduck: I agree but it's beyond the scope of this PR: they are widely used in chronos itself so deprecating them here and now would result in a lot of warnings.

cheatfate (approved):

PR #395, opened 2023-05-26
Memory cleanups

arnetheduck:
* Avoid `FutureContinue` bloat by moving type punning into iterator (which is typed anyway)
* ~~call cancellation callback also on futures with children (potentially fixes ratelimiter leak)~~
* clear closure iterator from future when iteration is done also on cancellation / exception
* remove some redundant local variables in `await`


cheatfate:
Cancellation callback MUST be deprecated and not definitely it MUST not be called on Futures with children, because in this case exception CancelledError is going to be generated by the children, otherwise you creating multiple points of cancellation which will introduce UB.

cheatfate (approved):

PR #398, opened 2023-06-01
Recover cancellation cleanup for AsyncEvent wait().

cheatfate:


chronos/asyncsync.nim
```diff
@@ -223,12 +223,16 @@ proc wait*(event: AsyncEvent): Future[void] =
   ## If the internal flag is `true` on entry, return immediately. Otherwise,
   ## block until another task calls `fire()` to set the flag to `true`,
   ## then return.
-  var w = newFuture[void]("AsyncEvent.wait")
+  let retFuture = newFuture[void]("AsyncEvent.wait")
+  proc cancellation(udata: pointer) {.gcsafe, raises: [Defect].} =
+    if not(retFuture.finished()):
```
arnetheduck, line 228: what's this check for? cancelcb should not be called on finished futures?

cheatfate: All the callbacks we attaching to the Future[T] via `addCallback` or `cancelCallback=` should check their state before applying action. In this specific case this check is possibly redundant.

cheatfate: So for example it is absolutely possible that there 2 callbacks got scheduled by `poll()` one performs completing this Future, but another could be cancellation. So when we got the flow in here we should check if the Future is in expected state, e.g. Pending.

arnetheduck: but we prevent that in `complete`/`finish`? it seems like an important invariant, that once the callback is scheduled the future is frozen and cannot change any more

arnetheduck (approved):

PR #349, opened 2023-01-19
Support implicit returns

arnetheduck:
Based on https://github.com/nim-lang/Nim/pull/20933

Bugs that we might hit:

* https://github.com/nim-lang/Nim/issues/21282
* https://github.com/status-im/nim-stew/pull/167
* https://github.com/nim-lang/Nim/issues/19849

PR #401, opened 2023-06-05
Support implicit returns

arnetheduck:
* based on https://github.com/nim-lang/Nim/pull/21898
* also fixes generic Future[T] where T ends up being `void`

supercedes #349 & #328

### DIFF
--- a/chronos/asyncmacro2.nim
+++ b/chronos/asyncmacro2.nim
@@ -9,46 +9,73 @@
 
 import std/[macros]
 
-proc skipUntilStmtList(node: NimNode): NimNode {.compileTime.} =
-  # Skips a nest of StmtList's.
-  if node[0].kind == nnkStmtList:
-    skipUntilStmtList(node[0])
+# `quote do` will ruin line numbers so we avoid it using these helpers
+proc completeWithResult(fut, baseType: NimNode): NimNode {.compileTime.} =
+  # when `baseType` is void:
+  #   complete(`fut`)
+  # else:
+  #   complete(`fut`, result)
+  if baseType.eqIdent("void"):
+    # Shortcut if we know baseType at macro expansion time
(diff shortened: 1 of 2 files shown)


chronos/asyncmacro2.nim
```diff
@@ -9,46 +9,73 @@
...
+      ),
+      nnkElseExpr.newTree(
+        newCall(ident "complete", fut, ident "result")
+      )
+    )
+
+proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
+  #   when typeof(`node`) is void:
+  #     `node` # statement / explicit return
+  #     complete(fut, result)
```
cheatfate, line 36: why not `complete(fut)`?

tests/testmacro.nim
```diff
@@ -123,6 +123,24 @@ suite "Macro transformations test suite":
...
+    proc gen(T: typedesc): Future[T] {.async.} =
+      proc testproc(): Future[T] {.async.} =
+        when T is void:
+          return
+        else:
+          return default(T)
+      await testproc()
+
+    waitFor gen(void)
+    discard waitFor gen(int)
```
cheatfate, line 136: We testing implicit return value with `discard`?

cheatfate: So check it for `default(t) == waitFor gen(int)`.

cheatfate: Maybe with some crazy optimization this check would be removed?

arnetheduck: 77219d5b1aac7f581bc1e9f19b4a83c44bb90aae

cheatfate:
I think test suite is not complete, i wish to see `case`, `block`, `try/except/finally` in tests.

cheatfate (approved):

PR #406, opened 2023-06-08
Asynchronous thread notification mechanism.

cheatfate:


chronos/threadsync.nim
```diff
@@ -0,0 +1,344 @@
...
+        retFuture.fail(newException(AsyncError, osErrorMsg(errorCode)))
+    elif res != sizeof(data):
+      retFuture.fail(newException(AsyncError, osErrorMsg(EINVAL)))
+    else:
+      retFuture.complete()
+
+  retFuture
+
+when defined(windows):
+  proc wait*(signal: ThreadSignalPtr) {.async.} =
```
arnetheduck, line 292: hm, what are the calling constraints here for `fire` and `wait`? ie which threads can they each be called from?

what happens for example if I `wait` from multiple threads?

cheatfate: So for any number callers to `fire()/fireSync()` from single or many threads and for any number of waiters via `wait()/waitSync()` for single or many thread, every call to `fire()/fireSync()` will wake up only one waiter via `wait()/waitSync()`, but ThreadSignal is not a semaphore so there could be only one pending `fire()` signal until waiter will not receive it.

chronos/threadsync.nim
```diff
@@ -0,0 +1,416 @@
...
+          cint(signal[].efd)
+        else:
+          cint(signal[].wfd)
+      checkFd =
+        when defined(linux):
+          cint(signal[].efd)
+        else:
+          cint(signal[].rfd)
+
+    if checkBusy(checkFd):
```
arnetheduck, line 186: is it thread-safe to skip the writing here? if another thread reads just after this check, it might reset the conditions for the "writer" to fire again

arnetheduck (approved):
Tried integrating this both with a channel and with nim-taskpools - works fine :+1:

PR #407, opened 2023-06-12
[WIP] introduce the virtual clock for simulations

cskiraly:
This allows running code written for Chronos using
`-d:asyncTimer=virtual`
, turning it into a simple event-based simulation.

Used in https://github.com/codex-storage/nim-libp2p-dht/pull/57

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -781,8 +781,14 @@ elif unixPlatform:
     # Moving expired timers to `loop.callbacks` and calculate timeout.
     loop.processTimersGetTimeout(curTimeout)
 
+    let timeout =
+      when asyncTimer == "virtual":
+        0
+      else:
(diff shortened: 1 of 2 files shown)


cskiraly:
I use this virtual timer for testing performance in https://github.com/codex-storage/das-dht-emulator, using UDP only.

The problem it solves for me is the CPU bottleneck when emulating many instances of a process on a single machine.

Normally, CPU gets bottlenecked, and you can't keep up with real time clock. If you use this timer, all CPU operations become 0 virtual time, and thus you can focus on delays and timeouts set up in the event loop. Of course there are more sophisticated approaches, but it achieves what I need with a simple patch, and mostly without modifications to the code on top.

cskiraly:
@cheatfate I have largely simplified the code. Purpose is still the same: to run simulations without code change.

PR #408, opened 2023-06-16
Less flaky rate limit tests

Menduist:
Let the caller specify the "now" (which outside of tests, can be useful to avoid querying the clock multiple times, if the caller already knows the `now`)

In sync test, modify the `now` parameter instead of sleeping to avoid flakyness
In async tests, this is unfortunately not possible since the `worker` will actually sleep to wait for the bucket to fill

### DIFF
--- a/chronos/ratelimit.nim
+++ b/chronos/ratelimit.nim
@@ -33,7 +33,9 @@ proc update(bucket: TokenBucket, currentTime: Moment) =
     bucket.budget = min(bucket.budgetCap, bucket.budget)
     return
 
-  assert currentTime >= bucket.lastUpdate
+  if currentTime < bucket.lastUpdate:
+    return
+
   let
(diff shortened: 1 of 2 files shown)


chronos/ratelimit.nim
```diff
@@ -28,13 +28,13 @@ type
     pendingRequests: seq[BucketWaiter]
     manuallyReplenished: AsyncEvent
 
-proc update(bucket: TokenBucket) =
+proc update(bucket: TokenBucket, currentTime: Moment) =
   if bucket.fillDuration == default(Duration):
     bucket.budget = min(bucket.budgetCap, bucket.budget)
     return
 
+  assert currentTime >= bucket.lastUpdate
```
arnetheduck, line 36: this might be a bit harsh - we can just treat it as 0

arnetheduck (approved):

PR #411, opened 2023-06-26
fix EpollEvent memory layout

arnetheduck:
on x86_64, `EpollEvent` is packed - the upstream version is correct here

### DIFF
--- a/chronos/osdefs.nim
+++ b/chronos/osdefs.nim
@@ -945,8 +945,6 @@ elif defined(linux):
                         SIGPIPE, SIGALRM, SIGTERM, SIGPIPE, SIGCHLD, SIGSTOP,
                         SIGCONT
 
-  from std/epoll import EpollData, EpollEvent
-
   export close, shutdown, sigemptyset, sigaddset, sigismember,
          sigdelset, write, read, waitid, getaddrinfo,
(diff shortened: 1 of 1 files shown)


chronos/osdefs.nim
```diff
@@ -1001,13 +1001,21 @@ elif defined(linux):
...
+    EpollData* {.importc: "epoll_data_t",
+        header: "<sys/epoll.h>", pure, final, union.} = object
+      `ptr`* {.importc: "ptr".}: pointer
+      fd* {.importc: "fd".}: cint
+      u32* {.importc: "u32".}: uint32
       u64* {.importc: "u64".}: uint64
 
-    EpollEvent* {.importc: "struct epoll_event", header: "<sys/epoll.h>",
-                  pure, final.} = object
+    EpollEvent* {.importc: "struct epoll_event", header: "<sys/epoll.h>", pure, final, epollPacked.} = object
```
cheatfate, line 1018: Why we need 105 characters per line here?

arnetheduck: makes it easier to compare for bugs with upstream, from where this code is copy-pasted

cheatfate: Please adjust it to 80 characters per line.

PR #412, opened 2023-06-27
futures: sinkify

arnetheduck:
this avoids copies here and there throughout the pipeline - ie `copyString` and friends can often be avoided when moving things into and out of futures

annoyingly, one has to sprinkle the codebase liberally with `sink` and `move` for the pipeline to work well - sink stuff _generally_ works better in orc/arc


arnetheduck:
blocked on https://github.com/nim-lang/Nim/issues/22175

arnetheduck:
superseded by #475 which includes a workaround for the bug

PR #413, opened 2023-06-27
futures: lentify

arnetheduck:
sometimes avoid copies when reading from `Future`

extracted from https://github.com/status-im/nim-chronos/pull/412

### DIFF
--- a/chronos/asyncfutures2.nim
+++ b/chronos/asyncfutures2.nim
@@ -451,19 +451,25 @@ proc internalCheckComplete*(fut: FutureBase) {.raises: [CatchableError].} =
       injectStacktrace(fut.internalError)
     raise fut.internalError
 
-proc internalRead*[T](fut: Future[T]): T {.inline.} =
-  # For internal use only. Used in asyncmacro
-  when T isnot void:
-    return fut.internalValue
+proc read*[T: not void](future: Future[T] ): lent T {.raises: [CatchableError].} =
+  ## Retrieves the value of ``future``. Future must be finished otherwise
+  ## this function will fail with a ``ValueError`` exception.
+  ##
+  ## If the result of the future is an error then that error will be raised.
+  if not future.finished():
+    # TODO: Make a custom exception type for this?
+    raise newException(ValueError, "Future still in progress.")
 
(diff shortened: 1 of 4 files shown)


chronos/asyncfutures2.nim
```diff
@@ -451,19 +451,25 @@ proc internalCheckComplete*(fut: FutureBase) {.raises: [CatchableError].} =
...
   ## Retrieves the value of ``future``. Future must be finished otherwise
   ## this function will fail with a ``ValueError`` exception.
   ##
   ## If the result of the future is an error then that error will be raised.
   if future.finished():
     internalCheckComplete(future)
-    internalRead(future)
   else:
     # TODO: Make a custom exception type for this?
     raise newException(ValueError, "Future still in progress.")
```
cheatfate, line 475: As for me attempt to read value data from pending `Future` is `Defect`, its like reading result from failed `Result[T]` object.

arnetheduck: the exception is a matter for a separate PR, but per our earlier discussion we were going to introduce a `FuturePendingError` here: https://github.com/status-im/nim-chronos/pull/373/files#diff-eee8cf24fd606aea1d32159dd2926d1a1e910af87c3de3de484374f0ca88db5aR100

I'm still planning on creating that PR.

Right now, `value` is the function to use in a `Result`-like way with a `Defect`. `read` is more similar to `tryValue` in Result which also raises. I've been thinking about renaming the function in `Result` to `read` as well then we would have the same behavior - `value` would require pre-checking state while `read` would raise catchable exception.


I guess the contention point here is that `Pending` is a special state of the `Future` that does not exist in `Result`. Similar to cancellation, mixing user-origin exceptions with those of the future itself leads to information loss - I'm not sure there's a great way to fix this though without reworking cancellation :/

cheatfate: So if we going to make this change i think we should not raise `ValueError`. There should be `Defect` or at least `AsyncError`.

arnetheduck: I'd prefer to see that done in a separate PR - ie the objective of this PR is to only address `lent` and nothing else (the control flow changes are needed to make `lent` work but there should be no changes vs the status quo in observable behavior)

PR #418, opened 2023-07-11
Finish closures before completing futures (fix #415)

### LINKED ISSUE #415: return statement in try/finally block breaks executing multiple await statements in finally block
#### Reproducible source

```nim
import chronos

type
  ClientResponse = object
    status*: int
    data*: string

proc subFoo1(): Future[int] {.async.} =
  return 200

proc subFoo2(): Future[string] {.async.} =
  return "SOMEDATA"

proc testFoo(): Future[ClientResponse] {.async.} =
  try:
    let status = await subFoo1()
    doAssert(status == 200)
    let data = await subFoo2()
    return ClientResponse(status: status, data: data)
  finally:
    echo "stage 1"
    echo "stage 2"
    await sleepAsync(200.milliseconds)
    echo "stage 3"

when isMainModule:
  echo waitFor testFoo()
```

#### Current output
```
stage 1
(status: 200, data: "SOMEDATA")
```

#### Expected output
```
stage 1
stage 2
stage 3
(status: 200, data: "SOMEDATA")
```

#### Workaround
If you change `return ClientResponse(status: status, data: data)` to `result = ClientResponse(status: status, data: data)` or implicitly return `ClientResponse(status: status, data: data)` all statements in `finally` will be executed.

Menduist:
The idea of this PR would be that the closure never completes the future, instead just writes to the internalValue, and the `futureContinue` would take care of completing the future (for now I've just applied this to the `return` rewrite, but could be used everywhere)
Besides fixing #415, it could also avoid a copy to the internalValue once the future is finished

However, might have some downsides


chronos/asyncfutures2.nim
```diff
@@ -317,6 +317,9 @@ proc futureContinue*(fut: FutureBase) {.raises: [], gcsafe.} =
       # `await` typically) or completes / fails / is cancelled
       next = fut.internalClosure(fut)
       if fut.internalClosure.finished(): # Reached the end of the transformed proc
+        if not fut.finished:
```
arnetheduck, line 320: hm, this is actually a really good point: we should not finish the future until the iterator itself has finished or an order inversion will indeed happen! it deserves to be documented.

it means that the above should not be a conditional but rather the norm - we should always arrive at this point with an unfinished future.

That in turn means that the expression return stuff needs to be rewritten as well to assign to value rather than calling `complete` - ditto error paths which need to set the error field but not complete the future etc..

arnetheduck: it's actually a classic bug not to run iterators until completion and thus missing part of the code - the same happens with simple `for` loop iterators that have code after the last `yield`: if you break the for loop that code will not run..

chronos/asyncfutures2.nim
```diff
@@ -317,6 +317,8 @@ proc futureContinue*(fut: FutureBase) {.raises: [], gcsafe.} =
       # `await` typically) or completes / fails / is cancelled
       next = fut.internalClosure(fut)
       if fut.internalClosure.finished(): # Reached the end of the transformed proc
+        #TODO hacky
```
arnetheduck, line 320: this needs a comment explaining why it's important that the future remains unfinished until the iterator is done (ie to run `finally` code fully, or really, any code after the last `yield`) - it's not actually hacky, it follows the way the compiler generates code for result - ie `return xxx` is typically rewritten as `result = xxx; return` - it's nuts but here we are.

chronos/asyncmacro2.nim
```diff
@@ -10,34 +10,13 @@
...
 proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
   #   when typeof(`node`) is void:
   #     `node` # statement / explicit return
-  #     -> completeWithResult(fut, baseType)
   #   else: # expression / implicit return
-  #     complete(`fut`, `node`)
-  if node.kind == nnkEmpty: # shortcut when known at macro expanstion time
-    completeWithResult(fut, baseType)
+  #     result = `node`
+  if node.kind == nnkEmpty: # shortcut when known at macro expantion time
```
arnetheduck, line 18:
```suggestion
  if node.kind == nnkEmpty: # shortcut when known at macro expansion time
```

arnetheduck:
funny enough, this PR should also improve performance because we will no longer generate a temporary variable that must be zero:ed then copied to the future

chronos/asyncfutures2.nim
arnetheduck: > setFinishLocation

or just use `internalLocation[...]` - ie none of this should be exposed in any public functions or API so it doesn't greatly matter - we can design a better public API for these kinds of things once the dust has settled

chronos/asyncfutures2.nim
Menduist: Done

chronos/asyncmacro2.nim
Menduist: That's team work :smile:

chronos/asyncfutures2.nim
Menduist: Done

Menduist: Note that this slightly changed the Finish location (the `proc` line instead of the first line of the proc)
Not sure this greatly matters

Menduist:
This branch is failing on nimbus:
```
/home/tavon/Status/nimbus-eth2/vendor/nim-presto/presto/serverprivate.nim(53, 14) template/generic instantiation of `async` from here
/home/tavon/Status/nimbus-eth2/vendor/nim-presto/presto/serverprivate.nim(212, 33) Error: undeclared identifier: 'defaultResponse'
make: *** [Makefile:437: nimbus_signing_node] Error 1
make: *** Waiting for unfinished jobs....
/home/tavon/Status/nimbus-eth2/beacon_chain/el/el_manager.nim(724, 18) template/generic instantiation of `isnot` from here
/home/tavon/Status/nimbus-eth2/beacon_chain/el/el_manager.nim(717, 42) Error: cannot instantiate Future [type declared in /home/tavon/Status/nimbus-eth2/vendor/nim-chronos/chronos/futures.nim(68, 3)]
got: <T>
but expected: <T>
make: *** [Makefile:374: block_sim] Error 1
/home/tavon/Status/nimbus-eth2/beacon_chain/spec/eth2_apis/rest_common.nim(48, 17) template/generic instantiation of `isnot` from here
/home/tavon/Status/nimbus-eth2/beacon_chain/spec/eth2_apis/rest_common.nim(38, 27) Error: cannot instantiate Future [type declared in /home/tavon/Status/nimbus-eth2/vendor/nim-chronos/chronos/futures.nim(68, 3)]
got: <T>
but expected: <T>
```

Some generic fun to fix

chronos/asyncmacro2.nim
```diff
@@ -173,34 +164,12 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
-              nnkVarSection.newTree(nnkIdentDefs.newTree(
-                nnkPragmaExpr.newTree(
-                  ident "result",
-                  nnkPragma.newTree(ident "used")),
-                baseType, newEmptyNode())
-                ),
-              quote do: {.pop.},
-            )
+            quote do:
+              template result: auto {.used.} = `castFutureSym`.internalValue
```
cheatfate, line 168: Whatever you want to do here, please avoid `quote do` statements because this statements makes all our stack traces to point into asyncmacro2.nim file instead of pointing to actual source.

chronos/asyncmacro2.nim
```diff
@@ -48,21 +27,33 @@ proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
...
     )
 
 proc processBody(node, fut, baseType: NimNode): NimNode {.compileTime.} =
   #echo(node.treeRepr)
   case node.kind
   of nnkReturnStmt:
-    let
-      res = newNimNode(nnkStmtList, node)
-    res.add completeWithNode(fut, baseType, processBody(node[0], fut, baseType))
+    let res = newNimNode(nnkStmtList, node)
```
cheatfate, line 41: Please add the comment here with generated source code.

arnetheduck:
this should have a test that shows the previously broken code snippet

chronos/asyncmacro2.nim
```diff
@@ -48,21 +27,33 @@ proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
           ident "is", nnkTypeOfExpr.newTree(node), ident "void"),
         newStmtList(
           node,
-          completeWithResult(fut, baseType)
         )
       ),
       nnkElseExpr.newTree(
-        newCall(ident "complete", fut, node)
+        newAssignment(ident "result", node)
```
arnetheduck, line 33: this should directly use `internalValue` rather than going via the template (more templates = more problems)

chronos/asyncmacro2.nim
```diff
@@ -48,21 +27,33 @@ proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
...
-    let
-      res = newNimNode(nnkStmtList, node)
-    res.add completeWithNode(fut, baseType, processBody(node[0], fut, baseType))
+    let res = newNimNode(nnkStmtList, node)
+    if node[0].kind != nnkEmpty:
+      res.add nnkWhenStmt.newTree(
+        nnkElifExpr.newTree(
+          nnkInfix.newTree(
+            ident "isnot", baseType, ident "void"),
+          newAssignment(ident "result", node[0])
```
arnetheduck, line 47: ditto, avoid result - also, `node[0]` needs to pass through `processBody`

arnetheduck: for the test suite:

```nim
return if 1 == 1:
  return 42
else:
  42
```

ie nested returns are valid code

chronos/asyncmacro2.nim
arnetheduck: and processBody..

Menduist: This proc is only used with `procBodyBlck` which is already went through `processBody`

arnetheduck: right

Menduist: > more templates = more problems

That's unfortunately true, since using internalValue instead of the template fixed some compilation issue in nimbus
Looks like something that might bite us later

chronos/asyncmacro2.nim
```diff
@@ -154,14 +152,19 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
+      procBody = prc.body.processBody(resultNode, baseType)
 
     # don't do anything with forward bodies (empty)
     if procBody.kind != nnkEmpty:
       let
         # fix #13899, `defer` should not escape its original scope
         procBodyBlck = nnkBlockStmt.newTree(newEmptyNode(), procBody)
 
+        # workaround https://github.com/nim-lang/Nim/issues/22645
+        internalFutureTypeSym = genSym(nskType, "internalFutureTypeSym")
```
arnetheduck, line 165:
```suggestion
        internalFutureTypeSym = genSym(nskType, "InternalFutureType")
```

chronos/asyncfutures2.nim
```diff
@@ -165,6 +165,12 @@ template complete*(future: Future[void]) =
   ## Completes a void ``future``.
   complete(future, getSrcLocation())
 
+proc completeWithInternalData(future: FutureBase) =
```
arnetheduck, line 168:
```suggestion
proc completeWithInternalValue(future: FutureBase) =
```

chronos/asyncmacro2.nim
```diff
@@ -10,34 +10,13 @@
...
-        nnkInfix.newTree(ident "is", baseType, ident "void"),
-        newCall(ident "complete", fut)
-      ),
-      nnkElseExpr.newTree(
-        newCall(ident "complete", fut, ident "result")
-      )
-    )
-
-proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
+proc completeWithNode(node, resultNode: NimNode): NimNode {.compileTime.} =
```
arnetheduck, line 13: assignOrProcessNode
```suggestion
proc assignOrProcessNode(node, resultNode: NimNode): NimNode {.compileTime.} =
```

chronos/asyncmacro2.nim
```diff
@@ -48,21 +27,40 @@ proc completeWithNode(fut, baseType, node: NimNode): NimNode {.compileTime.} =
...
+        nnkElifExpr.newTree(
+          nnkInfix.newTree(
+            ident "isnot", baseType, ident "void"),
+          newAssignment(resultNode, processBody(node[0], resultNode, baseType))
+        ),
+        nnkElse.newTree(
+          nnkPragma.newTree(
+            nnkExprColonExpr.newTree(
+              newIdentNode("error"),
+              newLit("no async return type declared"))
```
cheatfate, line 60: This is confusing error message for everybody, even i do not understand what it mean.

tests/testmacro.nim
```diff
@@ -136,6 +136,27 @@ suite "Macro transformations test suite":
...
+
+    check waitFor(nr()) == 42
+
+  test "Issue #415 (run closure to completion on return)":
+    var x = 0
+    proc test415 {.async.} =
+      try:
+        return
+      finally:
+        await stepsAsync(1)
```
cheatfate, line 155: I want to see more than one `await` in `finally`. and also i want to see how it works in exception handler, e.g. more than one `await`. Also we need to test `defer` functionality.

chronos/asyncmacro2.nim
arnetheduck:
```
proc x: void =
  return 42
```
```
testit.nim(2, 3) Error: no return type declared
```

this is what the compiler normally prints - not great but... maybe "cannot return value in function returning void"

also, the call to `error` should include `node` so that line number is included

chronos/asyncmacro2.nim
```diff
@@ -154,14 +152,19 @@ proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
...
 
     # don't do anything with forward bodies (empty)
     if procBody.kind != nnkEmpty:
       let
         # fix #13899, `defer` should not escape its original scope
         procBodyBlck = nnkBlockStmt.newTree(newEmptyNode(), procBody)
 
+        # workaround https://github.com/nim-lang/Nim/issues/22645
+        internalFutureTypeSym = genSym(nskType, "InternalFutureType")
+        castFutureSym2 = nnkCast.newTree(internalFutureTypeSym, internalFutureSym)
```
arnetheduck, line 166: expanded code comment

chronos/asyncmacro2.nim
Menduist: And lesson learned, shouldn't copy the compiler error messages :stuck_out_tongue:

cheatfate: Compiler provides `no return type declared` your code provides `no async return type declared`.
Where you found any mention of `async return type` and what the difference with `return type` is unclear.

cheatfate: A lot of compiler error messages are "cryptic" and sometimes when you generate code using macro, compiler could generate absolutely inadequate error message which could be understand only by macro code author. In this case generating error message is author's duty, because at the end error message will be read by library user, and it should not be cryptic, it should has meaning in context.

Menduist:
We found a failure with this PR, and no workaround yet:
```nim
  proc testReturner: Future[int] {.async.} =
    template returner =
      result = 5
    returner
```
I'll look a bit more for a workaround, otherwise the solution will be to retain the temporary `result` variable.

arnetheduck:
Superseded by #449

PR #429, opened 2023-08-01
Use random ports

diegomrsantos:
Use random ports in bind to local address test


arnetheduck:
Nice!

diegomrsantos:
> Nice!

but it failed on linux-amd64 (Nim version-2-0), so I'm not convinced this fixes the underlying issue.

PR #432, opened 2023-08-03
avoid annoyed hints

bung87:


arnetheduck:
fixed in a7f708bea897ab81ee57aab66628c62f12aa213a

arnetheduck:
regarding raising, we should remove `Defect` everywhere now that only 1.6+ is supported

PR #439, opened 2023-08-08
Callback duration metric

benbierens:
All right... this doesn't currently work and I need some help.

Here's the goal:
As a chronos user, I want to be able to measure callbacks durations (when using debug builds), so that I can be informed about long callbacks that will cause my app to block/not respond.

I've added a compile-flag and some code to asyncloop.nim to try and get this going, only to realize that chronos doesn't use nim-metrics. (And can't because nim-metrics uses chronos?) So I'm writing this message to ask help from you, experienced chronos developers! Do you see a use for a feature of this sort? How can this best be accomplished?

### DIFF
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -15,6 +15,12 @@ import std/[tables, strutils, heapqueue, deques]
 import stew/results
 import "."/[config, futures, osdefs, oserrno, osutils, timer]
 
+when defined(chronosEnableCallbackDurationMetric):
+  import std/monotimes
+  import pkg/metrics
+
+  declareGauge(chronosCallbackDuration, "chronos callback duration")
+
 export Port
 export futures, timer, results
 
@@ -260,7 +266,14 @@ template processCallbacks(loop: untyped) =
     if isSentinel(callable):
       break
     if not(isNil(callable.function)):
+      when defined(chronosEnableCallbackDurationMetric):
+        let startTime = getMonoTime().ticks
+
       callable.function(callable.udata)
+
+      when defined(chronosEnableCallbackDurationMetric):
+        let durationUs = (getMonoTime().ticks - startTime) div 1000
+        chronosCallbackDuration.set(durationNs)
 
 proc raiseAsDefect*(exc: ref Exception, msg: string) {.noreturn, noinline.} =
   # Reraise an exception as a Defect, where it's unexpected and can't be handled
--- a/chronos/asyncloop.nim
+++ b/chronos/asyncloop.nim
@@ -15,7 +15,7 @@ import std/[tables, strutils, heapqueue, deques]
 import stew/results
 import "."/[config, futures, osdefs, oserrno, osutils, timer]
 
-when defined(chronosEnableCallbackDurationMetric):
+when chronosEnableCallbackDurationMetric:
   import std/monotimes
   import pkg/metrics
 
@@ -266,14 +266,14 @@ template processCallbacks(loop: untyped) =
     if isSentinel(callable):
       break
     if not(isNil(callable.function)):
-      when defined(chronosEnableCallbackDurationMetric):
+      when chronosEnableCallbackDurationMetric:
         let startTime = getMonoTime().ticks
 
       callable.function(callable.udata)
 
-      when defined(chronosEnableCallbackDurationMetric):
+      when chronosEnableCallbackDurationMetric:
         let durationUs = (getMonoTime().ticks - startTime) div 1000
-        chronosCallbackDuration.set(durationNs)
+        chronosCallbackDuration.set(durationUs)
 
 proc raiseAsDefect*(exc: ref Exception, msg: string) {.noreturn, noinline.} =
   # Reraise an exception as a Defect, where it's unexpected and can't be handled
--- a/chronos/config.nim
+++ b/chronos/config.nim
@@ -70,7 +70,7 @@ when (NimMajor, NimMinor) >= (1, 4):
         ""
       ## OS polling engine type which is going to be used by chronos.
 
-    chronosEnableCallbackDurationMetric {.booldefine.} = false
+    chronosEnableCallbackDurationMetric* {.booldefine.} = defined(chronosEnableCallbackDurationMetric)
       ## At the cost of some performance, produce a 'chronosCallbackDuration' metric.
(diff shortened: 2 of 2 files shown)


elcritch:
> the way to do this would be to introduce a pre/post callback which gets enabled by a `config` parameter a bit like future tracking - in that pre/post, you could take whatever measurements you want (including capturing a call stack of the request and for example saving all callbacks that take >N ms or whatever other policy you might want for the specific problem you're looking at. You'd also need to more extensively track the origin of the callback, ie was it a user-added callback, a timer, some internal chronos mechanism or a future being woken up to continue its work.

I was thinking about that. It seems like it'd be possible to allow generic pre/post calls. It's not quite clear if it'd be more useful to track all callbacks or Future's or closure-iterators.

I took a stab at it using a more concrete timing for each closure iterator for each async proc: https://github.com/status-im/nim-chronos/pull/442

That approach doesn't allow for getting stacktraces, or only recording long running ones.

cheatfate:
> the way to do this would be to introduce a pre/post callback which gets enabled by a `config` parameter a bit like future tracking - in that pre/post, you could take whatever measurements you want (including capturing a call stack of the request and for example saving all callbacks that take >N ms or whatever other policy you might want for the specific problem you're looking at. You'd also need to more extensively track the origin of the callback, ie was it a user-added callback, a timer, some internal chronos mechanism or a future being woken up to continue its work.

If there no re-entrant calls to `poll` callstack will be always the same and looks like this:
```
NimMain
...
poll()
callback()
```
And `...` is some starting code your binary executes before it starts executing `poll()` in loop.

Calculating callbacks duration is pretty uninformative and intrusive for chronos scheduler. If you {.async.} procedure is slow - it is slow, if you need to get time of execution of this procedure - establish time measurement around this procedure. Metrics you trying to meter will help you measure parts of your procedures. For example:

```
proc someFoo() {.async.} =
  # Callback 1 measurement start
  ...
  # Callback 1 measurement finish
  await someOtherFooCall()
  # Callback 2 measurement start
  ...
  # Callback 2 measurement finish
  await someNewFooCall()
  # Callback 3 measurement start
  ...
  # Callback 3 measurement finish
```
So you got 3 measurement of code between calls to `await`. Is it what you need? I dont see any reasons why `chronos` should have facility to make time measurements of this places, and this become even more confusing if you have something like that:
```
proc someFoo() {.async.} =
  for item in iter():
    # Callback time measurement ends
    await someNewFooCall()
    # Callback time measurement starts
```

arnetheduck:
> I took a stab at it using a more concrete timing for each closure iterator for each async proc: https://github.com/status-im/nim-chronos/pull/442


I think this is being approached from the wrong end: ie I'd recommend _first_ building a useful end-to-end solution for something concrete - ie something that provides value - if, in the end, it so happens that there exists a feature of sufficient genericity that it indeed would be easier to create a generic / shared solution in chronos, we can look at it from the perspective of a specific useful thing - like this, we're just throwing mud at a wall to see if something will stick and what shape it'll have.

elcritch:
> So you got 3 measurement of code between calls to `await`. Is it what you need? I dont see any reasons why `chronos` should have facility to make time measurements of this places, and this become even more confusing if you have something like that:

That's a useful question, and one I'm trying to understand.

Measuring the non-async sections can be useful since these sections are often the parts that disrupt overall tail latencies in a project. It's often useful to tracking down incorrect usage of blocking calls (disk IO, compute, blocking network calls, unbroken for-loops, etc). 

Meanwhile, the time of the async portions can be somewhat gauged by proxy using total request time.

If I understand it, the callback approach described above wouldn't provide the granularity to measure the actual time it takes to execute the async proc time, only the total duration.

elcritch:
> I think this is being approached from the wrong end: ie I'd recommend _first_ building a useful end-to-end solution for something concrete - ie something that provides value - 

I disagree; this is a useful concrete solution to measure timings which would otherwise not be readily feasible to us. Also, the feedback from @cheatfate was helpful.

> if, in the end, it so happens that there exists a feature of sufficient genericity that it indeed would be easier to create a generic / shared solution in chronos, we can look at it from the perspective of a specific useful thing - like this, we're just throwing mud at a wall to see if something will stick and what shape it'll have.

I'm onboard with creating a more generic solution if possible, but wanted to spitball something concrete and see how well it worked. These sort of timings could provide our project value, and possibly others, since non-async portions can easily cause issues.

> Notably, the above can trivially be developed in a branch of chronos and showcased using that branch.

Isn't that what a draft PR is for? ;) Still that was my first approach, but I don't have access to create a branch and things get lost on a fork. I am also new to Chronos as a project and unsure the best way to get useful feedback. A draft PR gives a chance for a broader audience to chime in.

elcritch:
I took my PR from yesterday and used it to make a more generic solution based callbacks on `FutureBase`. It's not ready for upstreaming so I left it on my fork, but if you want to see it: https://github.com/elcritch/nim-chronos/pull/2

I'm not 100% sure if it's where you were thinking of adding the callbacks @arnetheduck. Future's made the most sense as it could cover both iterators and callbacks. It'd need more testing to see how usable it'd. 

Note, I also needed a third callback `pause` to enable collecting execution time and not just the total elapsed time from once the Future starts. As @cheatfate mentioned that wouldn't include "child futures". That might be possible by combining it with the FutureTracking feature. All the data should in theory be able to be collected downstream.

arnetheduck:
> thinking of adding the callbacks

@elcritch the end-to-end example would be easier to evaluate, ie it's easy to add a library feature that ends up not being very practical for different reasons - lmk if you get any further with your approach, ie what kind of inefficiencies it can catch and we'll take it from there.

since the rest of this PR is not moving, we'll close it for now.

elcritch:
> > thinking of adding the callbacks
> 
> @elcritch the end-to-end example would be easier to evaluate, ie it's easy to add a library feature that ends up not being very practical for different reasons - lmk if you get any further with your approach, ie what kind of inefficiencies it can catch and we'll take it from there.

I recall this approach being useful for profiling some expensive operations that'd been causing issues for months at the time. That team had done some more work combining it with Metrics library that worked well IIRC. 

However, I'm not using Chronos significantly currently, but if I were using it in production this profiling would be one of the first tools I'd reach for to understand P95, P99, etc latency. If I use it again for something I'll make an end-to-end example.

PR #440, opened 2023-08-08
Unroll `defer`s and remove `break`s.

cheatfate:


arnetheduck:
generally, I think #418 is a better general-purpose fix. The problem with `defer` and `finally` is that we don't run the iterator to completion so not all code is run - this is the same problem as below, without a closure iterator transformation:

```nim
iterator x(): int {.closure.} =
  yield 42
  echo "hello"

proc works() =
   var z = x
   for a in z():
     echo a

proc broken() =
  var z = x

  for a in z():
    echo a
    break

works()
broken()

```

cheatfate:
So should i wait for #418 to become ready or we are looking for leaks right now?

PR #442, opened 2023-08-10
Per callback duration stats

elcritch:
Variation on https://github.com/status-im/nim-chronos/pull/439 using per-closure-iterator timings.


elcritch:
Closed in favor of https://github.com/elcritch/nim-chronos/pull/2 which makes this PR doable in a generic fashion using instrumentation hooks (callbacks).

PR #446, opened 2023-09-04
HttpAddress errors should be not only critical.

cheatfate:


chronos/apps/http/httpcommon.nim
```diff
@@ -82,6 +82,17 @@ type
   HttpState* {.pure.} = enum
     Alive, Closing, Closed
 
+  HttpAddressErrorType* {.pure.} = enum
```
arnetheduck, line 85: or list specific error kinds we interpret then have a set of critical ones, ie `Critical = {ParseError, XxxErro}, Recoverable = {LookupError, ...}`

Issue #443, opened 2023-08-13
httpclient: The content of HostHeader is incorrect when redirecting to a different host


Pylgos:
When redirecting with `httpclient.redirect`, the new request inherits the original request's headers, so if the hostname of the redirection destination is different from the original hostname, the request will be invalid.

```nim
import chronos
import chronos/apps/http/[httptable, httpcommon, httpclient]
import std/uri

proc main {.async.} =
  let req = HttpClientRequestRef.get(
    HttpSessionRef.new(), "https://api.github.com/repos/status-im/nim-chronos/zipball/master").tryGet()
  let resp = await req.send()
  doAssert resp.status == 302
  
  # BUG HERE
  let redirect = req.redirect(resp.getNewLocation().tryGet()).tryGet()
  echo redirect.address.hostname # -> codeload.github.com
  echo redirect.headers.getString(HostHeader) # -> api.github.com

  await resp.closeWait()
  await req.closeWait()

  # let (status, data) = await redirect.fetch()
  # await redirect.closeWait()

  # echo status # -> 400 bad request

waitFor main()
```

cheatfate:
Fixed in 00614476c68f0553432b4bb505e24d6ad5586ae4

PR #447, opened 2023-09-07
Address issue #443.

cheatfate:


chronos/apps/http/httpclient.nim
```diff
@@ -1409,22 +1409,6 @@ proc consumeBody*(response: HttpClientResponseRef): Future[int] {.async.} =
     response.setError(error)
     raise error
 
-proc redirect*(request: HttpClientRequestRef,
```
arnetheduck, line 1412: why remove this overload?

PR #449, opened 2023-09-26
Complete futures in closure finally (fix #415)

### LINKED ISSUE #415: return statement in try/finally block breaks executing multiple await statements in finally block

Menduist:
Alternative to https://github.com/status-im/nim-chronos/pull/418
Fixes #415 

This moves the error handling to the closure (instead of the `futureContinue`). In the finally, we complete with `result` in case of success. Example:
```nim
proc test415: Future[int] {.async.} =
  return 5

# becomes (stripped the `when`s for clarity):
proc test415(): Future[int] {.stackTrace: off, gcsafe, raises: [].} =
  iterator test415_788529156(chronosInternalRetFuture: FutureBase): FutureBase {.
      closure, gcsafe, raises: [].} =
    var result {.used.}: int
    try:
       result = 5
       return nil
    except CancelledError:
      cancelAndSchedule(cast[Future[int]](chronosInternalRetFuture))
    except CatchableError as exc:
      fail(cast[Future[int]](chronosInternalRetFuture), exc)
    finally:
      if not finished(cast[Future[int]](chronosInternalRetFuture)):
        complete(cast[Future[int]](chronosInternalRetFuture), result)

  let resultFuture = newFuture[int]("test415")
  resultFuture.internalClosure = test415_788529156
  futureContinue(resultFuture)
  return resultFuture
```


tests/testmacro.nim
```diff
@@ -136,6 +136,93 @@ suite "Macro transformations test suite":
...
+      return
+        if 1 == 1:
+          return 42
+        else:
+          33
+
+    check waitFor(nr()) == 42
+
+  test "Issue #415 (run closure to completion on return)":
+    var x = 0
```
arnetheduck, line 150: let's use a fresh variable for each test, this ends up being a global which is annoying for gcsafety down the line

arnetheduck: nevermind, missed that it's in a `test`

Menduist: Yeah, could make it a suite instead if you want (and change `x` into a var)

arnetheduck: that's probably more clean, yes

arnetheduck (approved):

arnetheduck:
lgtm but needs a round of testing with nimbus-eth2

Menduist:
Defects are now handled by failing the future with a new exception, similarly to other exceptions
Could also leave the future running, but since the chronosStrictException will become default, not sure it's worth the trouble

arnetheduck:
> Could also leave the future running, but since the chronosStrictException will become default, not sure it's worth the trouble

let's leave the future running and propagate to `poll` like it was before

Menduist:
This is causing some regression in the libp2p CI, investigating..

Menduist:
Fixed the regression, I also took the opportunity to "improve" #401, since we now need a template to set the result anyway

The issue was:
```nim
    proc returnResult: Future[int] {.async.} =
      var result: int
      result = 12
      return result
```
was not working properly anymore, since `return result` was transformed to `result = result; return nil`
which was not setting the correct `result`

The fix is:
```nim
    var result {.used.}: int
    template setResult(code: untyped) {.used.} =
      result = code
    block:
       var result: int
       result = 12
       setResult(12)
```
since result is bound in the template, this will set the correct result.

And I took the opportunity to handle the implicit returns directly inside setResult:
```nim
    template setResult(code: untyped) {.used.} =
      when typeof(code) is void:
        code
      else:
        result = code
```

LMKWYT

Menduist:
libp2p & nimbus tests now pass with this branch

PR #453, opened 2023-10-17
Introduce chronos/internals, move some code

arnetheduck:
This PR breaks the include dependencies between `asyncfutures2` and `asyncmacros2` by moving the dispatcher and some other code to a new module.

This step makes it easier to implement `asyncraises` support for future utilities like `allFutures` etc avoiding the need to play tricks with include order etc.

Future PR:s may further articulate the difference between "internal" stuff subject to API breakage and regular public API intended for end users (rather than advanced integrators).


arnetheduck:
discussed with @cheatfate, moving on to the next PR

PR #454, opened 2023-10-18
introduce `asyncraises` to core future utilities

arnetheduck:
Similar to the introduction of `raises` into a codebase, `asyncraises` needs to be introduced gradually across all functionality before deriving benefit.

This is a first introduction along with utilities to manage raises lists and transform them at compile time.

Several scenarios ensue:

* for trivial cases, adding `asyncraises` is enough and the framework deduces the rest
* some functions "add" new asyncraises (similar to what `raise` does in "normal" code) - for example `wait` may raise all exceptions of the future passed to it and additionally a few of its own - this requires extending the raises list
* som functions "remove" raises (similar to what `try/except` does) such as `nocancel` with blocks cancellations and therefore reduce the raising set

Both of the above cases are currently handled by a macro, but depending on the situation lead to code organisation issues around return types and pragma limitations - in particular, to keep `asyncraises` backwards-compatibility, some code needs to exist in two versions which somewhat complicates the implementation.

* add `asyncraises` versions for several `asyncfutures` utilities
* when assigning exceptions to a `Future` via `fail`, check at compile time if possible and at runtime if not that the exception matches constraints
*

### DIFF
--- a/chronos/internal/asyncfutures.nim
+++ b/chronos/internal/asyncfutures.nim
@@ -13,8 +13,10 @@
 import std/[sequtils, macros]
 import stew/base10
 
-import ./asyncengine
+import ./[asyncengine, raisesfutures]
 import ../[config, futures]
+
+export raisesfutures.InternalRaisesFuture
 
 when chronosStackTrace:
   import std/strutils
@@ -40,12 +42,6 @@ func `[]`*(loc: array[LocationKind, ptr SrcLoc], v: int): ptr SrcLoc {.
   else: raiseAssert("Unknown source location " & $v)
 
 type
-  InternalRaisesFuture*[T, E] = ref object of Future[T]
-    ## Future with a tuple of possible exception types
-    ## eg InternalRaisesFuture[void, (ValueError, OSError)]
-    ## Will be injected by `asyncraises`, should generally
-    ## not be used manually
-
   FutureStr*[T] = ref object of Future[T]
     ## Future to hold GC strings
     gcholder*: string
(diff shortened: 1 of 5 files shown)


chronos/internal/asyncfutures.nim
```diff
@@ -52,6 +54,78 @@ type
     ## Future to hold GC seqs
     gcholder*: seq[B]
 
+  SomeFuture = Future|InternalRaisesFuture
+
+macro prepend(tup: typedesc[tuple], typ: varargs[typed]): typedesc =
```
cheatfate, line 59: Is it possible to move this code to `asyncmacro.nim`? I really do not want to see macro code here.

chronos/internal/asyncfutures.nim
```diff
@@ -1264,7 +1297,7 @@ proc sleepAsync*(duration: Duration): Future[void] =
   return retFuture
 
 proc sleepAsync*(ms: int): Future[void] {.
-     inline, deprecated: "Use sleepAsync(Duration)".} =
+     inline, deprecated: "Use sleepAsync(Duration)", asyncraises: [CancelledError].} =
   result = sleepAsync(ms.milliseconds())
 
 proc stepsAsync*(number: int): Future[void] =
```
cheatfate, line 1303: Why you missed this one? It still could raise `CancelledError` too, actually every async procedure could raise it, either raw or transformed one could raise `CancelledError`.

chronos/internal/asyncfutures.nim
```diff
@@ -1515,3 +1553,53 @@ when defined(windows):
...
+
+template fail*[T, E](
+    future: InternalRaisesFuture[T, E], error: ref CatchableError,
+    warn: static bool = true) =
+  checkFailureType(future, error, warn)
+  fail(future, error, getSrcLocation())
+
+proc waitFor*[T, E](fut: InternalRaisesFuture[T, E]): T = # {.raises: [E]}
+  ## **Blocks** the current thread until the specified future completes.
+  ## There's no way to tell if poll or read raised the exception
```
cheatfate, line 1567: This comment is ugly one, `poll()` do not raise exceptions anymore.

cheatfate:
I thinik you missed `or` here and it become as complex as `noCancel`. Because in case of `or` it should create union of exceptions of `Future1` and `Future2`. Personally i do not like this primitive because it has weird design, for some reason it returns only errors, but not values. From my point of view it should not return errors too, but i think there some code around which depends on this ugly implementation.

chronos/internal/asyncfutures.nim
arnetheduck: I wasn't sure if it can raise but I guess it can.

Raw or not, if an `async` function always returns a completed Future, it doesn't ever raise `CancelledError`.

This effectively means that only functions (raw or not) that create a future manually and have a non-finishing code path can ever raise `CancelledError` (plus those that add `CancelledError` manually for future-proofing the API).

arnetheduck: Also, notably, any function that doesn't have an `asyncraises` is implicitly treated as `asyncraises: [CatchableError]`

arnetheduck:
> I thinik you missed or here and it become as complex as

Well, I intentionally left it out for now because it's not used that much - because of the default `CatchableError` raises, it's not wrong, just sub-optimal. That said, it is indeed an odd implementation so I don't really want to touch it in this PR - maybe it should be deprecated and maybe not, but I don't think it's worth updating to asyncraises right now.

>  start removing 

Removing is not a problem generally but if another version is added with the same name, it becomes difficult to maintain compatibility or at least compile-time discoverability of what needs to be updated.

chronos/internal/asyncfutures.nim
arnetheduck: > Is it possible to move this code to asyncmacro.nim? 

messy - these should not become public, at least not with these names - they're not meant to be part of the API (yet) - later, maybe it makes sense to expose exception-list-manipulation-helpers but that would be a new module I think - they're here for convenience mainly.

chronos/internal/asyncfutures.nim
```diff
@@ -1110,7 +1140,8 @@ proc allFinished*[T](futs: varargs[Future[T]]): Future[seq[Future[T]]] =
 
   return retFuture
 
-proc one*[T](futs: varargs[Future[T]]): Future[Future[T]] =
+proc one*[F: SomeFuture](futs: varargs[F]): Future[F] {.
+    asyncraises: [ValueError, CancelledError].} =
```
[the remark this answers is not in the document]
arnetheduck: we cam deprecate maybe it but I think in a separate PR

chronos/internal/asyncfutures.nim
arnetheduck: 174644ef149ba6a40d05975cc4821bdfbb328735

chronos/internal/asyncfutures.nim
arnetheduck: moved to its own module - 03dd6ed104755a67fda2c5c733e84dd139ab4ec1

cheatfate (approved):

PR #456, opened 2023-10-25
IPv4/IPv6 dualstack

cheatfate:


chronos/handles.nim
```diff
@@ -78,10 +119,48 @@ proc getSocketError*(socket: AsyncFD, err: var int): bool =
...
+    true
+  else:
+    let fd = osdefs.socket(toInt(domain), toInt(SockType.SOCK_STREAM),
+                           toInt(Protocol.IPPROTO_TCP))
+    if fd == -1:
+      return if osLastError() == osdefs.EAFNOSUPPORT: false else: true
+    discard closeFd(fd)
+    true
+
+proc isIPv4Available*(): bool =
```
arnetheduck, line 141: why not `isAvailable(domain: Domain)`?

chronos/transports/common.nim
```diff
@@ -720,3 +726,57 @@ proc raiseTransportError*(ecode: OSErrorCode) {.
...
+      ok(AddressFamily.IPv6)
+    elif int(saddr.ss_family) == toInt(Domain.AF_UNIX):
+      ok(AddressFamily.Unix)
+    else:
+      err(EAFNOSUPPORT)
+
+proc setDualstack*(socket: AsyncFD, family: AddressFamily,
+                   flag: DualStackType): Result[void, OSErrorCode] =
+  if family == AddressFamily.IPv6:
+    let flags = getThreadDispatcher().networkFlags.get(
```
arnetheduck, line 765: instead of going via the flags here, why not simply ignore specific errors from `setDualstack`?

It's possible to enable/disable ipv6 dynamically (on linux yes - on windows, not, on mac not sure) which means that `networkFlags` might be out of date at the time of this call - I think it would be less error-prone if this wasn't cached.

chronos/transports/common.nim
```diff
@@ -29,7 +29,11 @@ type
   ServerFlags* = enum
     ## Server's flags
     ReuseAddr, ReusePort, TcpNoDelay, NoAutoRead, GCUserData, FirstPipe,
-    NoPipeFlash, Broadcast
+    NoPipeFlash, Broadcast, DualStackAuto, DualStackEnabled, DualStackDisabled,
+    DualStackDefault
```
arnetheduck, line 33:
```suggestion
    DualStackOSDefault
```

chronos/transports/common.nim
```diff
@@ -720,3 +726,57 @@ proc raiseTransportError*(ecode: OSErrorCode) {.
...
+    else:
+      err(EAFNOSUPPORT)
+
+proc setDualstack*(socket: AsyncFD, family: AddressFamily,
+                   flag: DualStackType): Result[void, OSErrorCode] =
+  if family == AddressFamily.IPv6:
+    let flags = getThreadDispatcher().networkFlags.get(
+                  {NetFlag.IPv4Enabled, NetFlag.IPv6Enabled})
+    case flag
+    of DualStackType.Auto, DualStackType.Enabled:
```
arnetheduck, line 768: Regarding ignoring errors, I think a useful difference between `Auto` and `Enabled` would be that `Auto` ignores errors while `Enabled` gives an error if we cannot enable dual-stack - for example if ipv4 is disabled on a host, `Auto` would result in no error while `Enabled` would result in an error.

chronos/transports/stream.nim
```diff
@@ -639,6 +639,7 @@ when defined(windows):
                 child: StreamTransport = nil,
                 localAddress = TransportAddress(),
                 flags: set[SocketFlags] = {},
+                dualstack = DualStackType.Auto
```
arnetheduck, line 642: why do we need dualstack on connect?

cheatfate: We need it, because otherwise IPv4 mapped address will not work when used in `connect()`.

arnetheduck (approved):

PR #457, opened 2023-10-26
per-function `Exception` handling

arnetheduck:
This PR replaces the global strict exception mode with an option to handle `Exception` per function while at the same time enabling strict exception checking globally by default as has been planned for v4.

`handleException` mode raises `AsyncExceptionError` to distinguish it from `ValueError` which may originate from user code.

* remove obsolete 1.2 config options


arnetheduck:
draft until #455 is merged

cheatfate (approved):

PR #459, opened 2023-10-30
Fix unreachable code places.

cheatfate:


chronos/streams/tlsstream.nim
```diff
@@ -151,102 +151,96 @@ proc new*(T: typedesc[TrustAnchorStore],
...
-    writer.state = AsyncStreamState.Error
-    writer.error = exc
-    return TLSResult.Error
-  except CancelledError:
-    if writer.state == AsyncStreamState.Running:
-      writer.state = AsyncStreamState.Stopped
-    return TLSResult.Stopped
-
-  return TLSResult.Error
+  return
```
arnetheduck, line 154: implicit return should work nowadays..

PR #465, opened 2023-11-09
Attempt to fix CI crash at Windows.

cheatfate:
Remove all cast[string] and cast[seq[byte]] from the codebase.


tests/testhelpers.nim
```diff
@@ -0,0 +1,25 @@
+#                Chronos Test Suite
+#            (c) Copyright 2019-Present
+#         Status Research & Development GmbH
+#
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+
+func toBytes*(src: openArray[char]): seq[byte] =
```
arnetheduck, line 9: isn't there `bytesToString` etc already?

arnetheduck (approved):

chronos/internal/asyncengine.nim
```diff
@@ -1125,14 +1125,18 @@ proc addTimer*(at: uint64, cb: CallbackFunc, udata: pointer = nil) {.
...
+    index =
+      block:
+        var res = -1
+        for i in 0 ..< len(loop.timers):
+          if (loop.timers[i].finishAt == at) and
+             (loop.timers[i].function.function == cb) and
+             (loop.timers[i].function.udata == udata):
+            res = i
+            break
+        res
```
arnetheduck, line 1139: won't `del` reorder the times below? or maybe that doesn't matter?

tests/testhelpers.nim
cheatfate: yep but it is declared in `http` module.

chronos/internal/asyncengine.nim
cheatfate: I think it was wrong initially because when it converts `HeapQueue` to `seq` - `del` for `seq` will be used, but now we using `del` for `HeapQueue` which is defined here https://github.com/nim-lang/Nim/blob/devel/lib/pure/collections/heapqueue.nim#L181-L193. So i think its safe.

tests/testhelpers.nim
arnetheduck: we have stew/byteutils too ..

PR #471, opened 2023-11-15
Strict Future value / error access and related fixes

arnetheduck:
Strict future access is now always enabled, making it a panic to access value/error when the Future is still pending - `read` and `readError` raise catchable exceptions instead which may be preferable to use.

* remove `chronosStrictFutureAccess` enabling its functionality by default
* disallow setting a cancellation callback on a non-pending future
* introduce `FuturePendingError` and `FutureCompletedError` when `read`:ing a future of uncertain state
* introduce `pollFor` which finishes a `Future` without reading it, mirroring `awaitne`
* fix `waitFor` / `read` to return `lent` values
* simplify code generation for `void`-returning async procs
* document `Raising` type helper


arnetheduck:
This PR is marked draft to better investigate the consequences of `Defect` on `error` - in particular, `if fut.error == nil` is a somewhat established idiom to check if the future failed even though accessors like `failed` exist for this purpose.

arnetheduck:
#474 contains the changes unrelated to the new Defect

PR #472, opened 2023-11-16
Asyncstreams asyncraises.

cheatfate:


chronos/streams/asyncstream.nim
```diff
@@ -24,15 +24,15 @@ const
     ## AsyncStreamWriter leaks tracker name
 
 type
-  AsyncStreamError* = object of CatchableError
+  AsyncStreamError* = object of TransportError
```
arnetheduck, line 27: I'm not sure it's a good idea to make this a transport error, ie this is a different abstraction level and the inheritance creates an abstraction leak.. 

better put the transport error in the `parent` field and translate, so that callers only ever have to understand `AsyncStreamError` and its derivatives for normal error handling

chronos/streams/asyncstream.nim
```diff
@@ -24,15 +24,15 @@ const
...
   AsyncStreamIncorrectDefect* = object of Defect
   AsyncStreamIncompleteError* = object of AsyncStreamError
   AsyncStreamLimitError* = object of AsyncStreamError
   AsyncStreamUseClosedError* = object of AsyncStreamError
   AsyncStreamReadError* = object of AsyncStreamError
-    par*: ref CatchableError
+    par*: ref TransportError
```
arnetheduck, line 33: hm, `Exception` already has a parent field

chronos/asyncsync.nim
```diff
@@ -267,16 +267,32 @@ proc addFirstNoWait*[T](aq: AsyncQueue[T], item: T) {.
...
-    raises: [AsyncQueueFullError].}=
+     raises: [AsyncQueueFullError].} =
   ## Put an item ``item`` at the end of the queue ``aq`` immediately.
   ##
   ## If queue ``aq`` is full, then ``AsyncQueueFullError`` exception raised.
   if aq.full():
     raise newException(AsyncQueueFullError, "AsyncQueue is full!")
   aq.addLastImpl(item)
 
+proc addFirstNoWaitSafe*[T](aq: AsyncQueue[T], item: T) {.
```
arnetheduck, line 278: how about a `distinct UnboundedQueue` for this case?

chronos/asyncsync.nim
```diff
@@ -285,19 +301,37 @@ proc popFirstNoWait*[T](aq: AsyncQueue[T]): T {.
...
+     raises: [AsyncQueueEmptyError].} =
   ## Get an item from the end of the queue ``aq`` immediately.
   ##
   ## If queue ``aq`` is empty, then ``AsyncQueueEmptyError`` exception raised.
   if aq.empty():
     raise newException(AsyncQueueEmptyError, "AsyncQueue is empty!")
   aq.popLastImpl()
 
-proc addFirst*[T](aq: AsyncQueue[T], item: T) {.async: (raises: [CancelledError]).} =
+proc popFirstNoWaitSafe*[T](aq: AsyncQueue[T]): T {.
```
arnetheduck, line 312: looking at the usages, we could perhaps create a pop iterator instead? something like:

```iterator popItems(): T =
    yield item
    del item
```

chronos/streams/asyncstream.nim
cheatfate: Every `AsyncStream` needs source of data, there is no empty sourced AsyncStreams right now. So at the end every AsyncStream will pop up TransportError if transport fails to deliver data.

chronos/asyncsync.nim
cheatfate: I'm not sure i understand, i need AsyncQueue in that place, i do not need any other queues.

chronos/asyncsync.nim
cheatfate: What about adding items until its not full?

chronos/streams/asyncstream.nim
arnetheduck: sure - but the inheritance here seems unnecesary - ie it doesn't satisfy the usual "is a" test - async stream is not a transport, so why should its errors be transport errors?

chronos/asyncsync.nim
arnetheduck: the goal this comment was to investigate if we can have an api that admit the risk of defects "naturally" - this is why I suggested a separate `UnboundedAsyncQueue` - its `add` would never raise.

I don't think that a program will want to switch between bounded or unbounded "dynamically" - I can understand that a program may want to control the size of a bounded queue, but it seems like we could separate the bounded and unbounded case using a separate type for unbounded - this unbounded queue would not raise exceptions on `add` - together with the `popItems` iterator (or some better name), we could have a "defect-free" API for unbounded queues that also doesn't have unwanted `raises` effects.

"safe" is a difficult term in general - raising panics is not a "Safe" thing to do, so it's hard to tell what is safe and what is not from looking at the word - it is not well defined in nim and therefore a source of confusion - better avoid defects altogether.

chronos/asyncsync.nim
arnetheduck: the idea is to not have these `Safe` suffixes but instead use a separate type for the unbounded use case - "safe" here is easy to interpret as "never raises defect" when in fact it is the opposite: the caller can only call them in special cases where it is safe to do so.

cheatfate: The problem is that AsyncQueue i'm using is not unbounded. It is bounded and has limits, but in the way i'm using it - it is safe.

arnetheduck: should we call it `addFirstNotFull` then?

arnetheduck (approved):

PR #473, opened 2023-11-17
fix `cancelAndWait` template return type

arnetheduck:


chronos/internal/asyncfutures.nim
```diff
@@ -1532,6 +1523,15 @@ proc read*[E](future: InternalRaisesFuture[void, E]) = # {.raises: [E, Cancelled
     # TODO: Make a custom exception type for this?
     raise newException(ValueError, "Future still in progress.")
 
+proc waitFor*[T, E](fut: InternalRaisesFuture[T, E]): T = # {.raises: [E]}
```
cheatfate, line 1526: This is incorrect version of `waitFor`.
It should look like this:
```nim
proc waitFor*[T](fut: Future[T]): T {.raises: [CatchableError].} =
  ## **Blocks** the current thread until the specified future finishes and
  ## reads it, potentially raising an exception if the future failed or was
  ## cancelled.
  var finished = false
  # Ensure that callbacks currently scheduled on the future run before returning
  proc continuation(udata: pointer) {.gcsafe.} = finished = true
  if not(fut.finished()):
    fut.addCallback(continuation)
    while not(finished):
      poll()
  fut.read()
```

arnetheduck: this is fixed in https://github.com/status-im/nim-chronos/pull/474

arnetheduck: (this code move is here to ensure that all versions of `read` come before `waitFor` which uses one of the overloads - this usually helps nim pick the right version)

arnetheduck:
merged as part of #474

PR #479, opened 2023-11-21
Asyncraises for HTTP server, part 2

cheatfate:
Fix processing callback missing asyncraises.
Fix tests.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -65,7 +65,7 @@ type
 
   HttpProcessCallback* =
     proc(req: RequestFence): Future[HttpResponseRef] {.
-      gcsafe, raises: [].}
+      async: (raises: [CancelledError, HttpResponseError]), gcsafe.}
 
   HttpConnectionCallback* =
     proc(server: HttpServerRef,
@@ -448,7 +448,7 @@ proc getBodyReader*(request: HttpRequestRef): HttpResult[HttpBodyReader] =
     err("Request do not have body available")
 
 proc handleExpect*(request: HttpRequestRef) {.
-     async: (raises: [CancelledError, HttpError]).} =
+     async: (raises: [CancelledError, HttpCriticalError]).} =
   ## Handle expectation for ``Expect`` header.
   ## https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect
   if HttpServerFlags.NoExpectHandler notin request.connection.server.flags:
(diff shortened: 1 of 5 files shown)


chronos/apps/http/httpserver.nim
```diff
@@ -905,10 +905,11 @@ proc getResponseFence*(connection: HttpConnectionRef,
     let address = connection.getRemoteAddress()
     ResponseFence.err(HttpProcessError.init(
       HttpServerError.RecoverableError, exc, address, exc.code))
-  except CatchableError as exc:
-    let address = connection.getRemoteAddress()
-    ResponseFence.err(HttpProcessError.init(
-      HttpServerError.CatchableError, exc, address, Http503))
+  except HttpResponseError as exc:
```
arnetheduck, line 908: as an idea, if instead we add `critical: bool` to httpresponseerror, here we can use `
```nim
  case exc.critical
  of true: ..
  of false: ..
```

derived type cannot express a finite set of options, hence we get these ugly "extra" exception handlers.

critical or not can also be an enum - if we have:
```
  enum HttpErrorKind =
     invalidParameter
     hostNotOnline
```
we can also make `const CriticalError = {HttpErrorKind.invalidParameter}` and users can write:
```
case exc.kind
of CriticalError: ...
of RecoverableError: ...
```

The downside of an enum is that if we add a new enum value, it breaks all `case` statements - as long as users use the categories / sets (`Critical` and `Recoverable`), this isn't a problem (because we can add to the categories), so maybe we can simply document that new enum values may be added in the future.

I think the field makes it easier to write good error handling code while at the same time it maintains more simple raises lists (assuming of course we're confident that we can categorise all response errors into critical or not in a meaningful way)

arnetheduck: hm, looking at the code, it looks like this PR already places `HttpCode` in `HttpResponseError` so all that's needed really is the critical / recoverable set .. we can even retain `HttpCriticalError` and `HttpRecoverableError` for backwards compatibility and document that they are equivalent to the `code` field - in new code, we can always catch only `HttpResponseError` so it's clean..

arnetheduck (approved):
per discussion, the comments are not blockers but perhaps worth taking into account either in this or future PR:s

chronos/apps/http/httpserver.nim
```diff
@@ -65,7 +65,7 @@ type
 
   HttpProcessCallback* =
     proc(req: RequestFence): Future[HttpResponseRef] {.
-      gcsafe, raises: [].}
+      async: (raises: [CancelledError, HttpResponseError]), gcsafe.}
```
arnetheduck, line 68: this change breaks nimbus - to restore backwards compatibility, a `UnsafeHttpProcessCallback` with the old definition needs to be added and overloads need to be added in the places that use it .. see https://github.com/status-im/nim-chronos/blob/28a100b1350d52668ce7b9c9ae8159fdc0acace2/chronos/transports/stream.nim#L124

PR #482, opened 2023-12-05
Asyncraises HTTP layer V3

cheatfate:


chronos/apps/http/httpserver.nim
```diff
@@ -61,15 +59,15 @@ type
...
   HttpResponseState* {.pure.} = enum
-    Empty, Prepared, Sending, Finished, Failed, Cancelled, Default
+    Empty, Prepared, Sending, Finished, Failed, Cancelled, ErrorCode, Default
 
-  # TODO Evaluate naming of raises-annotated callbacks
-  HttpProcessCallback2* =
+  HttpProcessCallback* =
     proc(req: RequestFence): Future[HttpResponseRef] {.
-      async: (raises: [CancelledError, HttpResponseError]).}
+      async: (raises: [CancelledError]), gcsafe.}
```
arnetheduck, line 66: `gcsafe` is already implied by `async` - it should not be needed here..

also, in general, I think it's a good idea to introduce a new type and a compatibility layer for this type since it's used in the wild - it seemed quite easy to do via a `2` type.

arnetheduck: The problem with using the old type name is that it's used in assignments where the added exception information makes things break.

cheatfate: So old type will be named as it is, and new type will be named as `2`?

PR #483, opened 2023-12-09
HTTP server middleware implementation.

cheatfate:
Middleware support enables integrating custom number of custom applications into single HTTP server, it also allows to implement more comprehensive and more flexible CORS support, HTTP request rate limiting and HTTP response caching.

Example of request filtering by URI presented in test.

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -11,11 +11,14 @@
 
 import std/[tables, uri, strutils]
 import stew/[base10], httputils, results
-import ../../asyncloop, ../../asyncsync
+import ../../[asyncloop, asyncsync]
 import ../../streams/[asyncstream, boundstream, chunkstream]
 import "."/[httptable, httpcommon, multipart]
+from ../../transports/common import TransportAddress, ServerFlags, `$`, `==`
+
 export asyncloop, asyncsync, httptable, httpcommon, httputils, multipart,
        asyncstream, boundstream, chunkstream, uri, tables, results
+export TransportAddress, ServerFlags, `$`, `==`
 
 type
   HttpServerFlags* {.pure.} = enum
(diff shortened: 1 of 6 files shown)


tests/testhttpserver.nim
```diff
@@ -1490,5 +1527,96 @@ suite "HTTP server testing suite":
     await server.stop()
     await server.closeWait()
 
+  asyncTest "HTTP middleware request filtering test":
+    proc init(t: typedesc[FirstMiddlewareRef],
+              data: int): HttpServerMiddlewareRef =
+      proc shandler(middleware: HttpServerMiddlewareRef, reqfence: RequestFence,
+                    handler: HttpProcessCallback2): Future[HttpResponseRef] {.
```
arnetheduck, line 1534:
```suggestion
                    nextHandler: HttpProcessCallback2): Future[HttpResponseRef] {.
```

or some other descriptive name making it clear is the responsibilty of each handler?

also, this  would ideally be described, ie the mechanism of establishing a handling priority and when handlers should call the "next" handler and when they should give a default response...

chronos/apps/http/httpserver.nim
```diff
@@ -188,20 +204,54 @@ proc createConnection(server: HttpServerRef,
...
+      block:
+        var res: HttpProcessCallback2
+        closureScope:
+          let
+            middleware = mws[index]
+            realHandler = currentHandler
+          res =
+            proc(request: RequestFence): Future[HttpResponseRef] {.
+              async: (raises: [CancelledError]).} =
+              await middleware.handler(middleware, request, realHandler)
```
arnetheduck, line 232:
```suggestion
            proc(request: RequestFence): Future[HttpResponseRef] {.
              async: (raises: [CancelledError]), raw: true.} =
              middleware.handler(middleware, request, realHandler)
```

arnetheduck (approved):

PR #485, opened 2023-12-23
Add custom ring buffer into chronos streams and transports.

cheatfate:


chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+    data: seq[byte]
+
+proc init*(t: typedesc[BipBuffer], size: int): BipBuffer =
+  ## Creates new Bip Buffer with size `size`.
+  BipBuffer(data: newSeq[byte](size))
+
+template len(pos: BipPos): Natural =
+  pos.finish - pos.start
+
+template zero(pos: var BipPos) =
```
arnetheduck, line 46: aka `reset`

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+  ## as pointer of type `pt` and size of type `st`.
+  ##
+  ## Note, this procedure will raise Defect if there is no free space available.
+  let (availableSpace, reserveStart) = bp.calcReserve()
+  doAssert(availableSpace > 0, "Not enough space available")
+  bp.r = BipPos.init(reserveStart, Natural(reserveStart + availableSpace))
+  (cast[pt](addr bp.data[bp.r.start]), cast[st](len(bp.r)))
+
+func getReserve*(bp: var BipBuffer,
+                 pt: typedesc, st: typedesc): tuple[data: pointer, size: cint] =
```
arnetheduck, line 113: `cint`??

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+  if len(bp.a) == 0 and len(bp.b) == 0:
+    bp.a.start = bp.r.start
+    bp.a.finish = bp.r.start + toCommit
+  elif bp.r.start == bp.a.finish:
+    bp.a.finish += toCommit
+  else:
+    bp.b.finish += toCommit
+  bp.r.zero()
+
+proc decommit*(bp: var BipBuffer, size: Natural): Natural =
```
arnetheduck, line 135: boost calls these `prepare`/`commit`/`consume` - `decommit` sounds strange - also, why do you need the number of bytes consumed? 

The only way you read data from this buffer is via `bytes`/`regions` so you always know how many bytes were available, so you should in principle never `decommit` more than that.

Also, I suspect the buffer needs modified-while-iterating protection, ie an easy bug for a consumer is to call `decommit` while iterating over `regions`.

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+  else:
+    let spaceAfterA = Natural(len(bp.data) - bp.a.finish)
+    if spaceAfterA >= bp.a.start:
+      (spaceAfterA, bp.a.finish)
+    else:
+      (bp.a.start, Natural(0))
+
+func freeSpace*(bp: BipBuffer): Natural =
+  ## Returns amount of free space in buffer `bp`.
+  var sum: Natural
```
arnetheduck, line 64: `bp.data.len - bp.usedSpace()` - following nim naming however, `proc capacity(): int = bp.data.len` - `usedSpace` == `len`, then this function can be removed (generic code normally expects `buf.capacity - buf.len` to be valid for a buffer)

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+    sum += bp.a.start
+  sum += Natural(len(bp.data)) - bp.a.finish
+  sum
+
+func availSpace*(bp: BipBuffer): Natural =
+  ## Returns amount of space available for reserve in buffer `bp`.
+  let (res, _) = bp.calcReserve()
+  res
+
+func usedSpace*(bp: BipBuffer): Natural =
```
arnetheduck, line 77:
```suggestion
func len*(bp: BipBuffer): Natural =
```

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+      bp.a.zero()
+      size - currentSize
+    else:
+      bp.a.start += currentSize
+      size
+  else:
+    bp.a.start += currentSize
+    size
+
+iterator bytes*(bp: BipBuffer): byte =
```
arnetheduck, line 154:
```suggestion
iterator items*(bp: BipBuffer): byte =
```

allows it to be used in for loop easily

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,167 @@
...
+    return err("Not enough space available")
+  let reserveLength = min(availableSpace, size)
+  bp.r = BipPos.init(reserveStart, Natural(reserveStart + reserveLength))
+  ok(availableSpace)
+
+proc reserve*(bp: var BipBuffer): Natural =
+  ## Reserve all available free space in buffer. Returns number of bytes
+  ## reserved.
+  ##
+  ## Note, this procedure will raise Defect if there is no free space available.
```
arnetheduck, line 95: it would be more natural to return a `Result` here, like the other `reserve` - callers that *know* there is space available can do `reserve().expect("free bytes")` - then it is clearly documented on the calling side what the expectation is

arnetheduck: basically, a single `reserve(bp: var BipBuffer, size = Natural.high()` is all that's needed

chronos/bipbuffer.nim
cheatfate: Because there is two regions, which could cover not all the buffer length we have an iterator `bytes` and `regions`, but you still can consume/decommit less bytes than the buffer has.

arnetheduck: the point was more that when you're calling this function, you already know how much data you have because the code paths that lead to it require you to acquire this information and now you call consume to tell the buffer how much data you used. 

You never call this function "speculatively" to find out how much data there was in the region.

arnetheduck: ie the suggested action point here is to remove the return value

arnetheduck: (and maybe make a Defect for trying to consume more data than is in the buffer)

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/e596388d5c2fa222433da97e96d72040473b2bcf

chronos/bipbuffer.nim
cheatfate: Primitive removed in https://github.com/status-im/nim-chronos/pull/485/commits/e596388d5c2fa222433da97e96d72040473b2bcf

chronos/bipbuffer.nim
cheatfate: Primitive removed in https://github.com/status-im/nim-chronos/pull/485/commits/e596388d5c2fa222433da97e96d72040473b2bcf

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/e596388d5c2fa222433da97e96d72040473b2bcf

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/e596388d5c2fa222433da97e96d72040473b2bcf

chronos/bipbuffer.nim
cheatfate: I think it is already fixed, no?

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/c74f451a373e4049178e2841a2b20695b5cd5615

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,147 @@
...
+  if len(bp.a) == 0 and len(bp.b) == 0:
+    bp.a.start = bp.r.start
+    bp.a.finish = bp.r.start + toCommit
+  elif bp.r.start == bp.a.finish:
+    bp.a.finish += toCommit
+  else:
+    bp.b.finish += toCommit
+  bp.r.reset()
+
+proc consume*(bp: var BipBuffer, size: Natural): Natural =
```
arnetheduck, line 115:
```suggestion
proc consume*(bp: var BipBuffer, size: Natural) =
```

I think it's reasonable to make it a `Defect` to consume more than is available - the only time you consume bytes is when you iterate over them with `regions` or discard the data with `bp.consume(bp.len)` - you never "guess" optimistically how many bytes you read from the buffer.

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,147 @@
...
+  ## Returns amount of space available for reserve in buffer `bp`.
+  let (res, _) = bp.calcReserve()
+  res
+
+func len*(bp: BipBuffer): Natural =
+  ## Returns amount of used space in buffer `bp`.
+  len(bp.b) + len(bp.a)
+
+proc reserve*(bp: var BipBuffer, pt, st: typedesc,
+              size: Natural = 0): Result[tuple[data: pt, size: st], cstring] =
```
arnetheduck, line 72:
```suggestion
              size: Natural = 0): tuple[data: pt, size: st] =
```

same here - I think we can make it a defect to try to reserve more than `bp.availSpace` - either you call this function with 0 or you can check for the available space with `availSpace` but I can't think of any situation where you call it with a larger value, that isn't a bug.

arnetheduck (approved):

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,147 @@
...
+func availSpace*(bp: BipBuffer): Natural =
+  ## Returns amount of space available for reserve in buffer `bp`.
+  let (res, _) = bp.calcReserve()
+  res
+
+func len*(bp: BipBuffer): Natural =
+  ## Returns amount of used space in buffer `bp`.
+  len(bp.b) + len(bp.a)
+
+proc reserve*(bp: var BipBuffer, pt, st: typedesc,
```
arnetheduck, line 71: do we really need `pt` and `pt` here? what else than `ptr byte`, `int` would they be? the arithmetic inside already makes this assumption - I can't for example use `ptr int` here because the resulting "length" will be in bytes, not in `number-of-integers`

chronos/bipbuffer.nim
arnetheduck: the problem is this part here: `cast[st](len(bp.r))` - if I do `st=int8` this can easily return a non-sensical value - if I do `pt = int` it _should_ be divided by the size of `pt[]` but then you also need to consider alignment and padding, and the buffer might no longer be contiguous.. - `ptr byte` is the only "safe" thing to use here without opening up a pandora's box of problems

chronos/bipbuffer.nim
arnetheduck: the problem is this part here: cast[st](len(bp.r)) - if I do st=int8 this can easily return a non-sensical value - if I do pt = int it should be divided by the size of pt[] but then you also need to consider alignment and padding, and the buffer might no longer be contiguous.. - ptr byte is the only "safe" thing to use here without opening up a pandora's box of problems

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/ef7d4563e791733aeab7e0edc906e75773ceaf42

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/ef7d4563e791733aeab7e0edc906e75773ceaf42

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/ef7d4563e791733aeab7e0edc906e75773ceaf42

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,141 @@
...
+    bp.a.start += currentSize
+
+iterator items*(bp: BipBuffer): byte =
+  ## Iterates over all the bytes in the buffer.
+  for index in bp.a.start ..< bp.a.finish:
+    yield bp.data[index]
+  for index in bp.b.start ..< bp.b.finish:
+    yield bp.data[index]
+
+iterator regions*(bp: var BipBuffer): tuple[data: ptr byte, size: int] =
```
arnetheduck, line 136: `int` vs `Natural` - some lengths in the api are returned as Natural and some as int, intended?

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,141 @@
...
+    raiseAssert ErrorMessage
+  let reserveLength =
+    if size == 0:
+      availableSpace
+    else:
+      if size < availableSpace:
+        raiseAssert ErrorMessage
+      size
+  bp.r = BipPos.init(reserveStart, Natural(reserveStart + reserveLength))
+  (cast[ptr byte](addr bp.data[bp.r.start]), len(bp.r))
```
arnetheduck, line 93:
```suggestion
  (addr bp.data[bp.r.start], len(bp.r))
```

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,141 @@
...
+  ## Iterates over all the bytes in the buffer.
+  for index in bp.a.start ..< bp.a.finish:
+    yield bp.data[index]
+  for index in bp.b.start ..< bp.b.finish:
+    yield bp.data[index]
+
+iterator regions*(bp: var BipBuffer): tuple[data: ptr byte, size: int] =
+  ## Iterates over all the regions (`a` and `b`) in the buffer.
+  if len(bp.a) > 0:
+    yield (cast[ptr byte](addr bp.data[bp.a.start]), len(bp.a))
```
arnetheduck, line 139:
```suggestion
    yield (addr bp.data[bp.a.start], len(bp.a))
```

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,141 @@
...
+  if len(bp.b) > 0:
+    yield (cast[ptr byte](addr bp.data[bp.b.start]), len(bp.b))
```
arnetheduck, line 141:
```suggestion
    yield (addr bp.data[bp.b.start], len(bp.b))
```

arnetheduck (approved):

chronos/bipbuffer.nim
```diff
@@ -0,0 +1,141 @@
...
+    bp.a.finish = bp.r.start + toCommit
+  elif bp.r.start == bp.a.finish:
+    bp.a.finish += toCommit
+  else:
+    bp.b.finish += toCommit
+  bp.r.reset()
+
+proc consume*(bp: var BipBuffer, size: Natural) =
+  ## The procedure removes/frees `size` bytes from the buffer.
+  ## Returns number of bytes actually removed.
```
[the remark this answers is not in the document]
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/76bb1ac88d422057a68ebb0a53f858ab15be0f4e

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/76bb1ac88d422057a68ebb0a53f858ab15be0f4e

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/76bb1ac88d422057a68ebb0a53f858ab15be0f4e

chronos/bipbuffer.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/485/commits/76bb1ac88d422057a68ebb0a53f858ab15be0f4e

PR #486, opened 2023-12-28
prevent http `closeWait` future from being cancelled

arnetheduck:
* simplify `closeWait` implementations
  * remove redundant cancellation callbacks
  * use `noCancel` to avoid forgetting the right future flags
* add a few missing raises trackers

### DIFF
--- a/chronos/streams/asyncstream.nim
+++ b/chronos/streams/asyncstream.nim
@@ -901,6 +901,8 @@ proc closeWait*(rw: AsyncStreamRW): Future[void] {.async: (raises: []).} =
   ## Close and frees resources of stream ``rw``.
   if not rw.closed():
     rw.close()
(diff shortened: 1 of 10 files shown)


chronos/streams/asyncstream.nim
```diff
@@ -897,44 +897,29 @@ proc close*(rw: AsyncStreamRW) =
...
-  if rw.future.finished():
-    retFuture.complete()
-  else:
-    rw.future.addCallback(continuation, cast[pointer](retFuture))
-  retFuture
+  if not rw.closed():
+    rw.close()
+    # noCancel needed here to prevent cancellation from reaching rw.future -
+    # see assert on init
+    await noCancel rw.future
```
cheatfate, line 906:
```nim
await noCancel(rw.join())

cheatfate (approved):

PR #487, opened 2023-12-28
check leaks after every test

arnetheduck:


cheatfate:
Very confusing PR name, because leaks already got checked in all the tests, so probably should be renamed. And problem not in leaks problem is in `all_tests` approach. When all tests combined into single package memory usage become enormous because Nim memory manager do not release memory.

arnetheduck:
> Very confusing PR name, because leaks already got checked in all the tests,

leaks are checked once per `suite` currently - after this PR they get checked after every `test` - that's the difference - ie the PR title follows the naming in `unittest2`.

When leaks are tested once per suite, you don't know which test within a suite actually introduced the leak and have to look for it one by one - with this PR, we get a more accurate representation.

Issue #489, opened 2024-01-11
`getAcceptInfo` can raise an unlisted exception: ref ValueError


EmilIvanichkovv:
If I try to compile `httpserver.nim` like that:
```bash
nim c chronos/apps/http/httpserver.nim
```
I get the following error:
```bash
nim-chronos/chronos/apps/http/httpserver.nim:563:16 Error: getAcceptInfo(getString(request.headers, AcceptHeaderName, "")) can raise an unlisted exception: ref ValueError
```

I trigger the problem using `nim v1.6.14` ( same result with `nim v1.6.18` )

EmilIvanichkovv:
With [this PR](https://github.com/status-im/nim-chronos/pull/490) I suggest a simple solution for this problem, and other unlisted exceptions in functions in the file

PR #490, opened 2024-01-11
Fix unlisted exceptions in `chronos/apps/http/httpserver.nim`

EmilIvanichkovv:
Fix for #489

### DIFF
--- a/chronos/apps/http/httpserver.nim
+++ b/chronos/apps/http/httpserver.nim
@@ -555,14 +555,16 @@ proc consumeBody*(request: HttpRequestRef): Future[void] {.
   finally:
     await reader.closeWait()
 
-proc getAcceptInfo*(request: HttpRequestRef): Result[AcceptInfo, cstring] =
+proc getAcceptInfo*(request: HttpRequestRef): Result[AcceptInfo, cstring] {.
+     raises: [ref ValueError].} =
   ## Returns value of `Accept` header as `AcceptInfo` object.
   ##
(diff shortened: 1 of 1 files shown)


chronos/apps/http/httpserver.nim
```diff
@@ -555,14 +555,16 @@ proc consumeBody*(request: HttpRequestRef): Future[void] {.
   finally:
     await reader.closeWait()
 
-proc getAcceptInfo*(request: HttpRequestRef): Result[AcceptInfo, cstring] =
+proc getAcceptInfo*(request: HttpRequestRef): Result[AcceptInfo, cstring] {.
```
arnetheduck, line 558: in general, functions returning `Result` should not raise excpetions - in this particular case, you should trace the call chain down to where the exception is being raised and catch it, translating it to a string instead.

That said, probably you have out-of-date dependencies - this error is not happening in CI

EmilIvanichkovv: > That said, probably you have out-of-date dependencies - this error is not happening in CI
Hmm
I did the following steps:
- git clone 
- nimble install
- nimble setup
- nim c chronos/apps/http/httpserver.nim

arnetheduck: Nimble dependency resolution is broken - you can often get it to work by removing its cache (`rm -rf $HOME/.nimble`) cc @zah

EmilIvanichkovv: Oddly it helped. Ty!

PR #492, opened 2024-01-17
deprecate `callback=`, UDP fixes (fixes #491)

### LINKED ISSUE #491: Apparent garbage collect of localCopy in datagram sendTo before data gets send in writeDatagramLoop
We've been hitting an issue in our uTP implementation in Fluffy where garbage data would be send over the wire (UDP).

It occurs on nim-chronos#head, but bisecting the commits points at this commit as first occurance: https://github.com/status-im/nim-chronos/commit/f5ff9e32ca4bd45781426246e4fb6b9fcceff0c2

What I think that happens is that the  `localCopy` var (https://github.com/status-im/nim-chronos/blob/92acf68b04070dfe8eb65bab71fbf63804979a16/chronos/transports/datagram.nim#L967) gets GC'ed before the data gets send in the `writeDatagramLoop` (https://github.com/status-im/nim-chronos/blob/92acf68b04070dfe8eb65bab71fbf63804979a16/chronos/transports/datagram.nim#L431)

What supports this theory is that adding a `GC_ref(localCopy)` as quick test did make the issue no longer occur.

I've tried to minimized to small testcase, but apparently minimized too far as I could no longer reproduce. I'll still try to create a reproducible testcase. 
For now, one way to reproduce is to run this `nim c -r -d:release ./tests/utp/test_protocol.nim` on this branch of nim-eth:  https://github.com/status-im/nim-eth/pull/654

It should be noted that the send data code in nim-eth uTP (https://github.com/status-im/nim-eth/blob/a6942e3a303d75e2d4b59242eb7a4f35fc331deb/eth/utp/utp_socket.nim#L455) isn't exactly pretty but it doesn't look fully "illegal" either.

That being said, this change on nim-eth uTP side does appear to fix the issue: https://github.com/status-im/nim-eth/pull/655/commits/7b21a48099169b357caffd23a21cddfce0a95cf1

Similar code exists for the TCP version (https://github.com/status-im/nim-chronos/blob/master/chronos/transports/stream.nim#L2299), so perhaps a similar issue is possible there too.

arnetheduck:
Using the callback setter may lead to callbacks owned by others being reset, which is unexpected.

* don't crash on zero-length UDP writes

### DIFF
--- a/chronos/transports/datagram.nim
+++ b/chronos/transports/datagram.nim
@@ -13,6 +13,7 @@ import std/deques
 when not(defined(windows)): import ".."/selectors2
 import ".."/[asyncloop, config, osdefs, oserrno, osutils, handles]
 import "."/common
+import stew/ptrops
 
 type
   VectorKind = enum
@@ -119,7 +120,7 @@ when defined(windows):
         ## Initiation
         transp.state.incl(WritePending)
         let fd = SocketHandle(transp.fd)
-        var vector = transp.queue.popFirst()
+        let vector = transp.queue.popFirst()
         transp.setWriterWSABuffer(vector)
         let ret =
           if vector.kind == WithAddress:
@@ -365,7 +366,7 @@ when defined(windows):
                                       udata: cast[pointer](res))
     res.wovl.data = CompletionData(cb: writeDatagramLoop,
                                       udata: cast[pointer](res))
-    res.rwsabuf = WSABUF(buf: cast[cstring](addr res.buffer[0]),
+    res.rwsabuf = WSABUF(buf: cast[cstring](baseAddr res.buffer),
                          len: ULONG(len(res.buffer)))
     GC_ref(res)
     # Start tracking transport
@@ -392,7 +393,7 @@ else:
     else:
       while true:
         transp.ralen = SockLen(sizeof(Sockaddr_storage))
-        var res = osdefs.recvfrom(fd, addr transp.buffer[0],
+        var res = osdefs.recvfrom(fd, baseAddr transp.buffer,
                                   cint(len(transp.buffer)), cint(0),
                                   cast[ptr SockAddr](addr transp.raddr),
                                   addr transp.ralen)
@@ -424,7 +425,7 @@ else:
       transp.state.incl({WritePaused})
     else:
       if len(transp.queue) > 0:
-        var vector = transp.queue.popFirst()
+        let vector = transp.queue.popFirst()
         while true:
           if vector.kind == WithAddress:
             toSAddr(vector.address, transp.waddr, transp.walen)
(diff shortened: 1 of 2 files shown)


chronos/transports/datagram.nim
```diff
@@ -1002,7 +1002,7 @@ proc getMessage*(transp: DatagramTransport): seq[byte] {.
     copyMem(addr res[0], addr transp.buffer[0], transp.buflen)
     res
   else:
-    default
+    @[]
```
cheatfate, line 1005: why allocation?

arnetheduck: there is no allocation for an empty seqs:  https://github.com/nim-lang/Nim/blob/v1.6.14/compiler/ccgexprs.nim#L1416C10-L1416C10

cheatfate: I'm not very familiar with C compiler source code but i'm not sure that code you point to will be executed when compiler will encounter `@` symbol.
Because in `system.nim` there present
```nim
proc `@`* [IDX, T](a: sink array[IDX, T]): seq[T] {.magic: "ArrToSeq", noSideEffect.}
```
Which performs something else from `newSeq[byte](0)`, so if you do not like my legacy approach you could convert this line to
```nim
default(seq[byte])
```

arnetheduck: The implementation is here: 

https://github.com/nim-lang/Nim/blob/71ba2e7f3c5815d956b1ae0341b0743242b8fec6/compiler/ccgexprs.nim#L1576C29-L1576C29 

https://github.com/nim-lang/Nim/blob/71ba2e7f3c5815d956b1ae0341b0743242b8fec6/compiler/ccgexprs.nim#L1559

This follows from the compile-time knowledge of the length of an array - @ always uses their length to create sequences in the most efficient way possible (by pre-allocating a correctly lengthed seq or using `nil` when no allocation is needed).

We can test this theory in a simple way:

```nim
proc myfunc*() =
  var vvv: seq[byte] = @[]
  var vvv2: seq[byte] = default(seq[byte])
myfunc()```

```c
N_LIB_PRIVATE N_NIMCALL(void, myfunc__testit_3)(void) {
        tySequence__6H5Oh5UUvVCLiakt9aTwtUQ* vvv;
        tySequence__6H5Oh5UUvVCLiakt9aTwtUQ* vvv2;
        nimfr_("myfunc", "/home/arnetheduck/status/nimbus-eth2/_exper/testit.nim");
        vvv = NIM_NIL;
        vvv2 = NIM_NIL;
        popFrame();
}
```

cheatfate: Ok, thank you, but let it will be styling requirement.

arnetheduck: 60c3a42b7fedc492c785755161f73d407039df95

jangko:
Thank you. This help fix discovery 4 UDP data corruption when compiled with nim 2.0

kdeme:
> Thank you. This help fix discovery 4 UDP data corruption when compiled with nim 2.0

Ah yes, I hadn't realized that discv4 does the same `f.callback ` in its send call. We should get that fix already in a separate PR as it can/will occur on nim 1.6 also.

PR #498, opened 2024-01-23
Add async waitNotEmpty proc

diegomrsantos:
This PR introduces a new proc, waitNotEmpty, to our AsyncQueue implementation. It suspends the execution asynchronously until at least one item is available in the queue. It is useful for use cases involving a priority and regular queue. Additionally, it can be used to reduce code duplication in other procs.


diegomrsantos:
I think I'll have to close this PR again. I was trying to create an async `waitNotEmpty` or even better an async `peek`, but I'm not sure this can be done correctly as we always wake up one getter when an element is added to the queue.

PR #502, opened 2024-02-13
Fix HTTP server accept() loop exiting under heavy load.

cheatfate:


chronos/apps/http/httpserver.nim
```diff
@@ -1118,29 +1106,42 @@ proc processLoop(holder: HttpConnectionHolderRef) {.async: (raises: []).} =
...
-      if resId.isErr():
-        # We are unable to identify remote peer, it means that remote peer
-        # disconnected before identification.
-        await transp.closeWait()
-        runLoop = false
-      else:
-        let connId = resId.get()
-        let holder = HttpConnectionHolderRef.new(server, transp, resId.get())
-        server.connections[connId] = holder
+  block mainLoop:
```
arnetheduck, line 1109: it's pretty hard to follow where the breaks go in this structure .. why not:

```
try:
  while true:
    try:
    except ..:
except ..:
```

arnetheduck (approved):

chronos/apps/http/httpserver.nim
cheatfate: Mostly because i do not trust nested `try/except` blocks. I'm going to introduce big test suite with all the `try/except` things i will find and expected results. So as soon as this test suite will be part of chronos i will adopt new schemes. Also there is only one statement which could raise an exception, so what the reason to have nested blocks?

arnetheduck: well, named blocks are just a little tricky to understand in terms of control flow so if there's an alternative way of structuring the code, this is probably desireable for the reader.. 

one "trick" I've been using to reduce the cognitive load of exceptions for `CancellationError` in particular (only) is to not catch it in these "loop" constructs but instead let it bubble up - if the loop owner uses `cancelAndWait` to stop the loop this has the desired effect of cancelling the loop but with less syntax (while still being explicit in the `raises: [CancelledError]` that it can happen).

cheatfate: Here we should stop `CancelledError` from being raised to upper level, we just going to break loop and return result.

cheatfate: And i afraid Nim exception transformation for closure iterators is not that stable as we all want it to be.

arnetheduck: The strategy to write a test case that confirms its correctness seems preferable to the introduction of hard-to-read code "just in case" :+1:

cheatfate: I agree that it hard to read, but this is workaround.

PR #512, opened 2024-02-29
Add automatic constructors for TCP and UDP transports.

cheatfate:


chronos/transports/stream.nim
```diff
@@ -2130,6 +2130,36 @@ proc createStreamServer*(host: TransportAddress,
   createStreamServer(host, StreamCallback2(nil), flags, sock, backlog, bufferSize,
                      child, init, cast[pointer](udata), dualstack)
 
+proc createStreamServer*(flags: set[ServerFlags] = {},
```
arnetheduck, line 2133: we need to be able to specify a port here because their typical use in servers is to listen on "any address, an ip version, port XXX".

arnetheduck: also, without the port, this looks like a constructor for unix domain sockets

cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/512/commits/9101e994591d096bc9df95b0fb2ad7d4447b29a2

chronos/transports/datagram.nim
```diff
@@ -824,6 +824,37 @@ proc newDatagramTransport6*[T](cbproc: UnsafeDatagramCallback,
                              cast[pointer](udata), child, bufSize, ttl,
                              dualstack)
 
+proc newAutoDatagramTransport*(cbproc: DatagramCallback,
```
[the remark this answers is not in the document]
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/512/commits/9101e994591d096bc9df95b0fb2ad7d4447b29a2

chronos/transports/stream.nim
```diff
@@ -2115,6 +2218,50 @@ proc createStreamServer*(host: TransportAddress,
   createStreamServer(host, StreamCallback2(nil), flags, sock, backlog, bufferSize,
                      child, init, cast[pointer](udata), dualstack)
 
+proc createStreamServer*(port = Port(0),
```
arnetheduck, line 2221:
```suggestion
proc createStreamServer*(port: Port,
```

Maybe this makes sense, so that one cannot call `createStreamServer()` - when using `Port`, it is implicit that we want an IP socket (and not unix) - this way, `createStreamServer(Port(0))` is explicitly required to force the "type" of server

arnetheduck (approved):

chronos/transports/stream.nim
cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/512/commits/5de7b1dd6362d9c93d270a4d7a074d804297c8c6

Issue #516, opened 2024-03-02
`cancelAndWait` on `or` triggers deadlock


etan-status:
this leads to deadlock:
```nim
import chronos

proc f(): Future[void] {.async.} =
  await sleepAsync(10.seconds) or sleepAsync(5.seconds)

proc g(): Future[void] {.async.} =
  let fx = f()
  await fx.cancelAndWait()

waitFor g()

echo "ok"
```

this works fine:

```nim
import chronos

proc f(): Future[void] {.async.} =
  discard await race(sleepAsync(10.seconds), sleepAsync(5.seconds))

proc g(): Future[void] {.async.} =
  let fx = f()
  await fx.cancelAndWait()

waitFor g()

echo "ok"
```

PR #517, opened 2024-03-03
Fix `or` deadlock issue.

cheatfate:
Address issue #516.

### DIFF
--- a/chronos/internal/asyncfutures.nim
+++ b/chronos/internal/asyncfutures.nim
@@ -786,7 +786,7 @@ template orImpl*[T, Y](fut1: Future[T], fut2: Future[Y]): untyped =
   fut2.addCallback(cb)
 
   retFuture.cancelCallback = cancellation
-  return retFuture
+  retFuture
 
 proc `or`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
   ## Returns a future which will complete once either ``fut1`` or ``fut2``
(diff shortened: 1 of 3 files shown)


chronos/internal/asyncfutures.nim
```diff
@@ -1667,8 +1667,7 @@ proc `or`*[T, Y, E1, E2](
   type
     InternalRaisesFutureRaises = union(E1, E2)
 
-  let
-    retFuture = newFuture[void]("chronos.wait()", {FutureFlag.OwnCancelSchedule})
+  let retFuture = newFuture[void]("chronos.or()", {})
```
arnetheduck, line 1670: `cancelSoon(fut or fut2)` will defect in this case?

cheatfate: why?

arnetheduck: nevermind, my mistake

chronos/internal/asyncfutures.nim
```diff
@@ -1667,8 +1667,7 @@ proc `or`*[T, Y, E1, E2](
   type
     InternalRaisesFutureRaises = union(E1, E2)
```
arnetheduck, line 1668: this might be a problem if neither of E1, E2 raise `CancelledError`

cheatfate: How i can do this? Because looks like `prepend` and `union` are not working here.

cheatfate: Fixed in https://github.com/status-im/nim-chronos/pull/517/commits/d5fd6663b2997328fda2a63aff9c9f69955c5b3f

PR #522, opened 2024-03-18
Work around type resolution with empty generic

arnetheduck:


arnetheduck:
* https://github.com/nim-lang/Nim/issues/23418
* https://github.com/vacp2p/nim-libp2p/pull/1068

PR #526, opened 2024-03-22
make `Raising` compatible with 2.0

arnetheduck:
See https://github.com/nim-lang/Nim/issues/23432

### DIFF
--- a/tests/testfut.nim
+++ b/tests/testfut.nim
@@ -2064,8 +2064,8 @@ suite "Future[T] behavior test suite":
     f(v)
 
     type Object = object
-      field: Future[void].Raising([ValueError])
       # TODO cannot use X[[ValueError]] here..
+      field: Future[void].Raising([ValueError])
     discard Object(field: v)
 
     check:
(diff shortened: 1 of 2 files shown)


tests/testfut.nim
```diff
@@ -2047,9 +2047,27 @@ suite "Future[T] behavior test suite":
...
+    type X[E] = Future[void].Raising(E)
+
+    proc f(x: X) = discard
+
+    var v: Future[void].Raising([ValueError])
+    f(v)
+
+    type Object = object
+      field: Future[void].Raising([ValueError])
+
```
arnetheduck, line 2068:
```suggestion
      field: Future[void].Raising([ValueError])
      # TODO cannot use X[[ValueError]] here..
```

tests/testfut.nim
```diff
@@ -2047,9 +2047,27 @@ suite "Future[T] behavior test suite":
...
+    type X[E] = Future[void].Raising(E)
+
+    proc f(x: X) = discard
+
+    var v: Future[void].Raising([ValueError])
+    f(v)
+
+    type Object = object
+      field: Future[void].Raising([ValueError])
+      # TODO cannot use X[[ValueError]] here..
```
arnetheduck, line 2068:
```suggestion
      # TODO cannot use X[[ValueError]] here..
      field: Future[void].Raising([ValueError])
```

PR #545, opened 2024-05-22
Fixes compilation issues in v3 compatibility mode (`-d:chronosHandleException`)

### LINKED ISSUE #544: Compilation errors with `-d:chronosHandleException`
As part of our V4 migration strategy we've been investigating using the V3 "compatibility" mode (i.e., compiling with `-d:chronosHandleException`) with Chronos, but I'm running into a compilation issue (`nim 1.6.20`, `Linux x86_64`) . The following example:

```nim
import chronos

proc myProc() {.async.} = 
  echo "hello"
```
will fail with:
```text
Error: expression 'cbproc(transp, remote)' is of type 'Future[system.void]' and has to be used (or discarded); start of expression here: /.../chronos/transports/datagram.nim(635, 5)
``` 

Looking at the offending code, I see:
```nim
proc wrap(transp: DatagramTransport, remote: TransportAddress) {.async: (raises: []).} =
  try:
    cbproc(transp, remote)
  except CatchableError as exc:
    raiseAssert "Unexpected exception from stream server cbproc: " & exc.msg
```
where `cbproc` is of type:
```nim
UnsafeDatagramCallback* = proc(transp: DatagramTransport,
                           remote: TransportAddress): Future[void] {.async.}
```
I suppose the intent here is to make this into a safe `DatagramCallback`:
```nim
DatagramCallback* = proc(transp: DatagramTransport,
                           remote: TransportAddress): Future[void] {.
                           async: (raises: []).}
```
 at which point this seems to be missing a call to `await`? 

As a side note, this code as it is now is pretty strange even when it compiles - the Future returned by `cbproc` is never set as a return value in the expansion of `wrap` (produced with `-d:chronosDumpAsync`); i.e., it doesn't seem to be calling `setResult` as other procs do:

```nim
proc wrap ... =
  iterator wrap_1895826090(chronosInternalRetFuture: FutureBase): FutureBase {.
      closure, gcsafe, raises: [].} =
    template result(): auto {.used.} = ...

    var closureSucceeded_1895826089 = true
    # no setResult do:
    try:
      try:
        cbproc(transp, remote)
      except CatchableError as exc:
        raiseAssert "Unexpected exception from stream server cbproc: " & exc.msg
    except Defect as exc:
      closureSucceeded_1895826089 = false
      raise exc
    finally:
      if closureSucceeded_1895826089:
        complete(cast[Future[void]](chronosInternalRetFuture))

```
I tried to figure out the implication of this by doing some testing on iterators and seeing what happens to those "fall through" values, and it looks like they're simply lost; e.g.:

```nim
proc proc1(): iterator = 
  iterator: int {.closure.} = 2

proc proc2(): iterator = 
  iterator: int {.closure.} = return 2

echo proc1()() # prints 0
echo proc2()() # prints 2
```

so not really sure what happens to the Future returned from `cbproc` after the call (guess it just runs as if we had done `discard cbproc(...)`?).

I'm opening a PR adding the calls to await as that seems to fix it, doesn't break anything, and seems to reflect the actual intent, but considering the above I'm not sure that's the only issue.

gmega:
fixes #544 

This PR adds the missing `await` calls which fixes the issue I was experiencing without visible ill effects. I've also added (somewhat naively) the ability to run the test suite in compat mode, though not sure this is a desirable change.

### DIFF
--- a/chronos/transports/datagram.nim
+++ b/chronos/transports/datagram.nim
@@ -720,7 +720,7 @@ proc newDatagramTransportCommon(cbproc: UnsafeDatagramCallback,
   proc wrap(transp: DatagramTransport,
             remote: TransportAddress) {.async: (raises: []).} =
     try:
-      cbproc(transp, remote)
+      await cbproc(transp, remote)
     except CatchableError as exc:
       raiseAssert "Unexpected exception from stream server cbproc: " & exc.msg
 
--- a/tests/testmacro.nim
+++ b/tests/testmacro.nim
@@ -8,6 +8,7 @@
 import std/[macros, strutils]
 import unittest2
 import ../chronos
+import ../chronos/config
 
 {.used.}
 
@@ -403,9 +404,10 @@ suite "Exceptions tracking":
     expect(IOError): waitFor test3()
     waitFor test6()
 
-  test "Cannot raise invalid exception":
-    checkNotCompiles:
-      proc test3 {.async: (raises: [IOError]).} = raise newException(ValueError, "hey")
+  when not chronosHandleException:
+    test "Cannot raise invalid exception":
+      checkNotCompiles:
+        proc test3 {.async: (raises: [IOError]).} = raise newException(ValueError, "hey")
 
   test "Explicit return in non-raising proc":
     proc test(): Future[int] {.async: (raises: []).} = return 12
@@ -430,11 +432,12 @@ suite "Exceptions tracking":
     checkNotCompiles:
       a = test2()
 
-  test "Await raises the correct types":
-    proc test1 {.async: (raises: [ValueError]).} = raise newException(ValueError, "hey")
-    proc test2 {.async: (raises: [ValueError, CancelledError]).} = await test1()
-    checkNotCompiles:
-      proc test3 {.async: (raises: [CancelledError]).} = await test1()
+  when not chronosHandleException:
+    test "Await raises the correct types":
+      proc test1 {.async: (raises: [ValueError]).} = raise newException(ValueError, "hey")
+      proc test2 {.async: (raises: [ValueError, CancelledError]).} = await test1()
+      checkNotCompiles:
+        proc test3 {.async: (raises: [CancelledError]).} = await test1()
 
   test "Can create callbacks":
     proc test1 {.async: (raises: [ValueError]).} = raise newException(ValueError, "hey")
@@ -447,8 +450,9 @@ suite "Exceptions tracking":
     proc test2: Future[int] {.async: (raises: [ValueError, IOError, CancelledError]).} =
       return await test1()
 
-    checkNotCompiles:
-      proc test3: Future[int] {.async: (raises: [CancelledError]).} = await test1()
+    when not chronosHandleException:
+      checkNotCompiles:
+        proc test3: Future[int] {.async: (raises: [CancelledError]).} = await test1()
 
     check waitFor(test2()) == 12
 
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -60,6 +60,14 @@ task test, "Run all tests":
       run args & " --mm:refc", "tests/testall"
     run args, "tests/testall"
 
+task test_v3_compat, "Run all tests in v3 compatibility mode":
+  for args in testArguments:
+    if (NimMajor, NimMinor) > (1, 6):
+      # First run tests with `refc` memory manager.
+      run args & " --mm:refc -d:chronosHandleException", "tests/testall"
+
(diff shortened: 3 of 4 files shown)


tests/testmacro.nim
```diff
@@ -403,9 +404,10 @@ suite "Exceptions tracking":
     expect(IOError): waitFor test3()
     waitFor test6()
 
-  test "Cannot raise invalid exception":
-    checkNotCompiles:
-      proc test3 {.async: (raises: [IOError]).} = raise newException(ValueError, "hey")
+  when not chronosHandleException:
```
arnetheduck, line 407: why does this test need to be disabled? the global handleException default should not affect raises-annotated functions..

gmega: Oh... apparently it does, as this snippet starts to compile when I enable it; i.e., it seems to relax the raises constraint in raises-annotated procs as well. I thought this was intentional.

gmega: OK, running a simple test with the raises-annotated proc above and looking at the expansions I can see that the global flag does indeed modify the exception handlers (adds handlers for `Exception` and `CatchableError`, removes handler for `IOError`).

gmega: Hm... if I understand the implementation right:

```nim
proc decodeParams(params: NimNode): AsyncParams =
  ...
  var
    raw = false
    raises: NimNode = nil
    handleException = chronosHandleException
  ...
  if param[0].eqIdent("raises"):
    ... # doesn't reset handleException to false
  elif param[0].eqIdent("handleException"):
      handleException = param[1].eqIdent("true")
```
this will indeed simply start from the global value and then override it if `handleException` is set, but won't set it to false if there's a raises clause.

gmega: I'm experimenting with fixing this and seeing how it impacts our code.

gmega: OK so I've added an attempt at fixing this to this branch as it sort of fits the overall "compilation issues in v3 compatibility mode" theme, and I need those fixes to compile with the flag on anyways. :-)

gmega: OK here goes my attempt at fixing this. I've added it to this branch as it sort of fits the overall "compilation issues in v3 compatibility mode" theme, and I'd need the previous changes to be able to compile Chronos with the flag on anyways.

chronos/internal/asyncmacro.nim
```diff
@@ -219,7 +219,8 @@ proc decodeParams(params: NimNode): AsyncParams =
   var
     raw = false
     raises: NimNode = nil
-    handleException = chronosHandleException
+    handleException = false
+    handleExceptionOverridden = false
```
gmega, line 223: I tried to avoid using this, but I couldn't think of a way to otherwise distinguish a local override where the annotation in the proc sets `handleException` to false (which you shouldn't override with the global flag) vs when nothing has been set and you're just looking at the initialization value (in which case you should override with the global flag).

arnetheduck:
Hm, this looks like it's cycling down the wrong path. Let's revert to what handleExceptions is supposed to do: remap `Exception` to `AsyncExceptionError` - this means that for any function where `handleExceptions` is enabled _and_ it has `raises`, the raises list also needs to include `AsyncExceptionError` - of course, in legacy code this is not a problem (because `async` without `raises` allows any exception to pass), but the failing tests are likely failing because of this "extra" exception being raised - the solution is to write the tests such that when `hndleExceptions` is globally enabled, they have a different raises list

chronos/transports/datagram.nim
```diff
@@ -720,7 +720,7 @@ proc newDatagramTransportCommon(cbproc: UnsafeDatagramCallback,
   proc wrap(transp: DatagramTransport,
             remote: TransportAddress) {.async: (raises: []).} =
     try:
-      cbproc(transp, remote)
+      await cbproc(transp, remote)
```
arnetheduck, line 723: :+1: this part is good

arnetheduck:
hm, or .. maybe you're right, maybe it doesn't make sense for the global flag to apply to raises-annotated functions in which case the documentation needs to be updated as well. 

This makes more sense in the context of upgrading an existing codebase with mixed annotated and non-annotated code - let's go with your proposal, forget my previous comment.

gmega:
Yeah, even though my proposal actually comes from my understanding of what I thought was your proposal ( :smile: ), I came to realize that being able to get incremental exception checking as I annotate a legacy codebase doesn't seem bad at all. Gonna take a stab at updating the docs.

gmega:
OK, added a tentative update to the docs.

arnetheduck (approved):

PR #547, opened 2024-06-05
pretty-printer for `Duration`

arnetheduck:


chronos/timer.nim
```diff
@@ -370,53 +370,39 @@ template add(a: var string, b: Base10Buf[uint64]) =
...
+  ## number of parts returned can be limited thus truncating the output to
+  ## an approximation that grows more precise as the duration becomes smaller
+  var
+    res = ""
+    v = a.nanoseconds()
+    parts = parts
+
+  template f(n: string, T: Duration) =
+    if v >= T.nanoseconds():
+      res.add($(uint64(v div T.nanoseconds())))
```
cheatfate, line 384: Use `Base10.toBytes` because it is significantly faster `$`.

arnetheduck: 7ebdbe439035220e899b1ecaa395660f8ed72ae2

PR #551, opened 2024-07-04
Fix tests to be string hash order independent

c-blake:
For new Farm Hash-based hash(x:string).


narimiran:
I haven't explored the reported error in detail, but it looks like it might be a missing `import std/algorithm` problem, to have `sorted` on seqs.

c-blake:
`std/algorithm` is there right at the top of the file.  I think it is more missing a `<` as the error message says.  I pushed a commit to define one.  Can you re-run the tests?  Thanks.

c-blake:
Another compile error, another `git push` to try to fix it.  :-)

Needless to say, I am not running these tests locally (I am not set up to).  Much appreciated if you could approve the workflow again.

c-blake:
Ok.  Naming fields with `tuple[]` is non-optional. Dumb commit message.

This time I at least confirmed compilation & functioning locally of
```Nim
import std/algorithm
proc `<`(a, b: (string,string)): bool = a[0] < b[0]
echo @[("Header2", "value2"), ("Header2", "VALUE3"), ("Header1", "value1")].sorted
```
So, I have high hopes this version will work.  Thanks for approving the workflow yet again!

c-blake:
Ack.  I just kept mis-reading compiler errors.  This works for me locally and should work now if you approve the workflow yet again (famous last words, as they say!):
```Nim
import std/[algorithm, assertions]
proc `<`(a, b: (string, seq[string])): bool = a[0] < b[0]
let data1 = @[("Header2", "value2"), ("Header2", "VALUE3"), ("Header1", "value1")]
let data2 = @[("Header2", @["value2", "VALUE3"]), ("Header1", @["value1"])]
assert data1.sorted == sorted(@[("Header2", "value2"), ("Header2", "VALUE3"),
                                ("Header1", "value1")])
assert data2.sorted == sorted(@[("Header2", @["value2", "VALUE3"]),
                                ("Header1", @["value1"])])
```

narimiran:
Heh, we're back at `Error: type mismatch: got <seq[string], seq[string]>`

c-blake:
I don't know what's wrong now.  That above snippet really does run fine locally on nim-devel.  I could add a `<(a,b: seq[string])`, but I'm not sure if that's ok.  Do you know what's wrong?  Is this some known gotcha with the test suite thing?  Should I lift that `<` to outermost scope?

c-blake:
Well, with yesterday's final commit (lifting `<` to outermost scope as suggested in the last comment here), the checks pass now on nim-devel (& older).

There may be other considerations in tagging a new Chronos release so as to pass the main Nim CI, but as far as this test failure is concerned, I think the work is done and this PR is ok to merge.  Let me know if this is not the case.

Thanks

PR #556, opened 2024-08-29
make SomeFuture public

diegomrsantos:
It makes it possible to not use `InternalRaisesFuture` in procs like https://github.com/vacp2p/nim-libp2p/pull/1175/files.

### DIFF
--- a/chronos/internal/asyncfutures.nim
+++ b/chronos/internal/asyncfutures.nim
@@ -60,7 +60,7 @@ type
   FutureCompletedError* = object of FutureError
     ## Error raised when trying access the error of a completed Future
 
-  SomeFuture = Future|InternalRaisesFuture
+  SomeFuture* = Future|InternalRaisesFuture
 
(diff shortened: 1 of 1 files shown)


chronos/internal/asyncfutures.nim
```diff
@@ -60,7 +60,7 @@ type
   FutureCompletedError* = object of FutureError
     ## Error raised when trying access the error of a completed Future
 
-  SomeFuture = Future|InternalRaisesFuture
+  SomeFuture* = Future|InternalRaisesFuture
```
arnetheduck, line 63: this type is deliberately private because the whole idea of how this is implemented might change soonish - for now, it's better to use `InternalRaisesFuture` directly in code that needs it which will make it easy to find, once a stable future type is created.

arnetheduck:
these functions can use `proc test[T; F: Future[T]](f: F)`

PR #557, opened 2024-10-02
Add peak to AsyncQueue

bhartnett:
In some of my code I found the need to do something like this:
```
# wait for the first block data to be put on the queue
# so that we can access the first block once available
while blockDataQueue.empty():
# peek but don't remove it so that it can be processed later
let firstBlock = blockDataQueue[0]
```

Having a async peak proc that waits for the first item to be added to the queue would be good. The std/deques supports peakFirst and peakLast as well so this PR adds these to the AsyncQueue.

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -257,6 +257,12 @@ proc popLastImpl[T](aq: AsyncQueue[T]): T =
   aq.putters.wakeupNext()
   res
 
+proc peakFirstImpl[T](aq: AsyncQueue[T]): T =
+  aq.queue.peekFirst()
+
+proc peakLastImpl[T](aq: AsyncQueue[T]): T =
+  aq.queue.peekLast()
+
 proc addFirstNoWait*[T](aq: AsyncQueue[T], item: T) {.
      raises: [AsyncQueueFullError].} =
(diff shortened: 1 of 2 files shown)


chronos/asyncsync.nim
```diff
@@ -377,6 +444,11 @@ proc get*[T](aq: AsyncQueue[T]): Future[T] {.
   ## Alias of ``popFirst()``.
   aq.popFirst()
 
+proc peak*[T](aq: AsyncQueue[T]): Future[T] {.
```
arnetheduck, line 447:
```suggestion
proc peek*[T](aq: AsyncQueue[T]): Future[T] {.
```

bhartnett: My bad, will fix the spelling shortly.

tests/testsync.nim
```diff
@@ -353,6 +353,19 @@ suite "Asynchronous sync primitives test suite":
   test "AsyncQueue() contains test":
     check test9() == true
 
+  test "AsyncQueue() peak test":
+    let q = newAsyncQueue[int]()
+    q.putNoWait(1)
+    q.putNoWait(2)
+
+    check:
```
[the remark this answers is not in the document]
bhartnett: I've added more tests as suggested.

PR #564, opened 2025-01-29
Remove symbols which were deprecated 6 years ago.

cheatfate:


arnetheduck:
this would require a 5.0 release ..

PR #575, opened 2025-05-29
Add asMoment helper function

gmelodie:
Added a function to transform a `DateTime` into a `Moment` by converting the `DateTime` to a unix timestamp and from that to `Moment` (created it for use here https://github.com/vacp2p/nim-libp2p/pull/1358#discussion_r2114486483)

### DIFF
--- a/chronos/timer.nim
+++ b/chronos/timer.nim
@@ -31,8 +31,7 @@ const asyncTimer* {.strdefine.} = "mono"
 
 when defined(windows):
   when asyncTimer == "system":
-    proc fastEpochTime*(): uint64 {.
-         inline, deprecated: "Use Moment.now()".} =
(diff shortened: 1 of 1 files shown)


chronos/timer.nim
```diff
@@ -31,8 +31,7 @@ const asyncTimer* {.strdefine.} = "mono"
 
 when defined(windows):
   when asyncTimer == "system":
-    proc fastEpochTime*(): uint64 {.
-         inline, deprecated: "Use Moment.now()".} =
+    proc fastEpochTime*(): uint64 {.inline, deprecated: "Use Moment.now()".} =
```
arnetheduck, line 34: unrelated formatting changes are best left out of feature PR:s (notably, chronos does not use nph and it's thus best to follow the local style of the file)

PR #577, opened 2025-07-10
chore: utilize newSeqUninitialized

vladopajic:
utilizing `newSeqUninitialized` instead of `newSeq` to gain bit more performance when initializing sequences.

in this pr, `newSeqUninitialized` is utilized only in places where it can directly replace `newSeq` without any change to other logic.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -1027,7 +1027,7 @@ proc new*(t: typedesc[HttpClientRequestRef], session: HttpSessionRef,
     state: HttpReqRespState.Ready, session: session, meth: meth,
     version: version, flags: flags, headers: HttpTable.init(headers),
     address: ha, bodyFlag: HttpClientBodyFlag.Custom, buffer: @body,
-    headersBuffer: newSeq[byte](max(maxResponseHeadersSize, HttpMaxHeadersSize))
+    headersBuffer: newSeqUninitialized[byte](max(maxResponseHeadersSize, HttpMaxHeadersSize))
   )
   trackCounter(HttpClientRequestTrackerName)
   res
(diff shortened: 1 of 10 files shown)


vladopajic:
also, can we release new minor including this pr?

vladopajic:
> most of these seem to happen during initialization, ie have you measured that these indeed are a source of poor performance?

i'll measure soon, i needed commit for this. :)

i'am mainly betting on change in `BipBuffer` type with this pr.  but, i have changed all possible `seq[byte]` instances whether they may or may not lead to performance improvement - for sake of consistency and as reminder in code to be mindful for performance.

vladopajic:
just measured, i dont see any noticeable improvement. :( but i'll measure again tomorrow. 

fyi, reason why i was even looking in this direction is because currently chronos lib is one of biggest performance bottlenecks in nim-libp2p, as shown by pref report here:

image

arnetheduck:
> currently chronos lib

what you're looking at _most likely_ is a combination of two things:

* when something goes through the `{.async.}` transformation, an implicit copy of all its arguments happens - if you think about it, this is natural because the arguments have to "survive" til the end of the `await` which may happen in a different context - conservatively, a heap-based environment will be allocated into which all data is copied - this environment is part of the closure iterator that is used to implement async
* when "returning" a value from an `async` function, the value is actually _copied_ into the Future _most of the time_ - I think there are structural improvements that can be made here to avoiud some of the copies, specially in newer nim versions that have better "last-use" analysis where automatically, things should be moved instead - it's worth investigating, fixing and reporting to upstream
* similarly, when fetching the value _from_ a `Future`, another copy is _usually_ performed - ie `Future[string]` means that the value stored in the future gets copied on access - this again is related to the (lack of) lifetime analysis for the future instance and the subsequent "defensive" copy that is made - one option would be to move / drain the value from the Future in `await` and leave the Future empty - this however is a bit risky because currently, if you do something like `let fut = xxx(); echo await fut; echo await fut;`, the second call will still return the value.

In all these cases, the way to understand what's going on is to examine the generated C code for a simple function like `proc f(v: string): Future[string] {.async.} = await g(v)` and see where copies are happening. Any copies that are provably excessive should be reported as bugs to upstream - there might be a difference between refc and orc here as well - if there is, that is _also_ a bug.

PR #583, opened 2025-09-01
allFutures: more efficient implementation

arnetheduck:
We only need to consider the unfinished futures


narimiran:
It should be ~~fixed~~ mitigated with https://github.com/status-im/nim-chronos/pull/585

PR #589, opened 2025-09-26
feat: add WaitGroup

vladopajic:
a synchronization primitive similar to `WaitGroup` in go or `CountDownLatch` in java.

useful [here](https://github.com/vacp2p/nim-quic/blob/319577a771192202a3975331876a87e796496b50/tests/quic/testIntegrationUsecase.nim#L46)

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -79,6 +79,12 @@ type
     limit: int
     offset: int
 
+  WaitGroup* = ref object of RootObj
+    ## A synchronization primitive that waits for a collection of 
+    ## asynchronous tasks to finish.
+    count: int
+    fut: Future[void]
+
 proc newAsyncLock*(): AsyncLock =
   ## Creates new asynchronous lock ``AsyncLock``.
   ##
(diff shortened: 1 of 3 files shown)


cheatfate:
Personally i do not see how useful this primitive could be, so probably will ask @arnetheduck opinion.
But there some problem which should be fixed also.

chronos/asyncsync.nim
```diff
@@ -638,3 +644,21 @@ proc waitEvents*[T](ab: AsyncEventQueue[T],
...
+
+proc newWaitGroup*(count: int): WaitGroup =
+  doAssert(count >= 0, "WaitGroup count must be non negative number")
+  let fut = newFuture[void]("WaitGroup")
+  if count == 0:
+    fut.complete()
+  WaitGroup(count: count, fut: fut)
+
+proc wait*(wg: WaitGroup): Future[void] =
+  wg.fut
```
cheatfate, line 656: As an additional comment to this like you should return
```nim
  wg.fut.join()
```
instead of `fut` itself, because oherwise you transferring ownership of internal future which should be managed internally to outer world. So if somebody will cancel some operation which will be involved in `await waitGroup.wait()` this internal future will be cancelled as well, and so become actually finished.

vladopajic: aha, got it! please have a look again

vladopajic:
@arnetheduck could you please give feedback to this pr?

PR #602, opened 2025-11-30
Add `getAsyncStackTrace` API for a shorter async stacktrace

nitely:
Port of https://github.com/nim-lang/Nim/pull/24563

Only differences:

- I tried avoiding strutils hard dependency, 
- Added `getAsyncStackTrace`

It still injects the stack trace into the error msg, but at least that's behind a flag in Chronos, not sure in what case it's needed.


chronos/internal/asyncfutures.nim
```diff
@@ -410,60 +411,73 @@ proc futureContinue*(fut: FutureBase) {.raises: [], gcsafe.} =
...
+    "system.nim",
+    "threadimpl.nim"
+  ]
+  let (filename, procname) = getFilenameProcname(entry)
+  for line in internals:
+    if filename.endsWith line:
+      return true
+  false
+
+proc getAsyncStackTrace*(ex: ref Exception): string =
```
arnetheduck, line 455: hm, what's the use case (outside of tests) for having this public? ie longer term, it feels  like we should be working towards an abstract stack trace and not a string

nitely: API parity with `getStackTrace`. As a end-user you usually want a string. This could be implemented as `$getAsyncStackTraceEntries(ex)` or something else, though.

tests/testasyncstacktrace.nim
```diff
@@ -0,0 +1,224 @@
...
+    await bar()
+
+  proc main() {.async.} =
+    await foo()
+
+  try:
+    waitFor main()
+    doAssert false
+  except ValueError as err:
+    poll()
```
arnetheduck, line 206: poll?

chronos/internal/asyncfutures.nim
arnetheduck: I think what we actually want here might be an api in the upstream repo that sets / filters the raw entry-based stack trace of the exception itself, so that _its_ normal stack trace entry functions work as usual and so that eventually, we don't have to place the trace in the message itself.

In any case, we can probably leave as is for now and deprecate this function when the above is ready.

PR #611, opened 2026-01-07
chore(AsyncSemaphore): expose size as read only

vladopajic:
logos messaging needs to access size of semaphore in two places, having getter for size makes it convenient. 

logos semaphote usage:
https://github.com/logos-messaging/logos-messaging-nim/blob/a4e44dbe05347197ba367f967aa99078814a80fc/waku/node/peer_manager/peer_manager.nim#L751
https://github.com/logos-messaging/logos-messaging-nim/blob/a4e44dbe05347197ba367f967aa99078814a80fc/waku/node/peer_manager/peer_manager.nim#L1053

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -656,6 +656,9 @@ proc newAsyncSemaphore*(size: int = 1): AsyncSemaphore =
     waiters: initDeque[Future[void].Raising([CancelledError])](),
   )
 
+proc size*(s: AsyncSemaphore): int =
+  return s.size
+
 proc availableSlots*(s: AsyncSemaphore): int =
   return s.availableSlots
 
--- a/tests/testasyncsemaphore.nim
+++ b/tests/testasyncsemaphore.nim
@@ -14,20 +14,21 @@ suite "AsyncSemaphore":
   teardown:
     checkLeaks()
 
-  asyncTest "default size":
+  asyncTest "create default size":
     let sema = newAsyncSemaphore()
     check sema.availableSlots == 1
 
-  asyncTest "custom size":
+  asyncTest "create custom size":
     let sema = newAsyncSemaphore(3)
     check sema.availableSlots == 3
 
-  asyncTest "invalid size":
+  asyncTest "create invalid size":
     expect AssertionDefect:
       discard newAsyncSemaphore(0)
 
   asyncTest "should acquire":
     let sema = newAsyncSemaphore(3)
+    check sema.size() == 3
 
     await sema.acquire()
     check sema.availableSlots == 2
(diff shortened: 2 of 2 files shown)


vladopajic:
having that size getter would avoid having redundant properties here (maxConnectionsIn, maxConnectionsOut):
https://github.com/vacp2p/nim-libp2p/pull/2007/files

----

since that size value must be stored in semaphore, why would there ever be a problem having read only getter for that property?

arnetheduck:
> since that size value must be stored in semaphore, why would there ever be a problem having read only getter for that property?

It doesn't, actually - many sem implementations don't and instead structure their code differently to not need it, instead just keeping track of the remaining capacity which tends  to yield a simpler implementation - it just hasn't been important enough to address yet.

arnetheduck:
> size is the key property

Many other semaphore implementations disagree ;) 

What is called "size" in this implementation is in reality the "initial number of slots" available in the semaphore. From there, you can add slots and remove them - when you acquire a slot, you remove one slot from the semaphore and take it for yourself. When you release it, you add one (or more) slot to the semaphore.

Note however that you can, in "text-book" implementations, give back more slots than you took - this simply increases the number of slots available for the taking, even if they exceed the "initial size". The chronos semaphore presently does not allow this, but probably it should, to be "feature-complete" with  respect to a "normal" semaphore.

As a consequence, semaphores typically don't track the "initial number of slots" - instead, they track (and sometimes expose) the "currently available slots" - so if I create a semaphore with 3 slots and you acquire one slot, the semaphore "getValue" function would return 2 - this is different from (initial) "size", which currently stores 3 no matter how many slots have been acquired.

I think this is simply a difference in mental model over what a semaphore is - in posix, the semaphore keeps track only of the "current size". If you're coming from windows, there exists indeed a concept of a "maximum size" but it's separate from "initial size" - so if we're to support the full windows-inspired feature set, we'd have to add more than expose the size.

vladopajic:
> give back more slots than you took - this simply increases the number of slots available for the taking

aha! i was not thinking that this property of semaphore is allowed. okay, now i do understand your reasoning. BUT :)

- i wouldn't really call that a semaphore. or no matter what the terms should be they are different primitives that appear similar. 

- having size() is a feature. to elegantly solve this in `nim-libp2p` we actually need wrapper around semaphore that should expose size. instead of doing this, i did what most users would do, wrap this in place where it is used - it's a solution with a smell. we could have wrapper, but why not create library that helps users?

- if fancy semaphores are needed then cronos could have "newSemaphoreFancy(....)" to fulfill other usecase. if `newSemaphoreFancy(....)` is default (implementation with changing capacity) then people needing classical semaphore properties would have to do what? implement it again?

PR #614, opened 2026-01-21
Expose shutdown api

Ivansete-status:
- New `closeDispatcher` procs in `chronos/internal/asyncengine.nim`.
- ~~Expose simple `shutdown` proc that frees.~~

This has been tested in `status-go` when it loads `nim-sds`.

## Related PR
- https://github.com/logos-messaging/nim-sds/pull/43


chronos/ioselects/ioselectors_poll.nim
```diff
@@ -56,7 +56,7 @@ proc new*(t: typedesc[Selector], T: typedesc): SelectResult[Selector[T]] =
 
 proc close2*[T](s: Selector[T]): SelectResult[void] =
   s.fds.clear()
-  s.pollfds.reset()
+  s.pollfds.setLen(0)
```
arnetheduck:
```suggestion
  s.pollfds.reset()
```

chronos/internal/asyncfutures.nim
```diff
@@ -641,6 +641,17 @@ proc pollFor[F: Future | InternalRaisesFuture](fut: F): F {.raises: [].} =
 
   fut
 
+proc shutdown*(): Result[void, string] {.raises: [].} =
```
arnetheduck:
```suggestion
proc closeThreadDispatcher*(): Opt[string] {.raises: [].} =
```

throughout, `close` needs to be an operation that unconditionally closes dispatcher and optionally returns a diagnostic - this is similar to how `close` in posix closes the file descriptor but _may_  return an error code - even when it returns an error code, the resources are released and the descriptor gets closed.

chronos/internal/asyncfutures.nim
```diff
@@ -641,6 +641,17 @@ proc pollFor[F: Future | InternalRaisesFuture](fut: F): F {.raises: [].} =
...
+  ## Performs the shutdown and cleanup of all dispatcher resources.
+  ## Notice that this should be called only when sure that no new async tasks will be scheduled.
+  ##
+  ## This routine shall be called only after `pollFor` has completed. Upon
+  ## invocation, all streams are assumed to have been closed.
+
+  let disp = getThreadDispatcher()
```
arnetheduck: this needs to use `gDisp` directly and check for null, so that it doesn't _create_ a dispatcher .. it should also reset gDisp, so this code becomes something like `let disp = move(gDisp); if not isNil(disp): ..`
Opt

chronos/internal/asyncfutures.nim
```diff
@@ -641,6 +641,17 @@ proc pollFor[F: Future | InternalRaisesFuture](fut: F): F {.raises: [].} =
 
   fut
 
+proc shutdown*(): Result[void, string] {.raises: [].} =
+  ## Performs the shutdown and cleanup of all dispatcher resources.
+  ## Notice that this should be called only when sure that no new async tasks will be scheduled.
+  ##
+  ## This routine shall be called only after `pollFor` has completed. Upon
```
arnetheduck: there's no (public) `pollFor`.

chronos/internal/asyncfutures.nim
```diff
@@ -641,6 +641,17 @@ proc pollFor[F: Future | InternalRaisesFuture](fut: F): F {.raises: [].} =
...
+  ## invocation, all streams are assumed to have been closed.
```
arnetheduck: There's no such assumption - it's simply the case that it's undefined behavior to close the dispatcher when there are pending futures and sockets open.

The real risk here is that there are queued tasks that are about to be processed that are still pending - therefore, we should `doAssert disp.callbacks.len == 0` etc before closing - on windows in particular, when you close an IOCP, all scheduled overlapped operations are "lost" in the sense that there's no way to free the memory  associated with them - while we do not separately track these, closing the dispatcher when there is "known" work is always wrong so we can raise a defect at least for these cases - it won't be perfect but it should raise some awareness .. similarly, on windows, we should `doAssert GetQueuedCompletionStatusEx(..., zero_timeout) == zero_events` to catch the obviously wrong behavior.

chronos/ioselects/ioselectors_poll.nim
Ivansete-status: Addressed in 224be37207f5727ddf99dc7eed4f2682c2a48f7a

chronos/internal/asyncfutures.nim
Ivansete-status: Addressed in 834e4e760963cfca32f8f5fefd0a12375c393020

chronos/internal/asyncfutures.nim
Ivansete-status: Addressed in fa4b8110d813209a6a4a1575b9319a4dd96eefcf

chronos/internal/asyncfutures.nim
Ivansete-status: Addressed in c61df2ab45185dd339196587a82342e66f6f528f

chronos/internal/asyncengine.nim
```diff
@@ -1229,6 +1295,26 @@ else:
...
+  ##
+  ## Cancelled entries, whose callback has been cleared, and the sentinel of
+  ## the callback queue are not counted - they represent no work. Operations
+  ## that the OS queue has not reported yet are not counted either.
+  var res = 0
+  for callback in disp.callbacks.items():
+    if not(isSentinel(callback)) and not(isNil(callback.function)): inc(res)
+  # `HeapQueue` is indexed rather than iterated, because `items` for it was
+  # only added in later Nim versions.
+  for index in 0 ..< len(disp.timers):
```
arnetheduck: timers, idlers and other pure-nim constructs don't really matter - in fact, we can simply reset them in the dispatcher so that the refcount goes down, on close

what matters on windows is the completion queue, and an posidx the selector - these must be empty since they represent state that will break.

chronos/internal/asyncengine.nim
```diff
@@ -115,6 +115,16 @@ template preparePoll(loop: PDispatcherBase) =
   loop.inEventLoop = true
   defer: loop.inEventLoop = false
 
+proc resetBaseDispatcher(loop: PDispatcherBase) {.gcsafe, raises: [].} =
```
arnetheduck, line 118: just `loop[].reset()` ?

arnetheduck: or does that not work with the atomic?

arnetheduck: also, the thread callbacks need to be manually drained and deallocated one by one

chronos/internal/asyncengine.nim
```diff
@@ -283,6 +294,7 @@ elif defined(windows):
     Dispatcher = object of DispatcherBase
       ioPort: HANDLE
       handles: HashSet[AsyncFD]
+      waitables: HashSet[HANDLE]
```
arnetheduck: let's not add this runtime overhead just for the purpose of an assert

chronos/internal/asyncengine.nim
Ivansete-status: `loop[].reset()` cannot be used because it doesn't compile under `refc` because `MpscQueue` doesn't allow `=sink` (nor `=copy`.)

---

Great catch! 
The manual callback cleanup is addressed in 41420c242f876787dcab4e1d130ffc0ca6b21a2c

chronos/internal/asyncengine.nim
Ivansete-status: Addressed in 006dc86d73307ca776ebaa4fb4f469991a0d3a41

chronos/internal/asyncengine.nim
```diff
@@ -774,6 +793,75 @@ elif defined(windows):
     if not(isNil(aftercb)):
       loop.callbacks.addLast(AsyncCallback(function: aftercb))
 
+  proc pendingEventsCount(loop: PDispatcher): int =
+    ## Number of events carrying work waiting to be dequeued from loop's I/O
```
arnetheduck: this has the side effect of discarding the events so this needs a more descriptive name, ie it's destructive - as such, the assert should probably be inside this function and it should be called something like `assertNoEvents`.

Ivansete-status: Addressed in b26e9193cfd7bc42c8403baa7418b427e69319d2

PR #617, opened 2026-02-09
Docs: HTTP Client Tutorial

moigagoo:


docs/src/tutorials/uptimemon/chapter1.md
```diff
@@ -47,26 +29,25 @@ Now let's see what we're doing here line by line.
...
-let session = HttpSessionRef.new()
+{{#shiftinclude auto:../../../examples/uptimemon/chapter1.nim:4}}
 ```
 
 Let's focus on this line for a moment. Here, we're creating an HTTP session. Why would we do it if we need to make only only request, why can't we just send it? The reason is, Chronos is designed for multitasking and a session is a more natural concept than a singular request in this context. While we're just starting, using a session may feel redundant but since our end goal is to send many requests, it will fit just right.
 
 ```nim
-try:
-  let response = await session.fetch(parseUri(url))
+{{#shiftinclude auto:../../../examples/uptimemon/chapter1.nim:6:7}}
```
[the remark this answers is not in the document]
moigagoo: I'm still experimenting with includes. Tried anchors first, but wince I'm basically including the file like bu line, the file gets butchered anchor comments. Maybe including larger chunks is better...

docs/src/tutorials/http_client/chapter1.md
```diff
@@ -0,0 +1,95 @@
...
+
+Note that catching `HttpError` does not contadict the `raises` value at the function definition: since we handle the exception and not re-raise it, our promise that only `CancelledError` ever emits from `check` is held true. 
+
+```nim
+{{#shiftinclude auto:../../../examples/http_client/chapter1/src/uptimemon.nim:finally}}
+```
+
+Regardless of how successful our check was, we must close the session after we're done with in and return the resources back to your computer. [`closeWait`](/api/chronos/apps/http/httpclient.html#closeWait,HttpSessionRef) is a function that schedules all open connections within this session to be closed.
+
+We added [`noCancel`](/api/chronos/internal/asyncfutures.html#noCancel,F) to make sure the closing procedure is not cancelled with a propagated `CancellationError` from another function. Use `noCancel` in resource-critical operations or atomic operation groups that must either all complete or all fail.
```
arnetheduck: `noCancel` is only needed for functions that are not cancellation-aware - ie `closeWait` does generally not need it since it handles cancellation on its own - the hint is that it doesn't raise `CancelledError`

docs/src/tutorials/http_client/chapter1.md
```diff
@@ -0,0 +1,95 @@
...
+
+Note that this function is annotated with `async` pragma because we won't call it directly but instead will "book" its execution from Chronos in an asynchronous way.
+
+Also note the `raises: [CancelledError]` part. This is Chronos's way of announcing the exceptions that are expected to the raised by this function. This mechanism is called [checked exceptions](../../error_handling.md#checked-exceptions). In this particular case, we tell the compiler that this function has cancellable things inside it and propagates the cancellation to its caller. No other exceptions should leak from it and if they do, it's a defect in the program.
+
+```nim
+{{#shiftinclude auto:../../../examples/http_client/chapter1/src/uptimemon.nim:session}}
+```
+
+Let's focus on this line for a moment. Here, we're creating an HTTP session. Why would we do it if we need to make only only request, why can't we just send it? The reason is, Chronos is designed for multitasking and a session is a more natural concept than a singular request in this context. While we're just starting, using a session may feel redundant but since our end goal is to send many requests, it will fit just right.
```
arnetheduck: Sessions exist to manage long-lived / reuseed connections, more so than "multitasking" - ie multitasking (concurrency is the better word, multitasking is something the OS does) is chronos' domain in general while sessions are more about lifetime management of the connection so that when you make multiple requests, they are assigned to a free connection from a pool, either reusing an existing connection or making a new one depending on what's available. The session manages the pool of connections and I think we can state this as a point calling into question why it has to be done when making a single request - ie when presented as a connection manager, we don't have to make it a question here.

docs/src/tutorials/http_client/chapter2.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Goal:** Learn how to make arbitrarily many HTTP requests asynchronously.
+
+**Source code:** [chapter2/src/uptimemon.nim](https://github.com/status-im/nim-chronos/blob/master/docs/examples/http_client/chapter2/src/uptimemon.nim)
+
+OK, we have a working app that can check one URI at a time, which is not that much impressive. Let's update our project to do what Chronos was made for—concurrency!
+
+We'll take a somewhat unusual approach and **start with the wrong solution** before revealing the proper way of solving this problem. By highlighting the common mistakes, we'll help you avoid them in the future.
+
+## Wrong Solution: Naive Loop
```
arnetheduck: I would not call this "wrong" - instead I would use this section to introduce the point that when making multiple requests, you _reuse_ the session but create a new request every time - this naturally answers the question of how the session lifetime differs from that of a request.

There are many valid reasons to issue requests serially, for example to limit resource usage or because one request depends on the other - it's important for the user to understand that even in the case of serial requests, we still typically want to reuse the session for example so that we only make one connection to the server instead of two, when fetching two pages.

docs/src/tutorials/http_client/chapter2.md
```diff
@@ -0,0 +1,115 @@
...
+    waitFor check(uri)
+```
+
+If you run this code, you'll see that it works and does in fact check your URIs.
+
+So, why is it the wrong solution? Because we check URIs in a blocking, synchronous, and therefore slow loop.
+
+With this kind of solution, your app checks URIs one by one and the total time is the sum of the time spent getting a response for every single URI. This is hardly usable if you have as few as 20 URIs to check.
+
+## Correct Solution: Asynchronous Bulk Requests
```
arnetheduck: I would title this "Adding concurrency" or something like that ("correct" depends on the situation) and probably make it a separate page - it's an important distinction from the other http-client-related concepts (connection reuse, pipelining, chunked transfers etc). "bulk" usually means something else, but above all, it's not part of the http nomenclature (the closest thing is streaming or chunked transfers).

docs/examples/http_client/chapter2/src/uptimemon.nim
```diff
@@ -0,0 +1,39 @@
...
+    else:
+      echo "[NOK] " & uri & ": " & $response.status
+  except HttpError:
+    echo "[ERR] " & uri & ": " & getCurrentExceptionMsg()
+# ANCHOR_END: check_uri
+
+# ANCHOR: check_uris
+proc check(uris: seq[string]) {.async: (raises: [CancelledError]).} =
+  let session = HttpSessionRef.new()
+  var futures: seq[Future[void]]
```
arnetheduck: We typically write this as `uris.mapIt(session.check(it))` - if you want to be explicit about future types, they should include raising types: `seq[Future[void].Raising([CancelledError]])`.

In fact, the reason this code compiles is because it ignores exceptions raised by `session.check(..)` - `allFutures` only checks that all futures finished but it does not itself raise an error if any future fails. To find out if any exceptions were raised, you'd have to `await`/`read` each future individually after `allFutures` and if you were to do this with `for f in futures: await f`, the code would no longer compile.

There's another point of nuance which is not aimed at this chapter, but if you were to cancel `allFutures`, the requests would not be cancelled - to cancel requests, one has to do so explicitly, ie `try: await allfutures(..) except CancelledError: await futures.cancelAndWait()` -

The older form of cancelling a seq of futures is `await noCancel allFutures(futures.mapIt(it.cancelAndWait)))` - this is where `noCancel` becomes useful: unlike `closeWait`, `allFutures` responds to cancellation so we need to block cancellations of `allFutures` itself. An easy rule of thumb is that `noCancel` is always applied except to `closeWait` (or other explicitly documented exceptions). Nowadats we have a varags cancelAndWait so it's no longer needed.

docs/src/tutorials/http_client/chapter3.md
```diff
@@ -0,0 +1,40 @@
...
+```nim
+const uris = @[
+  "https://duckduckgo.com/?q=chronos", "https://mock.codes/403", "http://123.456.78.90",
+  "http://10.255.255.1",
+]
+```
+
+Run the program and you'll see that it'll run for 10+ seconds, stuck on this last IP.
+
+Let's add a timeout to our requests to cancel slow requests before they ruin our app: if a request takes longer than 5 seconds, we cancel it.
```
arnetheduck: we can just show the function that changed here, no need to repeat the infrastructure surrounding it

docs/src/tutorials/http_client/chapter1.md
```diff
@@ -0,0 +1,95 @@
...
+{{#shiftinclude auto:../../../examples/http_client/chapter1/src/uptimemon.nim:status}}
+```
+
+Once we've received our response, we can check its status. If it's 200, we mark this URL healthy (later in the tutorial, we'll improve this logic to handle empty and junk responses), otherwise—not healthy.
+
+```nim
+{{#shiftinclude auto:../../../examples/http_client/chapter1/src/uptimemon.nim:except}}
+```
+
+If the request fails (e.g. the connection is unstable or the host is unreachable), `fetch` would raise a `HttpError` exception. Since raising this exception is part of our business logic, we catch it and report the error with `getCurrentExceptionMsg`.
```
arnetheduck: It might be worthwhile to add an info callout here about "http status errors" and "communication errors" here - ie in the tutorial we call `40x` status codes "errors" as well so there's room for confusion over what consitutes a "success" and how a "successfult" fetch might result in an "error" status.

docs/examples/http_client/chapter3/src/uptimemon.nim
```diff
@@ -0,0 +1,39 @@
...
+
+    if await responseFuture.withTimeout(5.seconds):
+      let response = responseFuture.read()
+
+      if response.status == 200:
+        echo "[OK] " & uri
+      else:
+        echo "[NOK] " & uri & ": " & $response.status
+    else:
+      raise newException(AsyncTimeoutError, "Connection timed out")
```
arnetheduck: `AsyncTimeoutError` is raised by `wait`:
```nim
try:
  let ... = session.fetch(...).wait(5.seconds)
except AsyncTimeoutError:
 ...
```

no need to raise it manually, ie `withTimeout` is used when you _don't_ want exceptions.

docs/src/tutorials/http_client/chapter4.md
```diff
@@ -0,0 +1,174 @@
...
+[OK] https://duckduckgo.com/?q=chronos
+[ERR] http://10.255.255.1: Connection timed out
+[ERR] https://html.spec.whatwg.org/: Connection timed out
+```
+
+Let's optimize our check to handle large page like this one.
+
+## Getting Just the Status
+
+First, let's not download the page and just check the response status.
```
arnetheduck: it might be worth inserting an info box about how the http protocol divides each request and response into a "header" and a "body" where the header contains metadata like the status code while the body contains the actual content - this is true of both successes and error statuses.

docs/examples/http_client/chapter5/src/uptimemon.nim
```diff
@@ -0,0 +1,112 @@
...
+    result = "<html" in bytesToString(fetchedBytes)
+
+HttpError ANCHOR: check
+proc check(session: HttpSessionRef, uri: string) {.async: (raises: [CancelledError]).} =
+  try:
+    let request = HttpClientRequestRef.new(session, uri)
+
+    if request.isErr:
+      raise newException(HttpRequestError, request.error)
+
```
arnetheduck:
```suggestion
    let request = HttpClientRequestRef.new(session, uri).valueOr:
      raise newException(HttpRequestError, request.error)
```

arnetheduck: but even better, can use `parseUri` as in the other examples and then `new` always works without error, simplifying the code. the only time it fails is if the uri is invalid

docs/examples/http_client/chapter5/src/uptimemon.nim
```diff
@@ -0,0 +1,112 @@
...
+    buffer = newSeq[byte](1024)
+    fetchedBytes: seq[byte]
+
+  while not result and len(fetchedBytes) <= 10 * 1024:
+    let bytesRead = await bodyReader.readOnce(addr(buffer[0]), len(buffer))
+
+    if bytesRead == 0:
+      break
+
+    fetchedBytes &= buffer
```
arnetheduck: only `bytesRead` bytes of `buffer` are valid, not all of them .. this is a classic example of why beginners might prefer something like `let buffer = reader.read(1024)` which fetches 1024 bytes at a time but avoids such easy-to-make mistakes (it's not optimal, efficiency-wise, and might wait for slightly longer than readOnce, but good enough for small buffers where 1024 is unlikely to make any practical difference - readOnce should probably be left for a specific efficiency chapter)

docs/examples/http_client/chapter5/src/uptimemon.nim
```diff
@@ -0,0 +1,112 @@
...
+
+    result = "<html" in bytesToString(fetchedBytes)
```
arnetheduck: this is not incorrect - but it also rechecks all bytes all the time which is not a great thing to encourage - what you're looking for is an expression that only saves the last N-1 bytes in fetchedBytes where N is the length of the thing you're looking for, or similar partial-match strategies

arnetheduck: also, if we wanted a string and were using `readOnce`, we'd use a `string` for the buffer to avoid `bytesToString`

docs/examples/http_client/chapter6/src/uptimemon.nim
```diff
@@ -0,0 +1,135 @@
...
+      "https://ntfy.sh/" & ntfyTopic,
+      meth = MethodPost,
+      headers = headers,
+      body = body,
+    )
+
+  if request.isOk:
+    try:
+      let response = await request.get.send()
+      await response.closeWait()
```
arnetheduck: the request also needs closing, in `finally`

arnetheduck (commented):
here's a first batch of review comments - some of them may apply to the server tutorial as well since they touch upon general principles of use

docs/src/tutorials/http_client/chapter1.md
moigagoo: Removed `noCancel` here and everywhere in similar places and updated "Core concepts."

docs/src/tutorials/http_client/chapter1.md
moigagoo: Rephrased this bit to emphasize connection management instead of concurrency management.

docs/src/tutorials/http_client/chapter2.md
moigagoo: Fixed this part. Instead of calling things wrong and correct, I'm calling them serial and concurrent.

docs/src/tutorials/http_client/chapter2.md
moigagoo: I've renamed the section to simply "Concurrent requests."

I haven't moved it to a separate page though. Even though it is not universally the right answer, I still think that making concurrent requests is more natural in the context of Chronos and this tutorial in particular. "Serial Requests" is here only for additional context but the meat of the page is in "Concurrent Requests."

Maybe "Serial Requests" can be reduced to a paragraph so that that page focus is clearly on the concurrency.

docs/examples/http_client/chapter2/src/uptimemon.nim
moigagoo: Thanks!

I've updated the proc in all examples:
1. Cancellations are now handled internally, CancelledError is no longer raised.
2. We're not interested in checking exceptions after `allFutures` because we handle individual exceptions in check[uri] so only the cancellation logic is added to check[uris].

docs/src/tutorials/http_client/chapter4.md
moigagoo: Done.

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: There is no `new` variant that accepts `Uri` so we can't use `parseUri` as is (i.e. `HttpClientRequestRef.new(session, parseUri(uri))` won`t work).

There is a flavor of `new` that accepts `HttpAddress` but if we want to use that, we'll have to generate the address first with `getAddress(session, uri)` and then pass it to `new`. The code will be longer and we'll still have to use the same `valueOr`.

I can add a new `new` flavor that accepts `Uri`:

```nim
proc new*(t: typedesc[HttpClientRequestRef], session: HttpSessionRef,
          uri: Uri, meth: HttpMethod = MethodGet,
          version: HttpVersion = HttpVersion11,
          flags: set[HttpClientRequestFlag] = {},
          maxResponseHeadersSize: int = HttpMaxHeadersSize,
          headers: openArray[HttpHeaderTuple] = [],
          body: openArray[byte] = []): HttpResult[HttpClientRequestRef] =
  let address = ? session.getAddress(uri)
  let res = HttpClientRequestRef(
    state: HttpReqRespState.Ready, session: session, meth: meth,
    version: version, flags: flags, headers: HttpTable.init(headers),
    address: address, bodyFlag: HttpClientBodyFlag.Custom, buffer: @body,
    headersBuffer: newSeq[byte](max(maxResponseHeadersSize, HttpMaxHeadersSize))
  )
  trackCounter(HttpClientRequestTrackerName)
  ok(res)
```

But that's pretty much identical to the `new` flavor that accepts a string, we're just moving `parseUri` a level higher.

Do you think it would be the right thing to do?

I'll use `valueOr` as you suggested.

moigagoo:
Fixed an important problem with the code. Since I worked on the tutorial on Windows, I didn't notice it but now I'm on Linux and the issue is very much visible.

I used to resolve addresses as part of the URI check. This is wrong because address resolution is a blocking call and no matter if the check is async, when it hits `getAddress`, the whole program is blocked. And since some requests are created before the corrupted one, they would time out simply because they're blocked.

On Windows, invalid URLs are rejected immediately so the blocking happens faster than I could detect. On Linux, address resolution takes several seconds even for invalid URLs.

I've moved address resolution to a separate prep step.

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: Thanks, I've replaced the `readOnce` bit with a simple `read`.

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: > this is not incorrect - but it also rechecks all bytes all the time which is not a great thing to encourage - what you're looking for is an expression that only saves the last N-1 bytes in fetchedBytes where N is the length of the thing you're looking for, or similar partial-match strategies

We can't check just the last buffer because the marker could be split between two buffer reads (i.e. `<ht` is at the end of one read and `ml` is at the start of the next one). So I decided to create a moving window of length len(buffer) + len(marker) - 1. Then I shift the window on every buffer update and add the new buffer.

moigagoo: > also, if we wanted a string and were using readOnce, we'd use a string for the buffer to avoid bytesToString

I couldn't find a way to read directly to a string. We have `readLine` but that's no good at all: some sites return everything as a single line, which is way larger that the buffer we're happy to read.

I'd be happy to avoid the bytes though, I just couldn't find a way to.

docs/examples/http_client/chapter6/src/uptimemon.nim
moigagoo: Done.

.github/workflows/doc.yml
```diff
@@ -24,14 +24,15 @@ jobs:
 
       - name: Install mdBook and preprocessors
         run: |
-          cargo binstall mdbook@0.4.36 \
+          cargo binstall mdbook@0.4.51 \
```
arnetheduck:
```suggestion
          cargo binstall mdbook@0.4.52 \
```

.github/workflows/doc.yml
```diff
@@ -24,14 +24,15 @@ jobs:
...
             mdbook-toc@0.14.1 \
             mdbook-open-on-gh@2.4.3 \
-            mdbook-admonish@1.14.0
+            mdbook-admonish@1.14.0 \
```
arnetheduck:
```suggestion
            mdbook-admonish@1.20.0 \
```

.github/workflows/doc.yml
```diff
@@ -24,14 +24,15 @@ jobs:
 
       - name: Install mdBook and preprocessors
         run: |
-          cargo binstall mdbook@0.4.36 \
+          cargo binstall mdbook@0.4.51 \
             mdbook-toc@0.14.1 \
```
arnetheduck:
```suggestion
            mdbook-toc@0.14.2 \
```

docs/examples/http_client/chapter6/src/uptimemon.nim
```diff
@@ -0,0 +1,159 @@
...
+    result = marker in bytesToString(window)
+
+# ANCHOR: semaphore
+proc check(
+    session: HttpSessionRef, address: HttpAddress, semaphore: AsyncSemaphore
+) {.async: (raises: [CancelledError, AsyncSemaphoreError]).} =
+  await acquire(semaphore)
+
+  defer:
+    release(semaphore)
```
arnetheduck: better catch the AsyncSemaphoreError here, no point leaking it

docs/src/tutorials/http_client/chapter2.md
arnetheduck: > Maybe "Serial Requests" can be reduced to a paragraph so that that page focus is clearly on the concurrency.

Well, as-is, the page misses the opportunity to introduce _session reuse_ separately from concurrency - irrespective of whether you're making concurrent requests or not, you should be creating a single session for multiple requests - the "serial requests" part of the page is an excellent place to introduce this idea so that the reader can focus their understanding on this aspect alone before moving on to concurrency. Just adding a list and a loop is kind of poor use of attention since it doesn't introduce any chronos-specific detail to the section, hence the suggestion to use the section for talking about the session and what its role is.

docs/examples/http_client/chapter5/src/uptimemon.nim
arnetheduck:
```nim
var buffer = newString(1024)
buffer.setLen(await readOnce(addr buffer[0], buffer.len))
```

something like this works but it's also a bit uglier. not sure if it's worth the tradeoff in the tutorial at least.

docs/examples/http_client/chapter3/src/uptimemon.nim
arnetheduck: ping

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: I've rewritten this bit to use strings instead of arrays of bytes. The logic is also now simpler: we don't need a moving window, we just resize the string.

I've used your code just split it into two lines for readability.

If there was a flavor of readOnce that would take a string, that would be so cool. I've added one as part of this PR. IMO the code looks much nicer with it. https://github.com/status-im/nim-chronos/pull/617/changes#diff-e3580934a92d887f58d92d02e3e4771c977dbdb0bfe6298b39d9d80348ffae4aR405-R410

docs/examples/http_client/chapter3/src/uptimemon.nim
moigagoo: Thanks for the ping, I overlooked this comment. Replaces withTimeout with wait, the code became cleaner.

docs/examples/http_client/chapter6/src/uptimemon.nim
moigagoo: Done.

moigagoo:
@arnetheduck I've rewritten some chapters significantly and added a couple more.

arnetheduck:
> resolve addresses

#656 moves name resolution to the connect attempt which indeed may block - what's needed here is to rewrite the resolver in chronos to use [async](https://ziglang.org/devlog/2025/#2025-10-15) versions of name resolution - this is an interesting problem to solve but it's somewhat messy to do well - ie there are more or less 3 ways to approach it:

* platform-specific code, fall back on sync for random platforms - this is probably the best starting point (like the zig example above)
* thread pool - messy lifetime issues
* wrap c-ares or similar

Potentially, this could be an extension point where the dispatcher could have an extension mechanism for custom resolvers.

docs/src/tutorials/http_client/chapter1.md
```diff
@@ -0,0 +1,95 @@
...
+
+Note that catching `HttpError` does not contadict the `raises` value at the function definition: since we handle the exception and not re-raise it, our promise that only `CancelledError` ever emits from `check` is held true.
```
arnetheduck:
```suggestion
Note that catching `HttpError` does not contradict the `raises` value at the function definition: since we handle the exception and not re-raise it, our promise that only `CancelledError` ever emits from `check` is held true.
```

docs/src/tutorials/http_client/chapter5.md
```diff
@@ -0,0 +1,25 @@
...
+
+Because this happens *inside* our async `check` function *before* we even get a `Future` to `await`, it blocks the Chronos event loop entirely. To fix this, we must separate the blocking address resolution from our concurrent async loop by resolving all URIs into a list of `HttpAddress` objects **before** we start any async operations.
+
+Here's the improved version:
+
+```nim
+{{#shiftinclude auto:../../../examples/http_client/chapter5/src/uptimemon.nim:all}}
+```
+
+The key is the `resolveUris` helper function, which uses [`session.getAddress(uri)`](/api/chronos/apps/http/httpclient.html#getAddress,HttpSessionRef,string) to perform the blocking DNS lookup *sequentially* before starting the async loop.
```
arnetheduck: this function is deprecated as of #656 - in fact, this page can probably be removed for now and replaced with a warning box in the previous page - the longer-term solution here is that chronos needs to implement async name resolution.

the advice to "pre-resolve" is also problematic since DNS records change - it's usually not correct to resolve only once at the startup of the application, specially in a monitoring application. For now, highlighting that it might happen is sufficient and we'll come back to the problem when chronos has an async resolver

docs/examples/http_client/chapter6/src/uptimemon.nim
```diff
@@ -0,0 +1,100 @@
...
+
+# ANCHOR: findMarker
+proc findMarker(
+    response: HttpClientResponseRef
+): Future[bool] {.
+    async: (raises: [HttpUseClosedError, AsyncStreamError, CancelledError])
+.} =
+# ANCHOR_END: findMarker
+# ANCHOR: bodyReader
+  let bodyReader = response.getBodyReader()
```
arnetheduck: like everything else, the body reader must be closed.

an important aside here is also that if you don't read the whole body (either via `readOnce`/`read` or `consume`) the connection will effectively be broken and cannot be reused (since the server is still sending bytes on the same tcp connection). for a monitoring app this is fine, but it's something that should be called out.

moigagoo: Done. I wonder though, how does one determine what needs to be closed? Open connections are closed, that makes sense. But that a reader needs closing is not so obvious.

arnetheduck: >  But that a reader needs closing is not so obvious.

pretty much everything in chronos needs closing since we're dealing with OS resources (sockets) - a reader is a stream -> streams need closing

a reader needs closing for example because that signals to the runtime that reading is done that no more reading will be done using that particular reader - when you call `finish` on the response (also missing, now that I look at the code), chronos can know that it's safe to reuse the connection for the next request - or that it should be disconnected.

moigagoo: > an important aside here is also that if you don't read the whole body (either via readOnce/read or consume) the connection will effectively be broken and cannot be reused (since the server is still sending bytes on the same tcp connection). for a monitoring app this is fine, but it's something that should be called out.

I've added [request closing](https://github.com/status-im/nim-chronos/pull/617/commits/396bafcc9772163958c03031bf57abb6d5e1e587) in `finally`, will that be a more hygienic way to finish working with the request and free the connection properly?

arnetheduck: > will that be a more hygienic way to finish working with the request and free the connection properly?

It depends on what you want to achieve. When you use a body reader, it reads as many bytes as you asked for but the server will keep sending. Consider what happens when you ask for a 100 bytes file and only "read" 50 bytes - 50 bytes will still be queued on the socket and waiting there for someone to read them. If you were to try to reuse the connection at this point, there would be 50 bogus bytes belonging to the previous request, so the best thing chronos can do at that point is close the connection and not reuse it.

The body reader, although it _can_ be used for partial downloads, is mainly meant for streaming "whole" requests, not just for reading parts - for reading parts, http 1.1 doesn't really have a good mechanism except https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests which is useful mostly for static content.

When you don't read all the bytes of the body, the best thing chronos can do is close the connection and open a new connection for the next request.

If you end up read all bytes using the body reader, you should also call `finish(response)`. This is mirrored for the `request` as well, ie when you're making a request with a big body (such as when uploading a file using POST), you stream the body then call `finish(request)` to get the response.

So a fully streamed request/response looks something like:
```nim
let 
  req = HttpClientRequestRef.new(...)
  resp = 
    try:
      let writer = req.open()
      try:
        await writer.write(...)
      finally:
        await writer.closeWait()
      await req.finish() # On happy path only
    finally:
      await req.closeWait() # Always close

try:
  let reader = resp.getBodyReader()
  try: 
    let data = reader.read()
    ...
    await resp.finish() # On happy path only
  finally:
    await reader.closeWait() # Always close
finally:
  await resp.closeWait() # Always close
```

arnetheduck: I guess this brings us to another point: for an uptime monitor, instead of reading "parts of" the body, the normal thing to do would be to use a HEAD request (which gives the headers without the content) - doesn't really matter for this PR, but a future update to the tutorial could use a HEAD for the discovery but then you'd have to figure out a way to demonstrate body readers .. maybe by streaming some "result" page to a log file, where the entire body gets read.

docs/examples/http_client/chapter8/src/uptimemon.nim
```diff
@@ -0,0 +1,160 @@
...
+
+      if markerFound:
+        echo "[OK] " & address.hostname & address.path
+      else:
+        let message = "[NOK] " & address.hostname & address.path & ": Not valid HTML"
+        echo message
+        await session.sendAlert(message)
+    else:
+      let message = "[NOK] " & address.hostname & address.path & ": " & $response.status
+      echo message
```
arnetheduck, line 117:
```suggestion
      echo &"[NOK] {address.hostname}{address.path}: {response.status}
```

`strformat` can help make these less verbose

moigagoo: I think introducing new import and new syntax here would bring more confusion that benefit.

moigagoo: Especially the `&` syntax is weird for the uninitialized. `fmt` is better but still is something new and unrelated to the topic.

docs/examples/http_client/chapter8/src/uptimemon.nim
```diff
@@ -0,0 +1,160 @@
...
+        await allFutures(futures)
+      except CancelledError:
+        await cancelAndWait(futures)
+        break
+# ANCHOR_END: pass_semaphore
+
+# ANCHOR: sleep
+      echo "Done. Next check in 10 seconds."
+      try:
+        await sleepAsync(10.seconds)
```
arnetheduck, line 152: a sleep like this is usually a good time to bring up drift - ie the uptime monitor, as written, will not check once per 10 seconds - instead, it will check once per `10s + check time + timer delay` where timer delay is the time it takes for the system to wake up and "resume" execution .. if the runtime is busy / blocked, this can take some time.

A drift-free version of this is to do something like (exact logic needs checking):
```nim
var next = Moment.now() + 10.seconds
while true:
  await process()
  let now = Moment.now()
  await sleepAsync(max(now, next) - now)
  next = max(now, next + 10.seconds)
```

docs/examples/http_client/chapter5/src/uptimemon.nim
arnetheduck: hm, right, `parseUri` doesn't do any error checking (at all) .. 

so what we're looking for here is `getHttpAddress(mystring).valueOr: ...` - then use `new` that uses the HttpAddress and never returns an error: https://github.com/status-im/nim-chronos/blob/1f2151316aa764024b616d9f1b2a7f6a54ff63b3/chronos/apps/http/httpclient.nim#L1025

HttpAddress is "approximately" the same as Uri except it's been specialized for the http use case and is more strict (because the http standard is more strict).

docs/examples/http_client/chapter5/src/uptimemon.nim
arnetheduck: this is a limitation in nim - `async` functions can't use `var` parameters so we have to use raw pointers or return values.

arnetheduck: `ref string` could potentially work but that's pretty unusual

arnetheduck (approved):
LGTM as soon as the final touchups are done, solid base for continuing the docs work!

docs/examples/http_client/chapter8/src/uptimemon.nim
moigagoo: Since the chapter about preemptive DNS resolution is no more more and we're back to `check`ing raw strings instead of `HttpAddress`es, these bits are now less noisy.

docs/examples/http_client/chapter8/src/uptimemon.nim
moigagoo: AFAIU this code won't guarantee that the next check happens exactly 10 seconds after the previous one; in fact, nothing can. It just guarantees that we won't wait for _another_ 10 seconds if the scheduled time has already passed.

IMO it isn't critical if the check runs a little later. I've added a note about drift to chapter 7.

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: Oh bummer. The code looked so much better with the string-accepting `readOnce` variant 😅

docs/examples/http_client/chapter5/src/uptimemon.nim
moigagoo: I used to use this one when I used `HttpAddress` instead of string, which was done as part of the preemptive DNS resolution chapter, which was removed.

Since we're back to handling strings, I have this at the start of `check`:

```nim
proc check(session: HttpSessionRef, uri: string) {.async: (raises: [CancelledError]).} =
  let request = HttpClientRequestRef.new(session, uri).valueOr:
    echo "[ERR] " & uri & ": " & error
    return
  ...
```

I like that it is readable and the intention is clear: create a request; if that fails, print and return.

docs/examples/http_client/chapter6/src/uptimemon.nim
moigagoo: I rewrote the parts with request and response initialization and the `findMaker` proc.

Request, response, and body reader are now closed after use and response us `finish`ed if we managed to read the full response (i.e. `len(buffer) == 0`).

IMO the `request` and `response` creation now look very good. To me, the best sign is when I can read the code top to bottom as prose.

I noticed that I can't specify exception in `except` within `try` block in `let`: https://github.com/status-im/nim-chronos/pull/617/changes/3946f0daf0883d3e5b037fd92cc1070ccd92131c#diff-afdd5453693afd6c94de6c3761053aab468f293a392e91f67d2b596fa13f97b5R71

If I replace this line with `except HttpError, AsyncTimeoutError` (the two exception that can actually be raise here), the code won't compile:

```
      Info: using /home/moigagoo/.nimble/pkgs2/nim-2.2.10-17ec440fdb89f8903db29a17898af590087d2b64/bin/nim for compilation
   Building uptimemon/uptimemon using c backend
 Nim Output assertions.nim(38)       raiseAssert
        ... Error: unhandled exception: closureiters.nim(335, 17) `c[i].kind == nkType`  [AssertionDefect]
       Tip: 6 messages have been suppressed, use --verbose to show them.
vnext.nim(1128)          buildFromDir

    Error:  Build failed for the package: uptimemon
```

This looks like a bug but maybe I'm missing something?

arnetheduck: > This looks like a bug

Indeed, it might be worth extracting into a smaller reproducible sample.

That said, `except:` would also catch CancelledError which would change cancellation semantics which may or may not be desired.

docs/examples/http_client/chapter6/src/uptimemon.nim
```diff
@@ -0,0 +1,134 @@
...
+    totalRead = 0
+    buffer = newString(bufferSize)
+    sample = newString(len(marker) - 1)
+
+  while not result and totalRead <= 10 * 1024:
+    let bytesRead = await bodyReader.readOnce(addr buffer[0], len(buffer))
+    buffer.setLen(bytesRead)
+
+    if len(buffer) == 0:
+      await bodyReader.closeWait()
```
arnetheduck: the `bodyReader` is closed again by the `defer` - it's worth considering the order here, where ideally, the body reader should be closed only once and always _before_ `finish`.

arnetheduck: refactoring this to `findMarker(stream: AsyncStream)` is one option, ie move the body reader / finish stuff outside .. and in that case, `readMessage` might be appropriate, it would look something like:
```nim
proc findMarker(stream: AsyncStreamReader) {.async: ... .}=
  var sample = newString(len(marker) - 1)
  var totalRead = 0
  var found = false
  proc match(data: openArray[byte]): (int, bool) =
    if len(data) == 0: 
      (0, false)
    else:
      sample = sample[...
      found = marker in sample
      totalRead += data.len
      (data.len, not found and totalRead <= ...)
  await stream.readMessage(match)
  found

...
let bodyReader = ...getBodyReader()
let found = try:
   await bodyReader.findMarker()
finally:
   await bodyReader.closeWait()
...
```

docs/examples/http_client/chapter6/src/uptimemon.nim
moigagoo: Reported: https://github.com/status-im/nim-chronos/issues/659

Catching `CancelledError` is undesired here, I want it to bubble. So I applied a workaround: I'm catching one exception at a time.

docs/examples/http_client/chapter5/src/uptimemon.nim
```diff
@@ -68,7 +68,10 @@ proc check(session: HttpSessionRef, uri: string) {.async: (raises: [CancelledErr
     response =
       try:
         await request.send().wait(5.seconds)
-      except:
+      except HttpError:
+        echo "[ERR] " & uri & ": " & getCurrentExceptionMsg()
```
arnetheduck: fwiw, we _tend to_ prefer `HttpError as exc: exc.msg` for getting the message - `getCurrentExceptionMsg` has been known to fail / be buggy on many nim versions

moigagoo: My God this looks [horrible](https://github.com/status-im/nim-chronos/pull/617/commits/23b2578ceea92887b392b478726db53513031b31) 😭 `getCurrentExceptionMsg` is so cool, it's a shame we can't use it.

Especially since the fix for the bug with multiple exceptions in one `except` block is not released yet, the code is full of repeating `except`s now, it's just painful to look at.

Maybe we could make the code samples depend on nim#head and in the future update it to use the then latest released version? I feel bad leaving code like that to future docs readers:

```nim
  try:
    if response.status == 200:
      let
        bodyReader = response.getBodyReader()
        markerFound =
          try:
            await bodyReader.findMarker()
          finally:
            await bodyReader.closeWait()

      if markerFound:
        echo "[OK] " & uri
      else:
        let message = "[NOK] " & uri & ": Not valid HTML"
        echo message
        await session.sendAlert(message)
    else:
      let message = "[NOK] " & uri & ": " & $response.status
      echo message
      await session.sendAlert(message)
# This looks so bad...
  except HttpError as e:
    let message = "[ERR] " & uri & ": " & e.msg
    echo message
    await session.sendAlert(message, 4)
  except AsyncStreamError as e:
    let message = "[ERR] " & uri & ": " & e.msg
    echo message
    await session.sendAlert(message, 4)
# up until this point
  finally:
    await response.closeWait()

```

arnetheduck: > getCurrentExceptionMsg is so cool

It's not that you can't use it - you just have to understand the consequences.

For example, in a finally block, its value depends on whether you're exiting the try block from an exception or the happy case - it can be hard to reason about in that sense, because its value fundamentally depends on an implicit global that the compiler generates. Whether you prefer hidden globals or explicitly named variables is a matter of taste, I guess, though being explicit _tends to_ aid understanding when facing problems, at the expense of being more verbose.

The bugs that we've encountered were of the kind where its value was not well-defined in nested try/except/finally blocks precisely for this reason: it's hard to reason about hidden globals because you don't see where and how they are created and updated - using `getCurrentExceptionMsg` makes simple cases more simple and hard cases harder.

arnetheduck: > # This looks so bad...

https://github.com/nim-lang/Nim/issues/25854 - another reason we _tend to_ avoid it is because it loses the type information of the exception - if you re-raise `getCurrentException`, the `raises` effects are wrong - using an explicitly named variable preserves the type.

Used judiciously, `getCurrentExceptionMsg` has its uses, ie in the early chapters of the tutorial where error handling is .. basic .. its use was fine and there's no need to change them - however, as code gets more complex and you want to reraise the same exception after doing some processing or when exception handling becomes nested so there are multple "current" exceptions in-flight, it becomes more hairy and then we switch to the explicit form.

moigagoo: > Especially since the fix for the bug with multiple exceptions in one except block is not released yet, the code is full of repeating excepts now, it's just painful to look at.

I just realized Nim doesn't have `except Exc1, Exc2 as e` syntax. So the only way to handle multiple exception with one line is using `getCurrentExceptionMsg`.

> Used judiciously, getCurrentExceptionMsg has its uses, ie in the early chapters of the tutorial where error handling is .. basic .. its use was fine and there's no need to change them - however, as code gets more complex and you want to reraise the same exception after doing some processing or when exception handling becomes nested so there are multple "current" exceptions in-flight, it becomes more hairy and then we switch to the explicit form.

I don't want to split the focus in the tutorial between the core topic (HTTP client with Chronos) and Nim code patterns and style. So I'd rather use one exception handling style so that the reader wouldn't have to change gears mid-reading.

> fwiw, we tend to prefer HttpError as exc: exc.msg for getting the message - getCurrentExceptionMsg has been known to fail / be buggy on many nim versions

I just realized something. By "we" you meant IFT, the IFT Nim style guide, and the IFT best practices. But I wonder, how relevant is that to this tutorial? IFT style guide is not a universal guide. It is very good and well-thought and it has some strong arguments, but it is not something every Nim programmer must follow. Should we really make it point to enforce it while telling about Chronos?

After thinking about it a bit, I came to the opinion that in this particular case adding a `nim#head` dependency[1] and using `except Exc1, Exc2: ... getCurrentExceptionMsg()` pattern makes for shorter and more readable code. Yes, this code is _potentially_ more error-prone, especially if the reader decides to expand the project. But realistically, they won't. A tutorial ends with the last chapter, that's it.

What do you think?

---

[1] This will be replaced with `require nim >= 2.4.0` when the new version is released.

arnetheduck: > But I wonder, how relevant is that to this tutorial? IFT style guide is not a universal guide.

`chronos` itself follows the guide more or less - style guides provide the most benefit when applied consistently, so when you're reading the tutorial and looking at chronos' code, there should ideally be recognition and sympathy - ie the various points in the style guide complement each other and makes nim more coherent with itself, where there are multiple ways to achieve the same thing. Note though that throughout, and also in the style guide, we consider it a recommendation, not a requirement and sometimes we choose other styles (if we can motivate them).

Regarding exception styles in particular, the reason `getCurrentExceptionMsg` works is because it hides the type of the exception - the reason `except ValueError, OSError as e` doesn't work is because then you can't assign the right type to `e` - that's why generics are needed (as in the issue I linked).

Since we tend to use explicit `raises` types, we also end up needing the type of the exception more often - in this tutorial, when all you do is `echo` the error, the type obviously doesn't matter so `getCurrentExceptionMsg` is fine, but with actual logic handling the error, the type becomes more important and the user would have to "switch" to the explicit `as e` syntax anyway. All in all, you're mostly delaying the inevitable (which is fine, sometimes, depending on what you want to focus on).

moigagoo: Thanks for the comment. To keep the reader's focus on the subject, I've decided to use `getCurrentExceptionMsg`. However, I've added an info box to chapter4 explaining that there are other ways to handle exceptions and we're using what we're using because our specific situation allows it.

docs/src/tutorials/http_client/chapter4.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+1. We use the [`.wait(timeout)`](/api/chronos/internal/asyncfutures.html#wait,Future[T],Duration) modifier on our `fetch` future.
+2. If the request takes longer than the provided duration, `.wait()` automatically cancels the underlying future and raises an [`AsyncTimeoutError`](/api/chronos/internal/errors.html#AsyncTimeoutError).
+3. We catch this error alongside other expected exceptions in our `except` block.
+
+```admonish info
+In Nim, there are several ways to capture the message from an exception:
+- using `getCurrentExceptionMsg()`, as we do in this tutorial
+- using `except Exception as e` and then calling `e.msg`
```
arnetheduck:
```suggestion
- using `except CatchableError as e` and then calling `e.msg`
```

`except Exception` is generally undefined behavior since it sometimes catches Defect and sometimes not - if we want to catch everything, we typically use `except CatchableError`:  https://status-im.github.io/nim-chronos/error_handling.html#the-exception-type

moigagoo: I meant `Exception` as a placeholder, i.e. "insert your exception type here," not as a concrete copyable code snippet. I never wanted to encourage catching overly general like `Exception` or `CatchableError`.

I've replaced `Exception` with `<Exception>` to make it more clear that it's meant to be replaced with a particular exception type.

PR #636, opened 2026-04-02
Set `epoll` engine as default for android

cheatfate:


arnetheduck (approved):

Ivansete-status:
@arnetheduck - we haven't tested it yet but we will go back to that if detect any issue when we work on mobile more intensively (cc @igor-sirotin)

PR #637, opened 2026-04-06
selector: dynamically adjust event count

arnetheduck:
When fetching events with epoll/kqueue, we currently use a fixed size array to fetch system events.

Taking an idea from
[libevent](https://github.com/libevent/libevent/blob/48296514d8fd9c0b3812b11d45ad80b0c002c14e/epoll.c#L568), we can instead dynamically grow the number of events fetched as the system comes under load thus reducing the number of poll calls, allowing more work to be done per batch.

The change also has the effect that data that already arrived at the time of the poll call gets processed before timers are fired - this means that timeouts are less likely to be triggered unfairly due to event queueing order and the time it takes to process each event.

* ~~increase max events per loop to 4096 (same as libevent)~~ - 64 is kept to not change the current balancing of timers vs network events
* reduce initial event list allocation to 32
* get rid of `ReadyKey.errorCode` which is unused


chronos/config.nim
```diff
@@ -65,10 +65,10 @@ const
     ## using `AsyncProcessOption.EvalCommand` and API calls such as
     ## ``execCommand(command)`` and ``execCommandEx(command)``.
 
-  chronosEventsCount* {.intdefine.} = 64
+  chronosEventsCount* {.intdefine.} = 4096
```
arnetheduck:
```suggestion
  chronosEventsCount* {.intdefine.} = 64
```

cheatfate (approved):

PR #640, opened 2026-04-08
Add ALPN support to TLSStream

fox0430:
Add Application-Layer Protocol Negotiation (ALPN) support to TLSStream for both client/server connections.

`{.emit.}` is used to set protocol_names on the BearSSL engine because the C header declares it as `const char **` while the Nim binding types it as `cstringArray (char **)`. Direct assignment fails on gcc 14+ where -Wincompatible-pointer-types is an error by default. Should I send a PR to nim-bearssl to fix the binding?

### DIFF
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -10,7 +10,7 @@ skipDirs      = @["tests"]
 requires "nim >= 1.6.16",
          "results",
          "stew >= 0.5.0",
-         "bearssl >= 0.2.7",
+         "bearssl >= 0.2.8",
          "httputils",
          "unittest2"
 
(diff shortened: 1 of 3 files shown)


arnetheduck:
> Should I send a PR to nim-bearssl to fix the binding?

Yes please, better fix these things at the source than working around them here.

fox0430:
I opened a PR to nim-bearssl. https://github.com/status-im/nim-bearssl/pull/87

arnetheduck:
thanks, now this branch needs to update requires to bearssl 0.2.8

fox0430:
Only the macos-amd64 Nim devel failed. Is it a Flaky test?

arnetheduck:
> Is it a Flaky test?

yeah, there sometimes port sharing issues on these afair, which would require some test refactoring to fix.

PR #644, opened 2026-04-09
streams: replace queue+loop with vtable

arnetheduck:
The current approach of specializing streams for specific protocols relies on each stream operating a buffer/queue and a read/write loop that replenishes or drains these.

For some protocols such as TLS, a buffer is naturally required since we're mutating the data along the way (in this case by encrypting it) - for others, such as chunking or limiting, the data remains unchanged and all the stream does is to either frame the data or provide limits and safety nets - for such streams, the buffer approach carries significant overhead - using a read as an example:

* a read loop starts reading from an underlying data source
* when data arrives, data is copied to a buffer
* when application wants to read, it checks the buffer and copies data to user buffer
* if data is not available, it waits for the buffer to be replenished by the read loop

In the above case, every byte must be copied twice and we have to wait an additional loop iteration for each stream layer as the loop wakes the actual reader.

Futher, each layer requires memory for the buffer - using a small buffer means high latency due to the extra event loop round trips while a large buffer is wasteful since most of the time, it's not actually being used (ie while waiting for data from a socket).

The API of streams is fairly complex - operations such as `readLine` and `readUntil` also require buffering since they must "look ahead" in the data to find separators and at the same time not "overread" their supporting source stream.

However, other operations do not require this extra buffering - `readOnce` for example is already given a buffer to write to and all it needs to do is continue using this buffer, similar as happens with [bulk reads](https://github.com/status-im/nim-chronos/pull/632).

Similarly, when servicing a `write`, we do not actually need to copy the data into a queue - all that's needed is that the data remains available throughout the write operation, a guaranteee conveniently provided by `async` already.

All in all, this means that for simple reads and writes, there is no inherent need to do buffering - the buffering is merely a side effect of the loop-based implementation.

By replacing the loop with a vtable, we can specialize the operations that a stream based on the specific requirements of the stream protocol being implemented instead of forcing them all to use a buffer.

This allows implementations that don't require a buffer to specialize each operation according to its specific needs - a chunked reader for example can pass the user buffer directly to the next layer, and if the next layer is a socket, it means that the read is done directly into the user buffer for the majority of data. Using knowledge about the limits, it can also pre-allocate a target buffer of the correct size instead of incrementally building it up.

A common structure of protocols is that they start with a header and then perform bulk transfers - for these protocols, we can use the stream buffer for the header read and perform the rest of the transfer directly to the user buffer.

The effect is significant - here's the http benchmark (http is a heavy user of streams):

Pre:
```
| Small/small    |    0.063s |   1000 | 15817.438 |    0.031 MB |    0.016 MB |    0.483 MB/s |    0.256 MB/s |
| Medium/small   |    0.845s |   1000 | 1182.766 | 1000.000 MB |    0.021 MB | 1182.766 MB/s |    0.025 MB/s |
| Small/Medium   |    1.045s |   1000 |  956.632 |    0.031 MB | 1000.000 MB |    0.029 MB/s |  956.632 MB/s |
| Medium/Medium  |    1.286s |   1000 |  777.588 | 1000.000 MB | 1000.000 MB |  777.588 MB/s |  777.588 MB/s |
```

Post:
```
| Small/small    |    0.073s |   1000 | 13774.390 |    0.031 MB |    0.016 MB |    0.420 MB/s |    0.223 MB/s |
| Medium/small   |    0.312s |   1000 | 3200.978 | 1000.000 MB |    0.021 MB | 3200.978 MB/s |    0.067 MB/s |
| Small/Medium   |    0.537s |   1000 | 1862.990 |    0.031 MB | 1000.000 MB |    0.057 MB/s | 1862.990 MB/s |
| Medium/Medium  |    0.880s |   1000 | 1136.555 | 1000.000 MB | 1000.000 MB | 1136.555 MB/s | 1136.555 MB/s |
```

* deprecate `udata` from streams - this is not being used anywhere and just creates surface area for bugs
* deprecate `loop`-based streams
* move all streams to vtable-based approach, avoiding `if`-based implementation selection
* keep readUntil/readLine implementation close to the buffer
* maintain atomicity of `write` calls with `AsyncLock` instead of a queue

Once these foundational changes are done, there are a few more things that can be improved:

* HTTP streams can be further simplified - in particular, the bounded stream serves many masters and could be split into a "byte-limiting" stream and a "readUntil"-limiting stream so that the implementation of each is simpler, composing them instead.
* TLS streams use a trick to simplify the implementation: instead of reading and writing concurrently, a read is initiated but all writes are processed before that read is awaited - the two operations could be made fully concurrent with some additional implementation complexity
* TLS streams are documented not to perform `close_notify` since this would require expanding the API to provide a notion of "full duplex close", something that the split reader/writer streams don't do well (in TLS 1.3, the protocol has been upgraded to support half- closed streams which would fit the model better)
* Closing sequences could be simplified - in particular, the concept of ownership should be introduced so that closing a "parent" stream also closes the underlying source, when this is desireable. Similarly, finishing a write stream should lead to `SHUT_WR` being sent on the wire making the underlying TCP closing sequence more predictable and efficient.
* Parts of the stream API can be simplified or removed - for example, there's no need to maintain so many "states" since we no longer need to track the activity of a "loop" that no longer exists.
* Potentially, buffered operations could be moved out from the API entirely and/or offered only by certain stream types, dividing the VTable into a "simple" unbuffered subset and the full buffered set of operations.

### DIFF
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -10,7 +10,7 @@ skipDirs      = @["tests"]
 requires "nim >= 1.6.16",
          "results",
          "stew >= 0.5.0",
-         "bearssl >= 0.2.7",
+         "bearssl >= 0.2.8",
          "httputils",
          "unittest2"
 
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -10,8 +10,8 @@
 {.push raises: [].}
 
 import std/[uri, tables, sequtils]
-import stew/[base10, base64, byteutils, ptrops], httputils, results
-import ../../asyncloop, ../../asyncsync
+import stew/[assign2, base10, base64, byteutils, ptrops, shims/sequninit], httputils, results
+import ../../[asyncloop, asyncsync, config]
 import ../../streams/[asyncstream, tlsstream, chunkstream, boundstream]
 import httptable, httpcommon, httpagent, httpbodyrw, multipart
 export results, asyncloop, asyncsync, asyncstream, tlsstream, chunkstream,
@@ -255,7 +255,7 @@ template isReady(conn: HttpClientConnectionRef): bool =
 
 template isIdle(conn: HttpClientConnectionRef, timestamp: Moment,
                 timeout: Duration): bool =
-  (timestamp - conn.timestamp) >= timeout
+  (timestamp - conn.timestamp) >= timeout or conn.transp.atEof()
 
 proc sessionWatcher(session: HttpSessionRef) {.async: (raises: []).}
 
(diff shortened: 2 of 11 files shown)


cheatfate (approved):

arnetheduck:
Oops, the code for this change actually got merged as part of #643 (since the new test was built on top of this branch)

PR #646, opened 2026-04-14
fix(httpclient): pipelining session handles closed connections

markspanbroek:
Before this change, HttpSession tried to reuse connections that were already closed by the server, leading to stream reading failures.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -255,7 +255,7 @@ template isReady(conn: HttpClientConnectionRef): bool =
 
 template isIdle(conn: HttpClientConnectionRef, timestamp: Moment,
                 timeout: Duration): bool =
-  (timestamp - conn.timestamp) >= timeout
+  (timestamp - conn.timestamp) >= timeout or conn.transp.atEof()
 
 proc sessionWatcher(session: HttpSessionRef) {.async: (raises: []).}
 
--- a/tests/testhttpclient.nim
+++ b/tests/testhttpclient.nim
@@ -1578,3 +1578,49 @@ suite "HTTP client testing suite":
     await session.closeWait()
     await server.stop()
     await server.closeWait()
+
+  proc createServerOk200(): HttpServerRef =
+    proc process(r: RequestFence): Future[HttpResponseRef] {.async: (raises: [CancelledError]).} =
+      try:
+        await r.get().respond(Http200, "ok")
+      except HttpWriteError:
+        defaultResponse()
+    createServer(initTAddress("127.0.0.1:0"), process, secure = false)
+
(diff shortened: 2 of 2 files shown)


cheatfate (approved):

tests/testhttpclient.nim
```diff
@@ -1578,3 +1578,49 @@ suite "HTTP client testing suite":
...
+    let session = HttpSessionRef.new(flags = flags, idlePeriod = idlePeriod)
+    server.start()
+    for _ in 0 ..< 10:
+      let request = HttpClientRequestRef.new(session, url)
+      let response = await request.send()
+      discard await response.consumeBody()
+      await request.closeWait()
+      await response.closeWait()
+      await server.drop() # server closes all keep-alive connections
+    check session.connectionsCount > 0
```
arnetheduck, line 1621: getting some spurious fails on this test: https://github.com/status-im/nim-chronos/actions/runs/26871407843/job/79247388456?pr=652#step:16:455 - cc @markspanbroek any idea why this would be flakey?

chronos/apps/http/httpclient.nim
```diff
@@ -255,7 +255,7 @@ template isReady(conn: HttpClientConnectionRef): bool =
 
 template isIdle(conn: HttpClientConnectionRef, timestamp: Moment,
                 timeout: Duration): bool =
-  (timestamp - conn.timestamp) >= timeout
+  (timestamp - conn.timestamp) >= timeout or conn.transp.atEof()
```
arnetheduck, line 258: ok, I see why it is flakey - `atEof` only becomes true once the EOF marker from the OS has been read - this might not happen until later, ie calling `recv` on a socket might return some bytes then a second `recv` is needed to read 0 bytes at which point `atEof` becomes true .. it means, among other things, that EOF cannot reliably be detected this way.

in other words, there's always a risk that a connection with keepalive will turn out to be closed by the server and there's nothing to be done except to retry the request (in case of an idempotent request like GET) or perform some application-specific action (in case of POST etc where you don't know if the request reached the server or not).

markspanbroek: Yes, that is true. You cannot reliably know if the server closed the connection. The check for `atEof()` detects most cases, but it does not detect all. There can still be race conditions, so you always have to have an application specific retry.

However, this doesn't really explain why the check on line 1621 fails. It is the check before waiting on the idle timeout. When this check fails, the client has 0 connections, so the client has already seen that the connections are closed (`atEOF is true`) before waiting for the idle period.

Maybe this happens occasionally because the server.drop() takes a longer than 100ms, and therefore the idle check has already happened?

I think that it is ok to remove the check on line 1621 from the test. The real test is the `check` on line 1623.

PR #650, opened 2026-05-22
fix: don't sleep/block if ticks are outstanding

dryajov:
Prevent `poll()` from blocking when internal `ticks` are outstanding.

When `cancelSoon()` fails to deliver cancellation on the first try, it installs a `tick` callback. However, `poll()` ignores `ticks` when calculating the current timeout. If a long timeout is scheduled, `poll()` can block until that timeout expires or IO wakes up the loop.

Treat `tick` callbacks as outstanding work when calculating the current `poll()` timeout.

### DIFF
--- a/tests/testtime.nim
+++ b/tests/testtime.nim
@@ -8,6 +8,7 @@
 import std/os
 import unittest2
 import ../chronos, ../chronos/timer
+import ../chronos/timer
 
 {.used.}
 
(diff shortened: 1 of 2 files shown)


tests/testtime.nim
```diff
@@ -127,3 +128,34 @@ suite "Asynchronous timers & steps test suite":
 
     check:
       fut3.completed() == true
+
+  test "Cancellation retry tick does not wait for next timer":
+    const NextTimerSleep = 50.milliseconds
```
arnetheduck:
```suggestion
    const NextTimerSleep = 500.milliseconds
```

arnetheduck:
hm, that's a lot of test failures .. almost looks like it was going into a busy loop sometimes

dryajov:
Interesting, I'll take a look

arnetheduck:
related: https://github.com/status-im/nim-chronos/pull/655

PR #618, opened 2026-02-14
http: add proxy support (fixes #193)

arnetheduck:
Both regular and secure proxies!

PR #651, opened 2026-05-29
http: add absolute request-target support (fixes #193)

### LINKED ISSUE #193: proxy implementation in httpclient
First of all, thank you guys so much for putting effort this amazing project. I really enjoy using it and learning how to make the NIM project clean and organized. I was wondering, does the httpclient support regular http/https proxy, just like NIM std httpclient library? I went through the code and I didn't find code can realize this functionality but just want to make sure. 

If not I think it should be easy to implement and I can try to do that then make PR. The only thing worries me that is you have very well defined test cases and I don't have anything on top of my head how to make a robust test case for the http proxy (mainly to find a long lifetime public proxy server).

arnetheduck:
Using an absolute request URI and a connection provider that connects to a different server than the http address, generic proxies can be implemented

* add detail to http connection error
* fix transport/stream leak when http client construction fails

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -555,7 +555,7 @@ proc getUniqueConnectionId(session: HttpSessionRef): uint64 =
   inc(session.counter)
   session.counter
 
-proc new*(
+proc new(
        t: typedesc[HttpClientConnectionRef],
        session: HttpSessionRef,
        ha: HttpAddress,
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -584,11 +584,7 @@ proc new*(
                                   flags = session.flags.getTLSFlags(),
                                   bufferSize = session.connectionBufferSize)
         except TLSStreamInitError as exc:
-          # TODO closing the reader of a StreamTransport is a no-op except for
-          #      the tracker so it's safe to spawn here - a bit ugly though,
-          #      would be better to add a "regular" way of doing this
-          asyncSpawn treader.closeWait()
-          asyncSpawn twriter.closeWait()
+          # TODO treader and twriter leak here
           return err(exc.msg)
 
       res = HttpClientConnectionRef(
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -10,7 +10,7 @@
 {.push raises: [].}
 
 import std/[uri, tables, sequtils]
-import stew/[assign2, base10, base64, byteutils, ptrops, shims/sequninit], httputils, results
+import stew/[assign2, base10, byteutils, ptrops, shims/sequninit], httputils, results
 import ../../[asyncloop, asyncsync, config]
 import ../../streams/[asyncstream, tlsstream, chunkstream, boundstream]
 import httptable, httpcommon, httpagent, httpbodyrw, multipart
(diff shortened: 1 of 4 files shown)


arnetheduck:
this is an alternative to #618

chronos/apps/http/httpclient.nim
```diff
@@ -571,6 +584,11 @@ proc new(
                                   flags = session.flags.getTLSFlags(),
                                   bufferSize = session.connectionBufferSize)
         except TLSStreamInitError as exc:
+          # TODO closing the reader of a StreamTransport is a no-op except for
+          #      the tracker so it's safe to spawn here - a bit ugly though,
+          #      would be better to add a "regular" way of doing this
+          asyncSpawn treader.closeWait()
+          asyncSpawn twriter.closeWait()
```
cheatfate: Please do not use `asyncSpawn` in core code.
```nim
  await allFutures(treader.closeWait(), twriter.closeWait())
```
`allFutures()` is also NOP here if `treader.closeWait()` and `twriter.closeWait()` will return finished Futures.

chronos/apps/http/httpserver.nim
```diff
@@ -442,7 +442,7 @@ proc updateRequest*(request: HttpRequestRef, scheme: string, meth: HttpMethod,
     if request.rawPath != "*":
       let uri = parseUri(request.rawPath)
       if uri.scheme notin ["http", "https", ""]:
-        return err(HttpMessage.init(Http400, "Unsupported URI scheme"))
+        return err(HttpMessage.init(Http400, "Unsupported URI scheme" & uri.scheme))
```
cheatfate: Its actually a security risk of using unquoted `uri.scheme` which we do not control, for example this could be used to escape database quoting, if we store errors as logs in database.

chronos/apps/http/httpclient.nim
arnetheduck: `new` is not async so that can't be done easily .. probably the easiest thing to do is to remove the TLSStreamInitError entirely, or change how asyncstreams are created and closed so that they don't need closing if they haven't been used but both changes seem out of scope here .. can also remove the fix entirely and let them leak but this TODO seemed like a reasonable compromise.

chronos/apps/http/httpserver.nim
arnetheduck:
```suggestion
        return err(HttpMessage.init(Http400, "Unsupported URI scheme"))
```

chronos/apps/http/httpclient.nim
arnetheduck:
```suggestion
          # TODO treader and twriter leak here
```

chronos/apps/http/httpclient.nim
```diff
@@ -542,7 +555,7 @@ proc getUniqueConnectionId(session: HttpSessionRef): uint64 =
   inc(session.counter)
   session.counter
 
-proc new(
+proc new*(
```
arnetheduck:
```suggestion
proc new(
```

chronos/apps/http/httpclient.nim
cheatfate: So what about idea to use?
```nim
  let
    closeReaderFut = treader.closeWait()
    closeWriterFut = twriter.closeWait()
  doAssert (closeReaderFut.finished == true) and (closeWriterFut.finished == true)
```
or its not always will be true?

cheatfate: you already closing because of error, so there is no need to handle errors from `closeWait`.

arnetheduck: `closeWait` cannot fail (it's `raises: []`, it doesn't even cancel) - the problem is that it might finish in a separate tick so `finished` is not necessarily true - above all, whether it's finished or not might change.

This is why `asyncSpawn` was a good compromise - it verifies that there are no errors, but also allows it to finish in any tick, later too (we want this code to be robust against changes to `closeWait` that delay closing to some other tick).

This code path in general is not tested and rarely (if ever) will happen practically - all the ways in which tls stream init can fail at this point are programming errors effectively ("invalid private key" for example cannot happen because you're not supposed to create a stream server with bad keys, ie those kinds of errors would be bugs in chronos).

arnetheduck: #651 fixes the leak by refactoring connection / stream ownership

PR #656, opened 2026-06-15
httpclient: add CONNECT proxy / tunnel support

arnetheduck:
Using `tunnel`, a general TCP tunnel can be created for any protocol via a HTTP proxy. This tunnel is typically used for establishing TLS connections via a proxy but any protocol can be used.

When a tunnel is established, the connection is removed from the session pool and given to the caller who is now responsible for closing it.

Because the tunnel connection itself can be encrypted, we return an `AsyncStream` that the caller must close (rather than a `StreamTransport`).

This PR also moves DNS name resolution to the connection provider - this has the effect of delaying DNS lookup until the actual connect attempt is made - earlier, DNS lookups would happen when creating the HTTP address. The move allows the proxy to perform the name resolution when relevant - the move also prepares the API for async name resolution.

Callers can still skip name resolution by providing an explicit list of addresses to connect to.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -47,7 +47,6 @@ type
   HttpClientConnectionState* {.pure.} = enum
     Closed                    ## Connection has been closed
     Closing,                  ## Connection is closing
-    Resolving,                ## Resolving remote hostname
     Connecting,               ## Connecting to remote server
     Ready,                    ## Connected to remote server
     Acquired,                 ## Connection is acquired for use
@@ -153,6 +152,8 @@ type
     username*: string
     password*: string
     addresses*: seq[TransportAddress]
+      ## If addresses are set, name resolution is not performed and the client
+      ## will instead attempt to connect to the given addresses (in order)
 
   HttpClientRequest* = object
     state: HttpReqRespState
@@ -322,7 +323,6 @@ proc getTLSFlags(flags: HttpClientFlags): set[TLSFlags] =
 
 proc getHttpAddress*(
        url: Uri,
-       flags: HttpClientFlags = {}
      ): HttpAddressResult =
   let
     scheme =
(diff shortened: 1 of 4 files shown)


tests/testhttpclient.nim
```diff
@@ -1478,6 +1478,93 @@ suite "HTTP client testing suite":
   test "HTTP proxy connection provider test":
     check waitFor(testHttpProxyConnectionProvider()) == true
 
+  proc testConnectTunnel(): Future[bool] {.async.} =
```
cheatfate: I think new tests should follow async raises semantics.

tests/testhttpclient.nim
```diff
@@ -1478,6 +1478,93 @@ suite "HTTP client testing suite":
...
+    await req.closeWait()
+    await session.closeWait()
+    await server.stop()
+    await server.closeWait()
+    return true
+
+  test "HTTP connect tunnel test":
+    check waitFor(testConnectTunnel()) == true
+
+  proc testTunnelConnectionProvider(): Future[bool] {.async.} =
```
cheatfate: One more test which should follow async raises.

chronos/apps/http/httpclient.nim
```diff
@@ -1751,3 +1788,79 @@ proc getServerSentEvents*(
...
+      # httpProxyProvider above
+      targetHa = request.address
+      req = HttpClientRequestRef.new(proxySession, targetHa, MethodConnect)
+      resp =
+        try:
+          await req.send()
+        except HttpError as exc:
+          raiseHttpConnectionError("Could not connect to proxy: " & exc.msg)
+        finally:
+          await req.closeWait()
```
cheatfate, line 1854: `await` in finally still does not work correctly, it is better to avoid it.

chronos/apps/http/httpclient.nim
```diff
@@ -1751,3 +1788,79 @@ proc getServerSentEvents*(
...
+      # we can extract a stream that we can use to establish the end-to-end
+      # connection - both the proxy and the inner connections can be
+      # using TLS but typically only the inner connection will be
+      stream =
+        try:
+          await resp.tunnel()
+        except HttpError as exc:
+          raiseHttpConnectionError("Could not connect to proxy: " & exc.msg)
+        finally:
+          await resp.closeWait()
```
cheatfate, line 1866: `await` in finally again.

chronos/apps/http/httpclient.nim
```diff
@@ -1514,6 +1473,84 @@ proc consumeBody*(response: HttpClientResponseRef): Future[int] {.
...
+      if otherClosed:
+        if transp != nil:
+          pending.add transp.closeWait()
+      else:
+        otherClosed = true
+      pending.add inner(rstream)
+      noCancel allFutures(pending)
+  wclose.vtbl.close = block:
+    var inner = wclose.vtbl.close
+    proc(wstream: AsyncStreamWriter) {.async: (raises: [], raw: true).} =
```
cheatfate: I do not like anonymous procedures because its hard to read it in stack traces. Could you please make it not anonymous?

chronos/apps/http/httpclient.nim
```diff
@@ -1514,6 +1473,84 @@ proc consumeBody*(response: HttpClientResponseRef): Future[int] {.
...
+        (move(connection.treader), move(connection.twriter))
+    rclose = newAsyncStreamReader(reader)
+    wclose = newAsyncStreamWriter(writer)
+
+  # Create a stream wrapper that will close the underlying streams and eventually
+  # the transport too
+  var otherClosed = false
+  rclose.vtbl.close = block:
+    var inner = rclose.vtbl.close
+    proc(rstream: AsyncStreamReader) {.async: (raises: [], raw: true).} =
```
cheatfate: Could you please avoid anonymous procedures? Its hard to read stack traces and use other debugging utilities with it.

chronos/apps/http/httpclient.nim
```diff
@@ -630,55 +563,113 @@ proc closeWait(conn: HttpClientConnectionRef) {.async: (raises: []).} =
...
+      try:
+        await conn.tls.handshake()
+      except AsyncStreamError as exc:
+        raiseHttpConnectionError(exc.msg)
+    conn.state = HttpClientConnectionState.Ready
+    connected = true
+    conn
+  finally:
+    if not connected:
+      await noCancel conn.closeWait()
```
cheatfate, line 638: Please do not use `await` in `finally` blocks.

cheatfate (changes requested):

chronos/apps/http/httpclient.nim
arnetheduck: the finally bugs were fixed long long ago - if it didn't work, the tests would fail when checking the close tracker - since the trackers are ok, it means this works

chronos/apps/http/httpclient.nim
arnetheduck: aa6909165ee20969329e609540be091db9849413

chronos/apps/http/httpclient.nim
arnetheduck: aa6909165ee20969329e609540be091db9849413

tests/testhttpclient.nim
arnetheduck: 890d4ab7a00e43846ff759fa072bd482acb64949

tests/testhttpclient.nim
arnetheduck: 890d4ab7a00e43846ff759fa072bd482acb64949

cheatfate (approved):

PR #660, opened 2026-06-17
tlsstream: fix PEM decoder callback type for bearssl > 0.2.8

fox0430:
bearssl > 0.2.8 changed the PEM decoder callback source parameter to ConstPointer (const void *) for GCC 14 compatibility. Match that in itemAppend while keeping backward compatibility with <= 0.2.8 via when declared(ConstPointer).

Link https://github.com/status-im/nim-bearssl/pull/89#issuecomment-4729991711


arnetheduck:
No longer needed as of https://github.com/status-im/nim-bearssl/pull/95

PR #664, opened 2026-06-19
Avoid race when reading and writing concurrently on TLS stream

etan-status:
Engine state can change while awaiting lock acquisition from #643, and trigger `sslEngineSendrecBuf` even though different task already read: `length != 0` doesn't necessary hold in that case -> `AssertionDefect`.

### DIFF
--- a/chronos/streams/tlsstream.nim
+++ b/chronos/streams/tlsstream.nim
@@ -197,7 +197,7 @@ proc runUntil(rws: TLSAsyncStream, target: cuint): Future[void] {.
   let engine = rws.engine()
   while true:
     func remainingState(
-        rws: TLSAsyncStream, target: cuint
+        rws: TLSAsyncStream, engine: ptr SslEngineContext, target: cuint
     ): Opt[cuint] {.raises: [AsyncStreamError].} =
       let err = sslEngineLastError(engine[])
       if err != 0:
@@ -214,13 +214,13 @@ proc runUntil(rws: TLSAsyncStream, target: cuint): Future[void] {.
 
       Opt.some state
 
-    if rws.remainingState(target).isNone:
+    if rws.remainingState(engine, target).isNone:
       break
 
     # Prevent concurrent reads and writes from interfering with each other
     await rws.lock.acquire()
     try:
-      let state = rws.remainingState(target).valueOr:
+      let state = rws.remainingState(engine, target).valueOr:
         break
 
       # If the engine is ready for reading, start a read operation - although
--- a/chronos/streams/tlsstream.nim
+++ b/chronos/streams/tlsstream.nim
@@ -157,9 +157,6 @@ template newTLSStreamProtocolImpl[T](message: T): ref TLSStreamProtocolError =
   err.errCode = code
   err
 
-template newTLSUnexpectedProtocolError(): ref TLSStreamProtocolError =
-  newException(TLSStreamProtocolError, "Unexpected internal error")
-
 proc newTLSStreamProtocolError[T](message: T): ref TLSStreamProtocolError =
   newTLSStreamProtocolImpl(message)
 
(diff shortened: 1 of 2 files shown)


arnetheduck:
Good catch

etan-status:
There's also a lockup where readers get blocked when an incomplete write exists, but that one is only triggerable with pipelining, which we don't do I guess. At least not a crash.

chronos/streams/tlsstream.nim
```diff
@@ -199,22 +196,33 @@ proc runUntil(rws: TLSAsyncStream, target: cuint): Future[void] {.
...
-        rws.handshaked = true
+      Opt.some state
+
+    if rws.remainingState(target).isNone:
       break
 
     # Prevent concurrent reads and writes from interfering with each other
     await rws.lock.acquire()
     try:
+      let state = rws.remainingState(target).valueOr:
```
arnetheduck: actually, we redo this check in `readOnce`/`write` so it should not be needed here, ie a spurious wakeup is ok.. hm

tests/testasyncstream.nim
```diff
@@ -951,6 +951,68 @@ suite "AsyncStream/TLSStream":
...
+          await stepsAsync(1)
+          length += chunk.len
+
+      proc doRead() {.async.} =
+        var length = 0
+        while length < TotalSize:
+          let chunk = await s.reader.read(ChunkSize)
+          if chunk.len == 0:
+            break
+          await stepsAsync(1)
```
arnetheduck, line 975: this test gets stuck because once all data has been sent / read, there are no ongoing ticks here, so `stepsAsync` will never make progress (the last read on the last of the two sockets has a `len != 0`, but nothing is happening on the wire any more)

chronos/streams/tlsstream.nim
arnetheduck: nevermind though, this is more a comment about the general code flow

tests/testasyncstream.nim
etan-status: platform-specific, apparently

arnetheduck: `sleepAsync(1.ms)` instead and it'll be fine

arnetheduck:
> where readers get blocked

only one read at a time is allowed, if that's what you mean by pipelining

etan-status:
with pipelining I mean sending a couple requests before the response of the first one is read.

chronos/streams/tlsstream.nim
```diff
@@ -199,22 +196,33 @@ proc runUntil(rws: TLSAsyncStream, target: cuint): Future[void] {.
      async: (raises: [CancelledError, AsyncStreamError]).} =
   let engine = rws.engine()
   while true:
-    let err = sslEngineLastError(engine[])
-    if err != 0:
-      raise newTLSStreamProtocolError(err)
+    func remainingState(
+        rws: TLSAsyncStream, target: cuint
+    ): Opt[cuint] {.raises: [AsyncStreamError].} =
+      let err = sslEngineLastError(engine[])
```
arnetheduck, line 202: the engine access here causes an unnecessary memory allocation for the closure ..

arnetheduck:
> sending a couple requests before the response of the first one is read.

every time you send, you must also read when you use TLS because the protocol is bidirectional underneath - ie in order to send data, sometimes the engine must process incoming packets (to perform key renegotiations) and in order to process incoming packets your application must clear the incoming queue of already-decryptable data .. 

this is a general problem in many protocols, ie often it is the case that in order to make progress on sends, reading must be happening in an ongoing manner .. in chronos, this means that although you can queue up a bunch of sends, you must have a read loop that is independent of those sends.

etan-status:
I'm talking about the application layer RECVAPP / SENDAPP, if applications queues up a couple requests without reading the first (application layer) response, this impl will get stuck / won't write additional requests anymore.

tests/testasyncstream.nim
etan-status: With sleepAsync(1.ms), the test doesn't do the same check anymore, i.e., it no longer fails without the fix being present.

chronos/streams/tlsstream.nim
etan-status: Passing engine as well to avoid the closure

PR #666, opened 2026-06-23
tls: restore concurrent reads/writes

arnetheduck:
https://github.com/status-im/nim-chronos/pull/644 introduced a simplification that caused reads and writes to be mutually exclusive - while simpler in implementation it can lead to deadlocks where reading cannot proceed because writing is blocked when the other end is not reading because itself is blocked on writing.


chronos/streams/tlsstream.nim
```diff
@@ -173,110 +175,130 @@ proc new*(T: typedesc[TrustAnchorStore],
              "Anchors should be copied")
   TrustAnchorStore(anchors: res)
 
-proc dumpState*(state: cuint): string =
```
cheatfate, line 176: Please do not remove my `dumpState`, it is proved to being useful when debugging issues with this stream.

cheatfate (approved):

PR #667, opened 2026-06-23
Add tutorial examples to `nimble examples`

### LINKED ISSUE #665: Cover examples in CI
To make sure the examples in the docs are valid, we should run them during CI.

moigagoo:
Closes #665


chronos.nimble
```diff
@@ -55,10 +55,20 @@ proc tryExec(cmd: string) =
...
   # Build book examples
-  for file in listFiles("docs/examples"):
+  for file in listFiles("examples"):
     if file.endsWith(".nim"):
       build "--threads:on", file
 
+  # Build HTTP client tutorial examples
+  for chapterDir in listDirs("examples/http_client"):
+    withDir(chapterDir):
+      tryExec "nimble build"
```
nitely, line 65: this will install and test the examples with the latest chronos release instead of the PR. I'm not sure why the book examples are projects. It would make sense if they depended on some lib, but they only depend on chronos.

Also, some of the examples require nim#head which seem to build nim which makes the CI take longer, and not test the intended Nim CI version...

moigagoo: > this will install and test the examples with the latest chronos release instead of the PR.

Thanks, fixing this now.

> I'm not sure why the book examples are projects. It would make sense if they depended on some lib, but they only depend on chronos.

The intention was to provide examples as self-contained code that you can simply clone and run with `nimble build`.

> Also, some of the examples require nim#head which seem to build nim which makes the CI take longer, and not test the intended Nim CI version...

This is a poor decision to circumvent around a bug with getCurrentExceptionMsg() in Nim 2.2.0. I'll fix it.

PR #668, opened 2026-06-23
Do not return to scheduler while completing {.async.} proc

etan-status:
Chronos can have an extra implicit return to scheduler when an `{.async.}` proc completes.

```nim
import chronos

proc competitorCb(udata: pointer) =
  echo "competitor"

proc inner {.async.} =

proc middle {.async.} =
  echo "B"
  await inner()
  callSoon(competitorCb)
  echo "C"

proc outer {.async.} =
  echo "A"
  await middle()
  echo "D"

waitFor outer()
```

Output:

```
A
B
C
competitor
D
```

Since `inner` does not complete synchronously, Chronos adds an extra return to scheduler between C -> D, during which different callbacks / events may be processed. However, the `{.async.}` transformation macro hides this extra return to scheduler from the user: Visually this looks like a regular `return` to the outer proc, even though potential side effects similar to those from an explicit `await` have to be considered.

This PR removes the extra return to scheduler when completing an `{.async.}` proc, even when the proc did not fully run to completion synchronously. For other, explicitly added callbacks, existing behaviour is unchanged; a new FutureFlag.SyncContinuations can be used to emulate the {.async.} proc behaviour

⚠️ New behaviour is only activated when the opt-in config `chronosSyncContinuations` is enabled.

This is in line with other popular languages, where continuations across returns commonly complete without returning to the scheduler. Users more familiar with these other languages as well as LLMs may introduce subtle bugs, because the keywords and control flow look the same, while semantically behaving slightly different in Chronos in certain situations.

| Language | Output |
|--------|--------|
| Nim (chronos) | `A B C competitor D` | 
| C# | `A B C D competitor` |
| Python (asyncio) | `A B C D competitor` | 
| JavaScript (setTimeout) | `A B C D competitor` |
| JavaScript (queueMicrotask) | `A B C competitor D` |
| Rust (tokio) | `A B C D competitor` |
| C++ (asio) | `A B C D competitor` |
| Swift | `A B C D competitor` |
| Kotlin | `A B C D competitor` |

Note the two JavaScript rows, where the `queueMicrotask` API enables queueing the competitor ahead of the return-continuation. In contrast to Chronos, this still does not allow timers and I/O events to be executed (they are macrotasks).

-----

LLM-generated translations.

### C#

```cs
using System;
using System.Threading;
using System.Threading.Tasks;

ThreadPool.SetMinThreads(1, 1);
ThreadPool.SetMaxThreads(1, 1);

Task competitor = Task.CompletedTask;

await Outer();
await competitor;

async Task Inner() => await Task.Delay(100);

async Task Middle()
{
    Console.WriteLine("B");
    await Inner();
    competitor = Task.Run(() => Console.WriteLine("competitor"));
    Console.WriteLine("C");
}

async Task Outer()
{
    Console.WriteLine("A");
    await Middle();
    Console.WriteLine("D");
}
```

### Python

```python
import asyncio

def competitor_cb():
    print("competitor")

async def inner():
    await asyncio.sleep(0.1)

async def middle():
    print("B")
    await inner()
    asyncio.get_running_loop().call_soon(competitor_cb)
    print("C")

async def outer():
    print("A")
    await middle()
    print("D")

asyncio.run(outer())
```

### JavaScript (setTimeout)

```js
function competitor() { console.log("competitor"); }

async function inner() {
  await new Promise(r => setTimeout(r, 100));
}

async function middle() {
  console.log("B");
  await inner();
  setTimeout(competitor, 0);
  console.log("C");
}

async function outer() {
  console.log("A");
  await middle();
  console.log("D");
}

outer();
```

### JavaScript (queueMicrotask)

```js
function competitor() { console.log("competitor"); }

async function inner() {
  await new Promise(r => setTimeout(r, 100));
}

async function middle() {
  console.log("B");
  await inner();
  queueMicrotask(competitor);
  console.log("C");
}

async function outer() {
  console.log("A");
  await middle();
  console.log("D");
}

outer();
```

### Rust

```rust
use tokio::time::{sleep, Duration};

async fn inner() {
    sleep(Duration::from_millis(100)).await;
}
async fn middle() {
    println!("B");
    inner().await;
    tokio::spawn(async { println!("competitor"); });
    println!("C");
}
async fn outer() {
    println!("A");
    middle().await;
    println!("D");
}
#[tokio::main(flavor = "current_thread")]
async fn main() {
    outer().await;
    tokio::task::yield_now().await;
}
```

### C++

```cpp
#include <asio.hpp>
#include <chrono>
#include <iostream>

using asio::awaitable;
using asio::co_spawn;
using asio::detached;
using asio::use_awaitable;
using namespace std::chrono_literals;

void competitor() { std::cout << "competitor\n"; }

awaitable<void> inner() {
    asio::steady_timer timer(co_await asio::this_coro::executor);
    timer.expires_after(100ms);
    co_await timer.async_wait(use_awaitable);
}

awaitable<void> middle() {
    std::cout << "B\n";
    co_await inner();
    asio::post(co_await asio::this_coro::executor, competitor);
    std::cout << "C\n";
}

awaitable<void> outer() {
    std::cout << "A\n";
    co_await middle();
    std::cout << "D\n";
}

int main() {
    asio::io_context ctx;
    co_spawn(ctx, outer(), detached);
    ctx.run();
}
```

### Swift

```swift
func competitor() { print("competitor") }

@MainActor func inner() async {
    try? await Task.sleep(for: .milliseconds(100))
}

@MainActor func middle() async {
    print("B")
    await inner()
    Task { competitor() }
    print("C")
}

@MainActor func outer() async {
    print("A")
    await middle()
    print("D")
}

await outer()
try? await Task.sleep(for: .milliseconds(50))
```

### Kotlin

```kt
import kotlinx.coroutines.*

fun competitor() { println("competitor") }

suspend fun inner() {
    delay(100)
}

suspend fun CoroutineScope.middle() {
    println("B")
    inner()
    launch { competitor() }
    println("C")
}

suspend fun CoroutineScope.outer() {
    println("A")
    middle()
    println("D")
}

fun main() = runBlocking {
    outer()
}
```


etan-status:
`make -j test` successful in status-im/nimbus-eth2

tests/testunwind.nim
```diff
@@ -0,0 +1,195 @@
...
+        @["producer raising", "consumer caught", "competitor", "observer"]
+
+  test "Manual wakeup interruptible test":
+    check testManualWakeup() == @["competitor", "producer returns"]
+
+  test "Fan-out depth-first test":
+    ## Each strain unwinds to its conclusion before the next strain begins
+    check testFanout() in [
+      @["produced", "subA", "strainA", "subB", "strainB"],
+      @["produced", "subB", "strainB", "subA", "strainA"]]
```
etan-status:
```suggestion
    # Each strain unwinds to conclusion before the next strain begins.
    # Executed in reverse registration order.
    check testFanout() ==
      @["produced", "subB", "strainB", "subA", "strainA"]
```

tests/testall.nim
```diff
@@ -7,20 +7,15 @@
 #              MIT license (LICENSE-MIT)
 import ".."/chronos/config
 
-when (chronosEventEngine in ["epoll", "kqueue"]) or defined(windows):
-  import testmacro, testsync, testsoon, testtime, testfut, testsignal,
+when chronosEventEngine != "" or defined(windows):
```
cheatfate: Please undo this change.

etan-status: a9a2746

chronos/internal/asyncfutures.nim
```diff
@@ -176,18 +176,22 @@ proc checkFinished(future: FutureBase, loc: ptr SrcLoc) =
...
 
+proc scheduleCall(
+    callback: AsyncCallback, isFromAsyncMacro: bool) {.raises: [], gcsafe.}
+
 proc finish(fut: FutureBase, state: FutureState, loc: ptr SrcLoc) =
   fut.checkFinished(loc)
 
   fut.internalState = state
   fut.internalCancelcb = nil # release cancellation callback memory
+  let isFromAsyncMacro = not(isNil(fut.internalClosure))
```
etan-status: this is not strictly necessary, but limits the impact to just the return-continuations of `{.async.}` proc.

we could also prioritize all `callback.function == internalContinue` if we want the same treatment for any continuations. that would effectively make it so that ready continuations keep running until returning to the event loop. manual `callSoon` would still defer to the next `poll` tick, as only `internalContinue` gets prioritized. 

It's hard to reason about the impact of widening this, it may make it easier to accidentally lag out the event loop, if one doesn't sprinkle in a couple explicit `await sleepAsync(ZeroDuration)` in tight callback chains.

maybe a new flag on `FutureBase` could be a compromise, that enables users with manually created futures to opt-in (or opt-out) to the proposed `{.async.}` behaviour. we would then check for the flag here, and the async macro would set it on its own synthesized futures.

would need to collect more data to see whether restrictions make sense. it could also entirely be practical to just priotize all `callback.function == internalContinue` without flags or restricting to `{.async.}` origin.

wondering what's the best way to test any of this...

etan-status:
Example bug that is addressed by this PR: https://github.com/status-im/nim-chronos/pull/273

Obviously, it is possible to write correct code even without this PR. But when translating from other languages, doing spec copying, or using LLMs, subtle bugs can creep in. These bugs are hard to find in a review because they come from hidden implicit behaviour rather than visibly incorrect lines of code. This PR simply removes this class of footgun.

How much it makes it easier to accidentally stall the event loop, remains to be seen.

chronos/internal/asyncfutures.nim
arnetheduck: > wondering what's the best way to test any of this...

we're not going to be able to exhaustively test it - that said, a `chronosLegacyScheduling` flag in the config should probably be available for a release or two so the behavior can be switched off.

etan-status: ah, with a global config flag, it's probably also alright to apply this to _all_ futures (plain `callback.function == InternalContinue`, don't restrict to `{.async.}` proc). if there's a problem with it, either sprinkle in a couple explicit `await sleepAsync(ZeroDuration)` or opt-out.

arnetheduck: > alright to apply this to all futures

this approach certainly feels easier to reason about, specially when you consider "raw" async functions that went through the macro but not the closure transform.

> chains

This is the more interesting bit indeed - the current loop has a sentinel for scheduling new work created during the execution of the "current" tasks to being run after the current tasks, for interleaving ticks and timers

> sprinkle

this I guess might be something to focus on a bit - ie whether to add an explicit yielding function and/or expose execution ordering to the user (`proc callSoon(..., immediate: static bool = false)`)

etan-status: > apply this to all futures

have introduced a `FutureFlag.SyncContinuations` so that `{.async.}` is not the only way to get the new behaviour.

> chains

yeah, doing it on _every_ future is too aggresssive, things like `AsyncEvent` document a very specific resumption order. with `FutureFlag.SyncContinuations` one can choose the behaviour at future creation time

> immediate: static bool = false

done

tests/testcontinuations.nim
```diff
@@ -0,0 +1,230 @@
...
+
+    test "Manual wakeup interruptible test":
+      check testManualWakeup() == @["competitor", "producer returns"]
+
+    test "Manual wakeup not interrupted test":
+      check testManualSyncWakeup() == @["producer returns", "competitor"]
+
+    test "Multiple waiters test":
+      check testMultipleWaiters() ==
+        @["produced", "subA", "strainA", "subB", "strainB"]
```
etan-status, line 330: - `produced subB strainB subA strainA` (depth-first LIFO): Swift, Rust (tokio)
- `produced subA strainA subB strainB` (depth-first FIFO): Python (asyncio), Kotlin, C++ (asio), C#
- `produced subA subB strainA strainB` (breadth-first FIFO): JavaScript

Intuitively, the Python/C# version seems clearest. Execute waiters in registration order. It needs the double iteration over the loop though in asyncfutures, to achieve the FIFO ordering.

chronos/config.nim
```diff
@@ -35,6 +35,11 @@ const
     ##
     ## `Exception` handling may be removed in future chronos versions.
 
+  chronosSyncContinuations* {.booldefine.}: bool = false
+    ## When enabled, `await` continuations of futures created with
+    ## `FutureFlag.SyncContinuations` are run immediately, without
+    ## allowing other callbacks or events to be processed in-between.
```
arnetheduck:
```suggestion
    ## `FutureFlag.SyncContinuations` are scheduled to run before processing
    ## other already queued events such as timer handlers and I/O.
    ##
    ## This feature is experimental and may be removed in future releases.
```

In particular, when a future has more than one callback registered, those callbacks will not run "immediately" - they will be interleaved by nested continuations.

chronos/internal/asyncfutures.nim
```diff
@@ -176,12 +177,24 @@ proc checkFinished(future: FutureBase, loc: ptr SrcLoc) =
...
   fut.internalState = state
   fut.internalCancelcb = nil # release cancellation callback memory
 
+  when chronosSyncContinuations:
+    if FutureFlag.SyncContinuations in fut.internalFlags:
+      proc process(callback: var AsyncCallback) =
+        if callback.isAwaitContinuation:
+          callSoon(move(callback), immediate = true)
+
+      for i in countdown(fut.internalCallbacks.high, 0):
```
arnetheduck: need to `reset` internal callbacks here or the code below will perform a pointless iteration

chronos/internal/asyncfutures.nim
```diff
@@ -176,12 +177,24 @@ proc checkFinished(future: FutureBase, loc: ptr SrcLoc) =
...
 proc finish(fut: FutureBase, state: FutureState, loc: ptr SrcLoc) =
   fut.checkFinished(loc)
 
   fut.internalState = state
   fut.internalCancelcb = nil # release cancellation callback memory
 
+  when chronosSyncContinuations:
+    if FutureFlag.SyncContinuations in fut.internalFlags:
+      proc process(callback: var AsyncCallback) =
+        if callback.isAwaitContinuation:
```
arnetheduck: why is this if here? the presence of the `SyncContinuations` flag should be enough?

chronos/internal/asyncfutures.nim
etan-status: this only processes the `immediate = true` continuations (have to loop in background order).
while the code below adds the regular `callSoon` continuations (forward order).

chronos/internal/asyncfutures.nim
etan-status: Gives the priority only for the `await` flow.

```nim
  proc competitorCb(udata: pointer) =
    cast[ptr Trace](udata)[].add "competitor"

  proc observerCb(udata: pointer) =
    cast[ptr Trace](udata)[].add "observer"

  proc testObserverReturn(): Trace =
    proc producer(
        trace: ptr Trace): Future[int] {.async: (raises: [CancelledError]).} =
      await sleepAsync(ZeroDuration)
      callSoon(competitorCb, trace)
      trace[].add "producer returns"
      7

    proc consumer(trace: ptr Trace) {.async: (raises: [CancelledError]).} =
      let fut = producer(trace)
      fut.addCallback(observerCb, trace)
      let v = await fut
      trace[].add "consumer returns " & $v

    runTest consumer
```

i.e., gives:

```nim
@["producer returns", "consumer returns 7", "competitor", "observer"]
```

instead of:

```nim
@["producer returns", "observer", "consumer returns 7", "competitor"]
```

Either order is fine, though. I guess treating `addCallback` and `await` equivalent is a bit more pure as it doesn't force one specific idiom.

chronos/internal/asyncfutures.nim
etan-status: de0f175

chronos/internal/asyncfutures.nim
etan-status: de0f175

tests/testcontinuations.nim
etan-status: de0f175

chronos/internal/asyncfutures.nim
```diff
@@ -297,37 +309,71 @@ proc clearCallbacks(future: FutureBase) =
...
+    future: FutureBase, cb: CallbackFunc, udata: pointer,
+    flags: static set[CallbackFlag] = {}) =
   ## Adds the callbacks proc to be called when the future completes.
   ##
   ## If future has already completed then ``cb`` will be called immediately.
   doAssert(not isNil(cb))
-  if future.finished():
-    callSoon(cb, udata)
+
+  const _: set[CallbackFlag] = flags
```
etan-status:
```suggestion
  const _: set[CallbackFlag] = flags  # https://github.com/nim-lang/Nim/issues/25938
```

chronos/internal/asyncengine.nim
```diff
@@ -1301,16 +1307,29 @@ proc removeTimer*(at: uint64, cb: CallbackFunc, udata: pointer = nil) {.
      inline, deprecated: "Use removeTimer(Duration, cb, udata)".} =
   removeTimer(Moment.init(int64(at), Millisecond), cb, udata)
 
-proc callSoon*(acb: AsyncCallback) =
-  ## Schedule `cbproc` to be called as soon as possible.
-  ## The callback is called when control returns to the event loop.
-  getThreadDispatcher().callbacks.addLast(acb)
+proc callSoon*(acb: AsyncCallback, immediate: static bool = false) =
```
cheatfate: I think there should be `callImmediately` or probably `callNow` primitive which probably should be private for now. There is no need to modify `callSoon`.

chronos/internal/asyncfutures.nim
```diff
@@ -182,6 +183,17 @@ proc finish(fut: FutureBase, state: FutureState, loc: ptr SrcLoc) =
   fut.internalState = state
   fut.internalCancelcb = nil # release cancellation callback memory
 
+  when chronosSyncContinuations:
+    proc process(callback: var AsyncCallback) =
+      if not(isNil(callback.function)):
+        callSoon(move(callback), immediate = true)
```
cheatfate: this should call `callNow` or `callImmediately`.

cheatfate:
One of the core questions about this PR? Why it is made only for `async` transformed procedures?

chronos/internal/asyncfutures.nim
etan-status: e107a9f776e4a91a3b00f09765c1ff44a6586257

chronos/internal/asyncengine.nim
etan-status: Renamed to callImmediately, and made private (moved to asyncfutures where it is used).

b42dd0578694aad8c00a82cbda4db08fb3bf9e0a

etan-status (commented):
> Why it is made only for async transformed procedures?

With `FutureFlags.SyncContinuations` it's possible to opt-in without requiring the `{.async.}` / `await` syntax.

But I just realized that now that the "Combinator first-finisher test" ordering is fixed, that we can drop the other flag (`CallbackFlag.Continuation`), making this implementation much simpler. Will repeat the CI / local testnet tests in nimbus-eth2/eth1.

etan-status:
And the other flag is also dropped :-) Should still work fine!

etan-status:
- nimbus-eth2: https://github.com/status-im/nimbus-eth2/pull/8737
- nimbus-eth1: https://github.com/status-im/nimbus-eth1/pull/4451

etan-status:
There's some client usage that assumes certain ordering, i.e., that a later `addCallback` gets executed after previous callbacks, e.g., in peer_pool `addPeerImpl`. I think I'll go back to a flag / a separate function, to separate continuations from callSoon'ed callbacks. The idea is that client code should not need any changes. We can only guarantee that for `{.async.}` / `await`, where we control both future creation and callback registration.

etan-status:
The flag on the future is also needed, so that ordering is preserved when having both `addCallback` as well as `await` continuations on same future. Otherwise `await` can overtake regular `addCallback`. So yeah, back to the flag to avoid breaking existing usage.

Synchronous continuation path is once again only used when **both** producer and consumer opt-in (by default that's the {.async.} + await combo, but the `FutureFlag.SyncContinuations` / `addContinuationCallback` can be used in other code to achieve same behaviour).

etan-status:
nimbus-eth2/eth1 CI back to green (some flaky HTTP 502 aside).

chronos/internal/asyncfutures.nim
```diff
@@ -136,6 +137,20 @@ proc done*(future: FutureBase): bool {.deprecated: "Use `completed` instead".} =
   ## This is an alias for ``completed(future)`` procedure.
   completed(future)
 
+proc callImmediately(acb: AsyncCallback) =
+  ## Schedule `acb` to be called immediately.
+  ## No other scheduled callbacks or events are called before it.
```
arnetheduck: except those from another `callImmediately` call before control is given back 🤷

chronos/internal/asyncfutures.nim
```diff
@@ -317,18 +348,57 @@ proc addCallback*(future: FutureBase, cb: CallbackFunc) =
   ## If future has already completed then ``cb`` will be called immediately.
   future.addCallback(cb, cast[pointer](future))
 
+proc addContinuationCallback*(
```
arnetheduck, line 351: wait, what happened here? why did this become a separate function and why does the regular addCallback no longer respect the SyncContinuations flag?

chronos/internal/asyncfutures.nim
etan-status: Yes, it completes everything in one recursive await chain until top (see the "Multiple waiters test" in testcontinuations)

```suggestion
  ## No other already scheduled callbacks or events are called before it.
```

Want to add "already" to the docs?

chronos/internal/asyncfutures.nim
etan-status: Was requested to split `callSoon(..., immediate = true)` to a separate `callImmediately`: https://github.com/status-im/nim-chronos/pull/668#discussion_r3838839610

Applied the same concept here for consistency, instead of having `addCallback(..., {CallbackFlag.Continuation})`, name it `addContinuationCallback(...)`. Do you prefer the flag over overloads? Note that it was always a _separate_ flag for the callback (the `FutureFlag.Synccontinuations` is only on the future). 

Both sides have to opt-in for the effect to take place. Otherwise logic that assumes certain ordering won't work (e.g., AsyncEvent waiters, or the peer_pool thing, where it is assumed that the waiters are waked in registration order).

chronos/futures.nim
```diff
@@ -65,6 +71,9 @@ type
       ## a spot in each future for that first one - the seq below will stay
       ## empty until a second callback is added
     internalCallbacks*: seq[InternalAsyncCallback]
+    when chronosSyncContinuations:
+      internalContinuation*: InternalAsyncCallback
+      internalContinuations*: seq[InternalAsyncCallback]
```
arnetheduck, line 76: what's the use case here for multiple continuations?

ie if we posit a future has one continuation (that runs "now") and (potentially) zero or more callbacks (that run "soon"), we can avoid the overhead of these two new fields - the non-seq field was added as an optimization for continuations in particular and doesn't have much utility outside of that use case.

What use cases do we lose? Arguably, it becomes easier to reason about the order of operations when each future has no more than one continuation.

etan-status: It's allowed by the syntax to have multiple `await` on the same future. See the "Multiple waiters test"

chronos/internal/asyncfutures.nim
arnetheduck: Regarding waking in registration order, I wonder if this is a property we want to retain also for continuations. 

as far as implementation goes, one possible way to get there is a `continuations: seq[...]` field next to `callbacks` in the scheduler - this is where all continuations would be added and we would run the continuation loop after each callback call.

Retaining the callback order _seems to_ resolve then eed for the `SyncContinuations` flag which is somewhat confusing (ie the double-gate of "special function to call"+flag and the state-based approach where setting a flag in the constructor changes the behavior of `complete` "later".

chronos/futures.nim
arnetheduck: > It's allowed by the syntax to have multiple await on the same future. See the "Multiple waiters test"

yes - but is this a scenario that we want to cater to? it causes interleaving between those two _regardless_, ie if awaiter 1 changes the state that awaiter 2 is using, we're back to where we started.

arnetheduck: also, to get to the point of having two awaiters, you must already break the await illusion and work with futures directly, putting them in a non-awaited variable and sharing that instance - since the core of the illusion is broken, it's somewhat questionable if the rest of the illusion needs to be maintained.

tests/testcontinuations.nim
```diff
@@ -0,0 +1,371 @@
...
+#              Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#              MIT license (LICENSE-MIT)
+import unittest2
+import ../chronos, ../chronos/config
+
+{.push raises: [], gcsafe.}
+{.used.}
+
+suite "Continuation scheduling test suite":
```
arnetheduck, line 14: the starvation test from https://github.com/nim-lang/Nim/issues/7193#issue-295224413 is interesting to add here - ie a function that `callSoon`s _itself_ should not starve the pipeline in general - injecting continuations is only safe for actual continuations that don't run this risk

chronos/internal/asyncfutures.nim
etan-status: The property is currently retained, continuations run in order of registration, same as callbacks (again the "Multiple waiters test" test, it runs a before b).

The `SyncContinuations` flag is needed, because some futures require specific callback ordering _across_ explicit `addCallback` and `await` users.  That's why the future itself also has to opt-in, otherwise the current behaviour of respecting continuation order isn't preserved in every situation.

This can be observed by running nimbus-eth2 `make -j test` against the earlier version of this PR which dropped all the flags. One could also choose to fix all the clients / tests to support this / not crash etc, but it's an explicit goal to not trigger a cascade of client changes; the default behaviour is only changed where it is safe to do so / where no specific ordering is currently assumed, because the entire future logic is under chronos control.

- 9542f8b has neither flag or 'special function to call'
- f81103e has no flag on the future, but has the 'special function to call'

Both fail the nimbus-eth2 test suite in their own ways.

I agree with the general idea that there _should_ be a cleaner way to achieve this that works more broadly than just the `{.async.} / await` pattern, but I don't see it without requiring explicit opt-ins / opt-outs in externally controlled usage. The current state consolidates callbacks that require `callSoon` with those that don't require it, so there has to be _something_ to distinguish them, and a heuristic can't reliably detect every situation.

chronos/futures.nim
etan-status: > we're back to where we started

no, it's not the same, as the state cannot change between nested awaits. i.e., the state that 'awaiter 2' is using cannot change while the 'awaiter 2' continuations are running, i.e., 'awaiter 1' changes would apply _before_ the first 'awaiter 2' continuation gets executed (the "multiple waiters test" shows both the current and the new order).

> work with futures directly

that's different from _creating_ a new future though. it's working with _existing_ futures, and it's quite common to, for example, use a `race` combinator to wait for a future that is also awaited outside. I'm not sure how prevalent having multiple `await` waiters is, specifically, but I'm not convinced that choosing the `await` syntax should come with more restrictions than choosing the `addCallback` syntax, when one was historically just syntactic sugar around the other.

In any case, I'd also be open to put a `doAssert` if multiple `await` are scheduled concurrently (i.e., multiple continuations get added), but given that the syntax allows it, maybe would be cleaner to treat that as a separate concern, i.e., if that shouldn't be a feature, it shouldn't be one regardless of this PR. This PR should just make sure that nothing that's allowed breaks, it shouldn't introduce new restrictions.

----------

Also, the `internalContinuations` is only used in exceptional cases, if the worry is about `seq` allocation. Another approach could be to choose a `Deque` for both queues, but then the single-item case is a bit tricky (one could choose to _move_ it into the `Deque` as the second callback is added, though, and simply stop using the single-field once there is a second entry). Implementation detail though.

tests/testcontinuations.nim
etan-status: `callSoon` isn't touched though? Or you mean, just as an additional test so that this doesn't get botched accidentally?

And yes, API misuse can't really be prevented with a cooperative scheduler - someone doing `while true` inside an `{.async.}` is as bad as someone using a continuation API for non-continuation, or someone setting up any other deliberate deadlock.

A separate PR should consider looking into optional yield points, and have the inner-most action choose to `await sleepAsync(ZeroDuration)` if no `poll()` happened for ≥ some value like 30ms. e.g., AsyncLock currently has that problem that if there is only a single acquire/release loop, that it simply never yields (concurrent tasks will never be scheduled. But that isn't what this is touching / separate concern.

chronos/internal/asyncfutures.nim
etan-status: Extended test suite so that we cover those ordering scenarios as well.
Note that the new tests share the same order in both `chronosSyncContinuations` and default configs.

tests/testcontinuations.nim
etan-status: Added a test in the bottom, not sure how applicable it is to this PR but doesn't hurt to have it here to avoid regressions.

chronos/internal/asyncfutures.nim
etan-status: Updated the docs

chronos/internal/asyncfutures.nim
arnetheduck: so this is a gap - if we make `callImmediate` private, we must also make the SyncCont flag private, else you can construct a future that calls callImmiedaite indirectly - ie if we want to limit this feature to transformed continuations (which is one strategy to prevent abuse), both the flag and `callImmediate` need this treatment I suspect

arnetheduck: > order

```nim
import chronos

proc waitfor(v: Future[void], b: int) {.async.} =
  await v
  echo "got ", b


proc run() =
  var a = newFuture[void](flags = {FutureFlag.SyncContinuations})
  var b = newFuture[void](flags = {FutureFlag.SyncContinuations})

  asyncSpawn waitfor(a, 1)
  asyncSpawn waitfor(b, 2)

  echo "completing a"
  a.complete()
  echo "completing b"
  b.complete()

  poll()

run()
```

etan-status: the order guarantee is for callbacks on the same future, not for completions across unrelated futures.

SyncCont flag + addCallbackContinuation are the way to not privilege the {.async.}/await syntax. they allow regular future syntax to be used to achieve same outcome. if one botches that, it's a usage error.

callImmediate is not needed to be separately available for that purpose, so that one can be fully private.

etan-status:
Based on PM discussion, this feature needs budgeting in some way to ensure periodic return to the event loop (and also measuring, so that offending tasks are highlighted and can be fixed in a targeted way).

PR #677, opened 2026-07-01
Support for Haiku

korli:
The use of kqueue depends on https://github.com/nim-lang/Nim/pull/25953

### DIFF
--- a/chronos/config.nim
+++ b/chronos/config.nim
@@ -82,7 +82,7 @@ const
       "epoll"
     elif defined(macosx) or defined(macos) or defined(ios) or
           defined(freebsd) or defined(netbsd) or defined(openbsd) or
-          defined(dragonfly):
+          defined(dragonfly) or defined(haiku):
       "kqueue"
     elif defined(posix):
       "poll"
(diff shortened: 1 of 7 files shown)


arnetheduck:
> The use of kqueue depends 

how so? For linking? `{.passl.}` in one of chronos' modules might do the trick (which makes it work for all nim releases)

korli:
> > The use of kqueue depends
> 
> how so? For linking? `{.passl.}` in one of chronos' modules might do the trick (which makes it work for all nim releases)

I now moved the linking in the chronos module. Links correctly.

arnetheduck:
Alright, thanks for the patch - you'll need to keep an eye on it since we don't have CI support for testing haiku but it looks pretty minimal so hopefully won't break

korli:
> Alright, thanks for the patch - you'll need to keep an eye on it since we don't have CI support for testing haiku but it looks pretty minimal so hopefully won't break

I'm watching releases on both repositories. Building at HaikuPorts is likely blocked by https://github.com/nim-lang/Nim/issues/22590. Will have to workaround it seems.

PR #678, opened 2026-07-02
Add helper for parsing Authorization headers

etan-status:
Clients such as nimbus-eth2 rest_key_management_api currently hand-parse the HTTP Authorization header, and do so incompletely (scheme should be case-insensitive according to HTTP specs). Providing a helper enables migration to correct usage.

### DIFF
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -11,7 +11,7 @@ requires "nim >= 1.6.16",
          "results",
          "stew >= 0.5.0",
          "bearssl >= 0.2.8",
-         "httputils",
+         "httputils >= 0.4.2",
          "unittest2"
 
 import os, strutils
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -11,7 +11,7 @@ requires "nim >= 1.6.16",
          "results",
          "stew >= 0.5.0",
          "bearssl >= 0.2.8",
-         "httputils >= 0.4.2",
+         "httputils >= 0.4.3",
          "unittest2"
 
 import os, strutils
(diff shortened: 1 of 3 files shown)


arnetheduck:
https://github.com/status-im/nim-http-utils/blob/master/httputils.nim might be the better place for this?

etan-status:
https://github.com/status-im/nim-http-utils/pull/45

arnetheduck:
no, I mean the full parser - seems that httputils is the place for http parsers

etan-status:
Ahhh, yeah makes sense:

- https://github.com/status-im/nim-http-utils/pull/47

And then in chronos we would only have a thin wrapper around http-utils similar to `getContentType`.

arnetheduck:
can't make up my mind about this one ... httpcommon is more like an internal helper library for the features offered by server/client in general .. should authorization instead be exposed as a field similar to the content type or such? ie then the server could automatically reject invalid requests even before they get to the stage where the "consumer" uses them

etan-status:
If you prefer, can also move the remaining logic from this PR directly into nimbus-eth2. Just let me know your preference :-)

arnetheduck:
> If you prefer, can also move the remaining logic from this PR directly into nimbus-eth2. Just let me know your preference :-)

well, I think either make it fully implemented on the chronos side (ie add  the authorization as a field and reject the request altogether if erroneous, similar to how a parsing failure would fail) or do it in nimbus-eth2 for now - else we have this dangling helper in httpcommon which I'm not a fan of

etan-status:
Superseded by https://github.com/status-im/nimbus-eth2/pull/8718

PR #679, opened 2026-07-02
avoid cyclic reference in `race`/`or`/`one` (fixes #676)

### LINKED ISSUE #676: Regression using `chronos/streams/tlsstream` with `--mm:arc`
```
import chronos/streams/tlsstream
discard handshake(TLSAsyncStream(nil))
```
creates a compile-time warning using `--mm:arc`
```
nim-chronos/chronos/internal/asyncfutures.nim(1076, 6) Warning: ':env.cb`gensym602 = (:anonymous_2550139136, :env)' creates an uncollectable ref cycle; annotate 'cb`gensym602' with .cursor [CycleCreated]
```
introduced by:
- #666

Before that, it was clean.

arnetheduck:
although the cycle was being manually broken, it still caused a warning

### DIFF
--- a/chronos/internal/asyncfutures.nim
+++ b/chronos/internal/asyncfutures.nim
@@ -734,7 +734,6 @@ template orImpl*[T, Y](fut1: Future[T], fut2: Future[Y]): untyped =
           retFuture.fail(fut.error, warn = false)
         else:
           retFuture.complete()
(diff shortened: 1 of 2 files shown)


arnetheduck:
https://github.com/nim-lang/Nim/issues/25963

PR #680, opened 2026-07-06
Fix getAddress deprecated warning

moigagoo:
While migrating [Nimble to Chronos](https://github.com/nim-lang/nimble/pull/1746), I bumped into a couple of these warnings:

```
/home/moigagoo/projects/nimble/vendor/chronos/chronos/apps/http/httpclient.nim(1052, 27) Warning: use getHttpAddress; getAddress is deprecated [Deprecated]
```

This PR replaces getAddress with getHttpAddress.

### DIFF
--- a/chronos/apps/http/httpclient.nim
+++ b/chronos/apps/http/httpclient.nim
@@ -438,13 +438,13 @@ proc redirect*(srcuri, dsturi: Uri): Uri =
     combine(srcuri, tmpuri)
 
 proc redirect*(session: HttpSessionRef,
-               srcaddr: HttpAddress, uri: Uri): HttpResult[HttpAddress] =
+               srcaddr: HttpAddress, uri: Uri): HttpAddressResult =
   ## Transform original address ``srcaddr`` using redirected url ``uri`` and
   ## session ``session`` parameters.
   let srcuri = srcaddr.getUri()
   var newuri = srcuri.redirect(uri)
   if newuri.hostname != srcuri.hostname:
-    session.getAddress(newuri)
+    session.getHttpAddress(newuri)
   else:
     let scheme =
       case newuri.scheme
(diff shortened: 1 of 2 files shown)


chronos/apps/http/httpclient.nim
```diff
@@ -438,13 +438,13 @@ proc redirect*(srcuri, dsturi: Uri): Uri =
     combine(srcuri, tmpuri)
 
 proc redirect*(session: HttpSessionRef,
-               srcaddr: HttpAddress, uri: Uri): HttpResult[HttpAddress] =
+               srcaddr: HttpAddress, uri: Uri): HttpAddressResult =
```
arnetheduck: we generally avoid breaking changes in the public api

moigagoo: Got it. I reverted the return type to HttpResult and added explicit conversion from HttpAddressErrorType to string in `redirect`.

chronos/apps/http/httpclient.nim
```diff
@@ -1049,7 +1051,7 @@ proc new*(t: typedesc[HttpClientRequestRef], session: HttpSessionRef,
           maxResponseHeadersSize: int = HttpMaxHeadersSize,
           headers: openArray[HttpHeaderTuple] = [],
           body: openArray[byte] = []): HttpResult[HttpClientRequestRef] =
-  let address = ? session.getAddress(parseUri(url))
+  let address = ? session.getHttpAddress(parseUri(url))
```
arnetheduck, line 1054: This leads to a compile failure at generic instantiation time (and is not covered by the test suite, apparently)

arnetheduck: When compiling nimbus-eth1:

results.nim(474, 42) Error: type mismatch: got 'HttpAddressErrorType' for 'v`gensym97.eResultPrivate' but expected 'string'

moigagoo: Got it. Error must be converted to string. Did it in another place, repeated here.

Also added the test cases to cover that.

PR #683, opened 2026-07-07
avoid `case` for httpversion

arnetheduck:


arnetheduck:
see also: https://github.com/status-im/nim-http-utils/pull/49

notably, this branch works with the previous HttpVersion incarnation as well

PR #689, opened 2026-07-09
Add `chronosStrictReentrancy` for runtime reentrancy detection

arnetheduck:
Reentrancy causes event reordering and stack explosions, in addition to leading to hard-to-debug scenarios.

Since none of chronos is actually tested under reentrant conditions (ie that all features such as exceptions, cancellations, buffer operations, timers etc work reliably when the loop is reentered), this is hard to support over time and prevents useful optimizations - this PR simply detects and disallows the behaviour to remove the uncertainty, simplifying reasoning about the event loop in general.

### DIFF
--- a/chronos/internal/asyncengine.nim
+++ b/chronos/internal/asyncengine.nim
@@ -160,7 +160,7 @@ template processCallbacks(loop: untyped) =
   when chronosStrictReentrancy:
     # Process existing callbacks but not those that follow, to allow the network
     # to regain control regularly
-    for _ in 0..<loop.callbacks.len():
+    for _ in 0 ..< len(loop.callbacks):
       let callable = loop.callbacks.popFirst()
       if not(isNil(callable.function)):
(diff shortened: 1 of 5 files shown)


chronos/internal/asyncengine.nim
```diff
@@ -143,12 +157,20 @@ template processTicks(loop: untyped) =
...
-  while true:
-    let callable = loop.callbacks.popFirst()  # len must be > 0 due to sentinel
-    if isSentinel(callable):
-      break
-    if not(isNil(callable.function)):
-      callable.function(callable.udata)
+  when chronosStrictReentrancy:
+    # Process existing callbacks but not those that follow, to allow the network
+    # to regain control regularly
+    for _ in 0..<loop.callbacks.len():
```
cheatfate:
```suggestion
    for _ in 0 ..< len(loop.callbacks):
```

PR #695, opened 2026-07-17
ci(mobile): run test suite on Android and iOS

richard-ramos:
Add dedicated GitHub Actions workflows that cross-compile and run `nimble test` on:

- Android x86_64 using an emulator
- iOS arm64 using a simulator

The Nimble test task now supports an optional external test runner so mobile binaries can be executed through `adb` or `simctl`. Mobile-specific test handling avoids host-only paths and excludes process tests that depend on host-side helper scripts.

### DIFF
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   test-android:
-    name: Android x86_64 tests
+    name: Android x86_64 tests (Nim version-2-2)
     runs-on: ubuntu-22.04
     timeout-minutes: 60
 
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -26,6 +26,7 @@ jobs:
         uses: jiro4989/setup-nim-action@v2
         with:
           nim-version: "2.2.x"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install dependencies
         run: nimble --useSystemNim install -y --depsOnly
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -1,0 +1,57 @@
+name: Mobile Android
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-android:
+    name: Android x86_64 tests
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install Nim
+        uses: jiro4989/setup-nim-action@v2
+        with:
+          nim-version: "2.2.x"
+
+      - name: Install dependencies
+        run: nimble --useSystemNim install -y --depsOnly
+
+      - name: Configure Android toolchain
+        run: |
+          toolchain="$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin"
+          echo "NIMFLAGS=--os:android --cpu:amd64 --cc:clang --clang.path:$toolchain --clang.exe:x86_64-linux-android23-clang --clang.linkerexe:x86_64-linux-android23-clang --passL:-llog" >> "$GITHUB_ENV"
+
+      - name: Enable KVM for Android emulator
+        run: |
+          sudo tee /etc/udev/rules.d/99-kvm4all.rules >/dev/null <<'EOF'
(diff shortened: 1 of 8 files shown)


tests/teststream.nim
```diff
@@ -51,6 +58,11 @@ suite "Stream Transport test suite":
       initTAddress("127.0.0.1:33335"),
       initTAddress(r"/LOCAL\testpipe")
     ]
+  elif defined(android):
+    let addresses = [
+      initTAddress("127.0.0.1:0"),
+      initTAddress(getAppDir() / "testpipe")
+    ]
```
richard-ramos, line 65: Android cannot use the host /tmp path. This uses a path inside the app directory for the Unix socket, and port 0 lets Android choose a free TCP port.

tests/teststream.nim
```diff
@@ -1328,7 +1340,7 @@ suite "Stream Transport test suite":
       check ok
 
     asyncTest prefixes[i] & "accept() too many file descriptors test":
-      when defined(windows):
+      when defined(windows) or defined(android):
```
richard-ramos, line 1343: Android has a lower open-file limit. This test reaches the system limit before it can complete its check, so we skip it on Android as we already do on Windows.

chronos.nimble
```diff
@@ -45,7 +47,11 @@ proc build(args, path: string) =
 
 proc run(args, path: string) =
   build args, path
-  exec "build/" & path.splitPath[1]
+  let executable = "build/" & path.splitPath[1]
+  if testRunner.len == 0:
+    exec executable
+  else:
+    exec testRunner & " " & quoteShell(executable)
```
richard-ramos, line 56: Desktop tests still run the binary directly. Mobile CI provides a runner that copies and starts the binary inside the emulator or simulator.

chronos.nimble
```diff
@@ -21,6 +21,8 @@ let lang = getEnv("NIMLANG", "c") # Which backend (c/cpp/js)
 let flags = getEnv("NIMFLAGS", "") # Extra flags for the compiler
 let verbose = getEnv("V", "") notin ["", "0"]
 let platform = getEnv("PLATFORM", "")
+let testRunner = getEnv("NIM_TEST_RUNNER", "")
+let testSuccessMarker = getEnv("NIM_TEST_SUCCESS_MARKER", "")
```
richard-ramos, line 26: These optional variables let mobile CI choose how to run each test binary and record when the full suite finishes. Empty values keep the normal desktop behavior.

tests/teststream.nim
```diff
@@ -850,9 +862,9 @@ suite "Stream Transport test suite":
           var transp = await connect(address)
           var ssize: string
           var handle = 0
-          var name = FilesTestName
-          var size = int(getFileSize(FilesTestName))
-          var fhandle = open(FilesTestName)
+          var name = filesTestName()
+          var size = int(getFileSize(filesTestName()))
+          var fhandle = open(filesTestName())
```
richard-ramos, line 867: The source file path exists on the CI host, but not inside the mobile simulator. filesTestName() returns a local file that the mobile test can open.

.github/workflows/mobile_android.yml
```diff
@@ -0,0 +1,65 @@
...
+          api-level: 23
+          arch: x86_64
+          target: default
+          force-avd-creation: false
+          emulator-options: -no-window -no-audio -no-boot-anim -no-metrics -gpu swiftshader_indirect
+          disable-animations: true
+          script: |
+            rm -f "$NIM_TEST_SUCCESS_MARKER"
+            nimble --useSystemNim test
+            test -f "$NIM_TEST_SUCCESS_MARKER"
```
richard-ramos, line 66: Nimble can return success when its task stops early. We remove the old marker before the run, and the test task creates it only after the full suite finishes. The final check makes the workflow fail if compilation or tests stop early.

tests/teststream.nim
arnetheduck: anything preventing 0 from being used throughout? we've had issues with busy ports on CI hosts occasionally

arnetheduck (approved):

.github/workflows/mobile_android.yml
arnetheduck: > Nimble can return success when its task stops early. 

is this a nimble bug? there was a similar one in one of the releases, don't remember which

richard-ramos: It is a nimble bug of version `0.22.2` which is what gets installed when using `nim-version: "2.2.x"` (which resolves to `2.2.10`). This was indeed fixed in `0.24.0`, but `2.2.x` does not have that fix yet. 

Once it does, i can drop the `NIM_TEST_SUCCESS_MARKER`

tests/teststream.nim
richard-ramos: Is this wrt the :33335 port used on windows in line 51? If so, this existed previous to my PR. 
I imagine it should be fine to change it to `:0` (also in `testdatagram.nim`, `testhttpserver.nim`, `testserver.nim`, `testproc.nim` and `teststream.nim`), but should probably be done on a separate PR

PR #697, opened 2026-07-21
Support compiling timer on bare metal

kdeme:
Guard the osdefs import for any/standalone and stub the clock to 0

### DIFF
--- a/chronos/timer.nim
+++ b/chronos/timer.nim
@@ -23,7 +23,9 @@
 ##
 ## You can specify which timer you want to use ``-d:asyncTimer=<system/mono>``.
 import stew/base10
-import "."/osdefs
+
+when not (defined(`any`) or defined(standalone)):
+  import "."/osdefs
 
 const asyncTimer* {.strdefine.} = "mono"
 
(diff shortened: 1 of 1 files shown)


chronos/timer.nim
```diff
@@ -136,6 +138,11 @@ elif defined(posix):
 
 elif defined(nimdoc):
   discard
+elif defined(`any`) or defined(standalone):
+  # The other branches use OS specific clock primitives, unavailable under
+  # any/standalone, so stub to 0. Real time could still be added via the
+  # portable C11 `timespec_get`.
+  proc fastEpochTimeNano(): uint64 {.inline.} = 0
```
kdeme, line 145: If preferred, we could also try to compile flag guard the `now` and `fromNow` calls instead. This version however seemed clean as there is already the `when/elif/elif...`. But it does causes a `now` to exist yet dysfunct (however, not that uncommon for clock usage on embedded targets, so devs would probably realize quickly?).

From what I've seen, these (`now`, `fromNow`) are not used via the imports where I'm hitting the issue (`spec/presets` and `spec/beacon_time` via `datatypes.base` in nimbus-eth2), so that change might be possible if wanted, but I'd have to verify if that works.

PR #698, opened 2026-07-22
perf(http): cut per-request allocs (buffer memset, header keys, empty tables

jmgomez:
image


chronos/apps/http/httptable.nim
```diff
@@ -20,18 +20,30 @@ type
...
+  ht.table.mgetOrPut(lowerKey(key), default).add(value)
 
 proc add*(ht: var HttpTables, key: string, value: SomeInteger) =
   ## Add integer ``value`` to header with key ``key``.
   ht.add(key, $value)
 
 proc set*(ht: var HttpTables, key: string, value: string) =
   ## Set/replace value of header with key ``key`` to value ``value``.
-  let lowkey = key.toLowerAscii()
+  let lowkey = lowerKey(key)
```
arnetheduck: this will still perform several unnecesary allocations - what you're looking for is:
```nim
proc replaceWithOne(v: var seq[string], s: string) =
  v.setLen(1)
  v[0] = s
ht.table[lowerKey(...)].replaceWithOne()
```

the values here should also use `chronosSink` and `move` into the seq

chronos/apps/http/httptable.nim
```diff
@@ -20,18 +20,30 @@ type
...
+  for ch in s:
+    if ch in {'A' .. 'Z'}:
+      return true
+  false
+
+template lowerKey(key: string): string =
+  ## Header keys are normalized to lowercase for case-insensitive storage.
+  ## Most keys (all chronos-internal constants and most real-world traffic)
+  ## are already lowercase — avoid allocating a lowered copy in that case.
+  if hasUpperAscii(key): key.toLowerAscii() else: key
```
arnetheduck: this still performs a copy in refc:
```nim
import std/strutils

func hasUpperAscii(s: string): bool =
  for ch in s:
    if ch in {'A' .. 'Z'}:
      return true
  false

template lowerKey(key: string): string =
  if hasUpperAscii(key): key.toLowerAscii() else: key

proc myfunc(v: string) =
  debugEcho repr(cast[pointer](addr v[0]))

proc testit() =
  var aa = "a"
  var bb = "A"

  debugEcho repr(cast[pointer](addr aa[0])), " ",repr(cast[pointer](addr bb[0]))
  myfunc(lowerKey(aa))
  myfunc(lowerKey(bb))

testit()
```

chronos/apps/http/httpserver.nim
```diff
@@ -466,7 +466,9 @@ proc updateRequest*(request: HttpRequestRef, scheme: string, meth: HttpMethod,
 
   # Conversion of request query string to HttpTable.
   request.query =
-    block:
+    if len(request.uri.query) == 0:
```
arnetheduck: why is this needed? given that the for loop below _should_ do nothing, the code should reduce to the same runtime behavior

chronos/apps/http/httpserver.nim
```diff
@@ -805,6 +807,20 @@ proc preferredContentType*(request: HttpRequestRef,
   ## client in request ``request``.
   preferredContentType(request.headers.getString(AcceptHeaderName), types)
 
+var
+  gCachedDate {.threadvar.}: string
+  gCachedDateMoment {.threadvar.}: Moment
+
+proc cachedHttpDate(): string =
```
arnetheduck:
```suggestion
proc cachedHttpDate(): lent string =
```

without lent, a copy is performed on return here

chronos/apps/http/httpserver.nim
```diff
@@ -904,11 +918,16 @@ proc getRequest(conn: HttpConnectionRef): Future[HttpRequestRef] {.
...
+    # costs a full memset of maxHeadersSize on EVERY request. Parse only the
+    # valid slice instead; makeCopy copies just the actual request bytes.
+    if len(conn.buffer) < conn.server.maxHeadersSize:
+      conn.buffer.setLen(conn.server.maxHeadersSize)
     let res = await conn.reader.readUntil(addr conn.buffer[0], len(conn.buffer),
                                           HeadersMark)
-    conn.buffer.setLen(res)
-    let header = parseRequest(conn.buffer)
+    let header = parseRequest(
+      conn.buffer.toOpenArray(0, res - 1), makeCopy = true)
```
arnetheduck: it might be that we don't need the makecopy here - afair, parseRequest will store offsets into the buffer so as long as the buffer is not touched, its length should really not matter

chronos/apps/http/httpserver.nim
```diff
@@ -977,6 +999,23 @@ proc closeWait*(req: HttpRequestRef) {.async: (raises: []).} =
     reset(req[])
     req.state = HttpState.Closed
 
+proc tryCloseSync(req: HttpRequestRef): bool =
+  ## Synchronous equivalent of HttpRequestRef.closeWait() for requests whose
+  ## response has no streaming writer — on that path closeWait() never
+  ## actually suspends, so the future allocation is pure overhead.
+  ## Returns false when the async variant is required.
+  if req.state == HttpState.Alive:
```
arnetheduck: instead of repeating this code, this should be refactored to something like:

```
template closeImpl(..) =
  if ...
``
```

then the template can be inlined in both the public `closeWait` and the other places - this should achieve the same goal but with less future maintenance - the state management is hairy enough as it is.

PR #672, opened 2026-06-24
Always report Event.Error to both read/write callbacks

etan-status:
When `Event.Error` is paired with additional flags, it does not always wake up the reader / writer callbacks. For example, `{Read, Error}` does not wake up the writer callack; however, standalone `{Error}` does.

Not issuing both callbacks can leak futures, e.g., the `writeStreamLoop` would never complete or fail as it is never resumed.

PR #700, opened 2026-07-24
Revert "Always report Event.Error to both read/write callbacks"

etan-status:
#672 has a regression, triggering a latent bug on kqueue (macOS / BSD) backend, as kqueue does not deduplicate events per fd, i.e., {Read, Error} and {Write, Error} are separately reported, leading to the read callback being called multiple times. If the first read callback unregisters the reader, the second copy will still execute, triggering downstream problems.

This PR initially tried to solve the footgun (deduplicate in the kqueue engine), then tried to fix the failing caller (stream.nim `if {TransportState.Closed, ReadPaused} * transp.state == {}:`), but now that the blame is shifted to #672 it's better to revert that one (restoring the flaky CI that it attempted to fix), and rely on the kqueue backend never emitting {Error} alone (so the callback is not called spuriously anymore).

A proper fix can be applied later, for now let's just get back to stability.


arnetheduck:
> assumes that events only trigger callback once.

there's no such guarantee - ie users have to account for spurious wakeups - consider what happens when two `fire` happen async-concurrently - in some cases, the second fire will be suppressed and in others no - depending on scheduling, ie order of events, that will result in either one wakeup or two and you cannot, as a caller of `fire`, which will happen.

This is a general signalling problem, ie either signals of this kind _count_ the number of fires and issue a corresponding amount of wakeups or they do this kind of coalescing behavior .. see https://man7.org/linux/man-pages/man2/eventfd.2.html and EFD_SEMAPHORE.

etan-status:
Without the fix, this one on macOS:

```sh
nim c -r tests/testbugs "Reader notification after buffer got full [poll()] test"
```

-->

```
Asynchronous issues test suite F (0.11s)
=======================================================
  vendor/nim-chronos/tests/testbugs 'Asynchronous issues test suite::Reader notification after buffer got full [poll()] test'
-------------------------------------------------------
vendor/nim-unittest2/unittest2.nim(1175) testbugs
vendor/nim-unittest2/unittest2.nim(1095) unittest2::runDirect(unittest2::Test)
chronos/unittest2/asynctests.nim(17) testbugs::runTestX60gensym994_(string, string)
chronos/internal/asyncfutures.nim(714) asyncfutures::waitFor(Future<void>)
chronos/internal/asyncfutures.nim(689) asyncfutures::pollFor(Future<void>)
chronos/internal/asyncengine.nim(179) asyncengine::poll
chronos/transports/stream.nim(1475) stream::readerCb(pointer)
chronos/transports/stream.nim(1426) stream::readIntoBuffer(ref<stream::StreamTransportcolonObjectType_>)
vendor/nim-chronos/chronos/bipbuffer.nim bipbuffer::reserve(var<bipbuffer::BipBuffer>, range09223372036854775807)
vendor/nimbus-build-system/vendor/Nim/lib/std/assertions.nim(40) assertions::raiseAssert(string)
vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(62) assertions::sysFatal(typeDesc<exceptions::AssertionDefect>, string)

    Unhandled defect: Not enough space available [AssertionDefect]

  [FAILED ] (  0.11s) Reader notification after buffer got full [poll()] test

(0.11s)   Asynchronous issues test suite

[Summary] 1 tests run (0.11s): 0 OK, 1 FAILED, 0 SKIPPED
Error: execution of an external program failed: 'vendor/nim-chronos/tests/testbugs 'Reader notification after buffer got full [poll()] test''
```

etan-status:
This is not addressing spurious wakeups across different poll invocations, i.e., keep getting asked to provide data as long as data can be sent, but instead tackles the way how removeReader2 behaves. As the engine queues all the callbacks, if there are multiple callbacks for the same event, a removeReader2 within the first one won't cancel additional ones.

So if we want to have removeReader2 apply immediately, it's something that the engine itself has to provide (or all callbacks need logic to handle the case where the engine still calls them even after they were removed).

etan-status:
With kqueue, essentially, {Read, Error} + {Write, Error} become two events, even if they both become ready in the same kqueue call.

PR #701, opened 2026-08-03
Docs: Add links to the app module API ref.

moigagoo:
It was bugging me that I couldn't find the API docs for the HTTP apps quickly.

### DIFF
--- a/docs/src/introduction.md
+++ b/docs/src/introduction.md
@@ -46,5 +46,14 @@ Several platforms are supported, with different backend [options](./concepts.md#
 
 ## API documentation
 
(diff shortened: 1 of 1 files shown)


arnetheduck:
this is fine, but it would also be nice to do something about the http tutorial menu which right now is unproportionally large - ie it's hard to find the "User guide" section - this can be fixed by moving the tutorial lower (under the user guide) or making the chapters expandable so that the `.1` etc are not visible by default

image

moigagoo:
>  this can be fixed by moving the tutorial lower (under the user guide) or making the chapters expandable

image

https://github.com/status-im/nim-chronos/pull/704

PR #707, opened 2026-08-18
newer nimble is broken; use the older one

narimiran:


arnetheduck:
cc @jmgomez

jmgomez:
@narimiran where it fails? Any link with the logs? I assume you referring to `0.24.1`, right?

arnetheduck:
https://github.com/status-im/nim-chronos/actions/runs/32142697011/job/95728931153

jmgomez:
Turns out the issue is in chronos itself. It only worked by chance due to a bug in old nimble: it installed the root package (chronos) into pkgs2, and the examples were silently compiling against that stale installed copy instead of the checkout. 

See https://github.com/status-im/nim-chronos/pull/708 for the required change.

PR #711, opened 2026-08-24
specify the versions of dependencies

narimiran:
Otherwise, nimble's `--resolver:minver` picks up too low versions, which don't work.

### DIFF
--- a/chronos.nimble
+++ b/chronos.nimble
@@ -2,7 +2,7 @@ mode = ScriptMode.Verbose
 
 packageName   = "chronos"
 # keep in sync: chronos/apps/http/httpagent.nim
-version       = "4.4.1"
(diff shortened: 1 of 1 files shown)


chronos.nimble
```diff
@@ -2,18 +2,18 @@ mode = ScriptMode.Verbose
 
 packageName   = "chronos"
 # keep in sync: chronos/apps/http/httpagent.nim
-version       = "4.4.0"
+version       = "4.4.1"
```
arnetheduck:
```suggestion
version       = "4.4.0"
```

I think we might branch off 4.4.1 from an earlier commit, the current master is a bit wonky and we need to selectively backport a few things - let's add the limits first

PR #718, opened 2026-08-26
simplify lock/semaphore

arnetheduck:
`AsyncLock` is a special case of `AsyncSemaphore` (with size 1) - make sure that they both behave the same.

In particular, simpify the `AsyncLock` implementation to use fewer poll rounds and futures during `acquire` while maintaining the same fairness guarantees as before.

Also, add `tryAcquire` and `tryRelease` which are exception-free variants of the existing functions - in particular, `release` is annoying to use in `finally` with `raises: []` since it then requires an exception handler that never gets triggered.

* release cancelled futures during `acquire` as well as `release`.

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -27,7 +27,6 @@ type
     ## ``release()`` call resets the state to unlocked; first coroutine which
     ## is blocked in ``acquire()`` is being processed.
     locked: bool
-    acquired: bool
     waiters: Deque[Future[void].Raising([CancelledError])]
 
   AsyncEvent* = ref object of RootRef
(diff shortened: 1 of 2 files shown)


chronos/asyncsync.nim
```diff
@@ -107,37 +106,48 @@ proc newAsyncLock*(): AsyncLock =
...
+  else:
+    # Drop cancelled waiters
+    while lock.waiters.len > 0 and lock.waiters[0].finished:
+      discard lock.waiters.popFirst()
+
+    lock.waiters.addLast(fut)
+
+  fut
+
+proc tryRelease*(lock: AsyncLock): bool =
```
etan-status, line 136:
```suggestion
proc tryRelease*(lock: AsyncLock): bool {.discardable.} =
```

etan-status: But in a way, discarding it isn't safe either. If there are logic bugs in the code, it just hides them...

PR #721, opened 2026-08-28
simplify lock/semaphore, deprecate `AsyncLockError`/`AsyncSemaphoreError`

arnetheduck:
`AsyncLock` is a special case of `AsyncSemaphore` (with size 1) - make sure that they both behave the same.

In particular, simpify the `AsyncLock` implementation to use fewer poll rounds and futures during `acquire` while maintaining the same fairness guarantees as before.

At the same time, deprecate `AsyncLockError` and `AsyncSemaphoreError` and introduce `release2` that panics instead of raising a checked exception when trying to release a lock / semaphore that is not held - this behavior matches `Lock` in the stdlib and avoids an annoying `try/except` in `finally` for code that uses `raises: []`.

`release2` is added as an interim solution for code that wants to avoid both warning and exception handler while waiting for a new chronos major release that changes the existing `release` function to the new behavior.

### DIFF
--- a/chronos/asyncsync.nim
+++ b/chronos/asyncsync.nim
@@ -17,6 +17,8 @@ import ./asyncloop
 export asyncloop
 
 type
+  Waiter = Future[void].Raising([CancelledError])
+
   AsyncLock* = ref object of RootRef
     ## A primitive lock is a synchronization primitive that is not owned by
     ## a particular coroutine when locked. A primitive lock is in one of two
@@ -27,7 +29,7 @@ type
     ## ``release()`` call resets the state to unlocked; first coroutine which
     ## is blocked in ``acquire()`` is being processed.
     locked: bool
-    waiters: Deque[Future[void].Raising([CancelledError])]
+    waiters: Deque[Waiter]
 
   AsyncEvent* = ref object of RootRef
     ## A primitive event object.
@@ -40,7 +42,7 @@ type
     ## state to be signaled, when event get fired, then all coroutines
     ## continue proceeds in order, they have entered waiting state.
     flag: bool
-    waiters: seq[Future[void].Raising([CancelledError])]
+    waiters: seq[Waiter]
 
   AsyncQueue*[T] = ref object of RootRef
     ## A queue, useful for coordinating producer and consumer coroutines.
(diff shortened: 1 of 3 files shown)


chronos/asyncsync.nim
```diff
@@ -107,60 +106,77 @@ proc newAsyncLock*(): AsyncLock =
...
+  ## Release the lock, allowing one of the waiters to acquire it.
+  ##
+  ## Releasing a lock that is not held is undefined behavior and may
+  ## result in a Defect being raised.
+  ##
+  ## `release2` is an interim interim solution to avoid breaking
+  ## existing code that uses `release` and `AsyncLockError`.
+  ## A future major chronos release will update `release` to behave
+  ## like `release2` and remove the `AsyncLockError` type.
+  assert lock.locked, "Lock is not acquired"
```
nitely: may break Nim CI since it tests in release mode. https://github.com/nim-lang/Nim/blob/c87926dadf4b2b3a9f274b607a61d2532b9e05b9/testament/important_packages.nim#L50

Also any reason why this is not a doAssert?

chronos/asyncsync.nim
```diff
@@ -671,18 +689,42 @@ proc tryAcquire*(s: AsyncSemaphore): bool =
...
+  ## Release a resource, allowing one of the waiters to acquire it.
+  ##
+  ## Releasing more resources than have been acquired is undefined behavior
+  ## and may result in a Defect being raised.
+  ##
+  ## `release2` is an interim interim solution to avoid breaking
+  ## existing code that relies on `AsyncSemaphoreError`.
+  ## A future major chronos release will update `release` to behave
+  ## like `release2` and remove the `AsyncSemaphoreError` type.
+  assert s.availableSlots < s.size
```
nitely: ditto

chronos/asyncsync.nim
```diff
@@ -646,18 +662,20 @@ proc waitEvents*[T](ab: AsyncEventQueue[T],
 
   events
 
-proc newAsyncSemaphore*(size: int = 1): AsyncSemaphore =
+proc newAsyncSemaphore*(size: int = 1, initial = int.high()): AsyncSemaphore =
```
nitely: initial not used?

chronos/asyncsync.nim
```diff
@@ -107,60 +106,77 @@ proc newAsyncLock*(): AsyncLock =
...
+      return
+
+  lock.locked = false
 
 proc locked*(lock: AsyncLock): bool =
   ## Return `true` if the lock ``lock`` is acquired, `false` otherwise.
   lock.locked
 
+{.push warning[Deprecated]: off.}
 proc release*(lock: AsyncLock) {.raises: [AsyncLockError].} =
```
nitely, line 175: I think it is probably better to deprecate this function. What if the user does not use AsyncLockError to catch errors? ie: no AsyncLockError deprecation warning.

arnetheduck: the aim is to get to a world where `release` remains the canonical name for this function is the same as for the regular.. it's a good point about never catching the error, but the runtime defect will have to do if someone is using release in this strange way.

chronos/asyncsync.nim
arnetheduck: oops, that's the next pr :)

chronos/asyncsync.nim
```diff
@@ -107,60 +135,123 @@ proc newAsyncLock*(): AsyncLock =
...
+proc acquire*(
+    s: AsyncSemaphore
+): Future[void] {.async: (raises: [CancelledError], raw: true).} =
+  ## Acquire a resource and decrement the resource counter.
+  ## If no more resources are available, the returned future
+  ## will not complete until the resource count goes above 0.
+
+  acquireImpl(s, s.tryAcquire())
+
+proc release2*(s: AsyncSemaphore) =
```
vladopajic, line 225: perhaps `releaseUnchecked` as name?

arnetheduck: this name will go away and become `release` in the next (major) release - it's just a temporary thing

PR #723, opened 2026-08-31
Fix `RangeDefect` when requesting large response buffer

etan-status:
Large uint value doesn't necessarily fit into int

### DIFF
--- a/chronos/streams/boundstream.nim
+++ b/chronos/streams/boundstream.nim
@@ -56,6 +56,9 @@ template newBoundedStreamIncompleteError(): ref BoundedStreamError =
 template newBoundedStreamOverflowError(): ref BoundedStreamOverflowError =
   newException(BoundedStreamOverflowError, "Stream boundary exceeded")
 
+func numRemainingBytes(rstream: BoundedStreamReader): int =
+  int(min(rstream.boundSize.get() - rstream.offset, uint64(high(int))))
+
 proc readUntilBoundary(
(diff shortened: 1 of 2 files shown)


chronos/streams/boundstream.nim
```diff
@@ -171,16 +174,21 @@ proc readBounded(
...
-      raise exc
-    except AsyncStreamIncompleteError:
-      raise newBoundedStreamIncompleteError()
+  let n = rstream.numRemainingBytes()
+  var res = newSeqUninit[byte](min(n, 128 * 1024))
+  if n > 0:
+    var o = 0
+    while o < n:
+      if o == res.len:
+        res.setLenUninit(int(min(uint64(n), uint64(res.len) shl 1)))
```
arnetheduck:
```suggestion
        res.setLenUninit(min(n, res.len shl 1))
```
why the uint?

arnetheduck: oh, so that we can do 2gb-1 bytes on 32-bit? mhmm

etan-status: yeah, so it doesn't overflow to 0

etan-status:
no, there's a `min` clamping it to `n` (which itself is clamped to `int.high`). no overflow

chronos/streams/boundstream.nim
etan-status: simplified it for clarity

PR #726, opened 2026-09-02
Fix multipart boundary parsing when trailing \r\n missing

etan-status:
Avoid IndexDefect when boundary \r\n is incomplete in multipart message.

### DIFF
--- a/tests/testhttpserver.nim
+++ b/tests/testhttpserver.nim
@@ -591,8 +591,9 @@ suite "HTTP server testing suite":
       for o in 0 .. 2:  # Incomplete trailing \r\n
         var mpreader = MultiPartReader.init(
           body.toOpenArray(0, body.high - o), boundary)
-        for (name, value) in [
-            ("key1", "value1"), ("key2", "value2"), ("key2", "value4")]:
+        const ExpectedParts = [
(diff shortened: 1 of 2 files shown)


tests/testhttpserver.nim
```diff
@@ -560,29 +560,48 @@ suite "HTTP server testing suite":
...
       await server.closeWait()
-      return serverRes and (data.find(expect) >= 0)
+
+      var bufferRes = true
+      for o in 0 .. 2:  # Incomplete trailing \r\n
+        var mpreader = MultiPartReader.init(
+          body.toOpenArray(0, body.high - o), boundary)
+        const ExpectedParts = [
+          ("key1", "value1"), ("key2", "value2"), ("key2", "value4")]
+        for (name, value) in ExpectedParts:
```
etan-status, line 596: If this is inlined, CI produces:

```
[GCASSERT] incRef: interiorPtrTraceback (most recent call last)
```

with separate const it goes away.

What's weird is that running it locally doesn't produce this assert.

etan-status: https://github.com/nim-lang/Nim/issues/26158

tersec: Fixed for Nim 2.2.14.

PR #727, opened 2026-09-02
Retain error on failed stream while error condition holds

etan-status:
Currently, reading from a stream is a SIGSEGV if it reported an error. For example, this is not valid.

```nim
    for _ in 0 ..< 3:
      try:
        var buffer: array[16, byte]
        discard await transp.readOnce(addr buffer[0], len(buffer))
      except TransportOsError:
        discard
```

Retain the error object while we are in the error condition and re-raise it on each use.

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -2545,6 +2545,7 @@ template readLoop(name, body: untyped): untyped =
     if len(transp.buffer) == 0:
       # We going to raise an error, only if transport buffer is empty.
       if ReadError in transp.state:
+        doAssert(not(isNil(transp.error)), "Attempt to read data after error")
         raise transp.getError()
 
     let (consumed, done) = body


arnetheduck:
in streams we keep re-raising the same error - this is probably what should happen here too, ie reraise the stored error instead of clearing it .. for writes in particular, it's important because you might queue multiple writes and then an early write fails - for reads, we don't support queueing but it's easier to keep to coherent across read/write and asyncstream/streamtransport

etan-status:
Have changed it to retain the error / not clear it

chronos/transports/stream.nim
```diff
@@ -2545,7 +2545,7 @@ template readLoop(name, body: untyped): untyped =
     if len(transp.buffer) == 0:
       # We going to raise an error, only if transport buffer is empty.
       if ReadError in transp.state:
-        raise transp.getError()
```
arnetheduck, line 2548: let's get rid of getError entirely, it's an internal helper

arnetheduck: we can probably clear the error explicitly on close as well, to dealloc earlier

etan-status: Inlined getError (replaced with `move` which seems to do the same?)

not sure about close, closing seems to incl a second state over the existing one, so it's not guaranteed that the error will never be read again.

arnetheduck: I was under that impression too with move, but it's undefined - or rather, it's under consideraion - right now it resets the source but there's an idea that it shouldn't - hm :/

etan-status: so, should I try explicitly or keep it move?

arnetheduck: I've been toying with the idea to add `stew/take` (this is what the op that has a well-defined default state for the source is called in rust) as a template for move so we mark these locations explicitly, where  we depend on the source being cleared and defaulted rather than just "destructor disarmed" which is the guaranttee for "wasMoved" which drives move.

arnetheduck: maybe add a `take` to chronos for now somewhere in its utils - it's the most impacted project so far that does these tricks

etan-status: chronos depends on stew, so could also be part of stew. Have added a simple wrapper that asserts that the delete-assumption holds. is this how you intended?

chronos/take.nim
```diff
@@ -0,0 +1,17 @@
...
+#                MIT license (LICENSE-MIT)
+
+{.push raises: [].}
+
+template take*[T](x: var T): T =
+  block:
+    let res = move(x)
+    when T is ref:
+      doAssert x == nil
+    res
```
arnetheduck:
```suggestion
  when not defined(release):
    block:
      let res = move(x)
      doAssert x == default(typeof(x))
      res
    else:
      move(x)
```

we don't want this in release builds for now - it's just a way to notice that the language hasn't rugpulled this fine detail

arnetheduck: this approach is also a bit weird in that it causes a read on x _after_ the move which disables the optimistic move that move is supposed to do, for local variables - so it's not .. great.. actually, we should use `ensureMove` in there

etan-status: what does ensureMove do that move doesn't? why is there an API to check whether the compiler itself is correct?

arnetheduck: er, now that I think of it, it wouldn't work: for local variables, it ensures that it's the last read of x - move is "optimistic" for local vars - it moves iff it's the last access and otherwise doesn't .. it's a pretty weird function overall.

etan-status: so, should we just explicitly `reset()`, instead of the assert?

arnetheduck: ok, let's do that for now - and document that move might not leave the source in a well-defined (readable) state.

etan-status: or, capture a pointer to the original, cast it to somethign else to break the "optimistic" analysis, and then access into the pointer to verify? to prevent raw "x" access after the move

arnetheduck: any decent analyzer should notice the escaped pointer and complain or do something :/

arnetheduck: `move` + `reset` is fine for now - I'm not even sure what the solution _should_ be .. this is generally the destructive (rust) vs non-destructive (c++) move problem and I haven't thought about it too deeply just yet, beyond noting that nim lacks infrastructure to deal with it well afaict (ie when writing a collection).

here, it's just a simple ref - ostensibly, the reset could actually be removed by the compiler automatically as long as the move does the resetting - I don't know if it's smart enough to do that, but that's one idea, ie make it an optimizer thing to remove any redundant resets if they can be proven to be unnecessary..

etan-status: ok, applied move+reset

PR #728, opened 2026-09-02
Adds Win32 CI with test bug

jmgomez:


.github/workflows/windows_i386.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          nim-version: "2.2.x"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Restore MinGW-W64 (i686) from cache
+        id: mingw32-cache
+        uses: actions/cache@v4
+        with:
+          path: external/mingw32
+          key: mingw32-i686-posix-dwarf-11.1.0
```
arnetheduck: we should be able to use llvm-mingw, same as the other CI jobs - this is easier to keep track of and it also supports compiling to i386

PR #730, opened 2026-09-04
Fix doc examples.

moigagoo:
https://github.com/status-im/nim-chronos/pull/667#discussion_r3930827582


examples/http_client/chapter1/uptimemon.nimble
```diff
@@ -5,4 +5,5 @@ license       = "MIT"
 srcDir        = "src"
 bin           = @["uptimemon"]
 
-requires "nim#head", "chronos >= 4.2.2"
+requires "nim >= 2.0.0"
+requires "chronos"
```
nitely: > The intention was to provide examples as self-contained code that you can simply clone and run with nimble build.

but then you already cloned chronos, why install chronos again?

also, I don't think these nimble files work now that there is a nim.cfg pointing to the chronos root; nim should prefer the root over the chronos package (unless imported as pkg/chronos, but then we are back to square one).

moigagoo: Yeah, makes sense. I guess the whole idea of independent projects was kind of forced.

Replaced with ordinary files.

arnetheduck: I actually think it makes sense to keep them - the scenario of self-contained examples is something nimble should support - it's useful for building examples that have external deps, ie if the example needed json-ser it should be possible to express this need in a nimble file that does not infect chronos proper.

there's a related "workspace" feature which turns multiple nimble projects into a unified dependency tree - right now, it's centered around the `develop` workflow but this is another use case for it where probably, nimble should be creating a single workspace for nimble files nested in the same git repo - the idea is that within a workspace, there's a single "package repository" where all dependencies are shared - it sits somewhere between "local" and "global", ie if I clone nimbus-eth1 and eth2 into a workspace, I should get the union of their dependencies into a single `pkgs` / `vendor` folder.

cc @jmgomez

moigagoo: Although I agree that having the ability to ship self-contained example projects is nice in general, I still think in this particular case using those would be an overcomplication. Plain files are easier to navigation across and massively reduce code duplication.

If an example depended on an external package, I would argue the best way to handle this situation would be to:
1. Instruct the user to `nimble init` and add the package to .nimble file, then run `nimble run` (as we did before this PR).
2. Ship that dependency in `examples/vendor` so that the CI would be able to run the examples, while the user could simply `cd` into the folder and run `nimble setup -ly` and have the deps installed to `nimbledeps`.

So, conceptually, if we have the same code working as both the source for tutorials and as tests, we have to handle it differently depending on its role in this particular scenario. I don't think it's a bad thing, we just have to acknowledge that this code file actually has two jobs and has to where a different hat every time.

PR #732, opened 2026-09-05
Add ASAN/UBSAN to CI

etan-status:
Adopt AddressSanitizer / UndefinedBehaviorSanitizer to catch bugs early.

### DIFF
--- a/chronos/transports/stream.nim
+++ b/chronos/transports/stream.nim
@@ -29,6 +29,7 @@ type
     offset: uint                    # Writer vector offset
     size: int                       # Original size
     writer: Future[int]             # Writer vector completion Future
+    gcholder: ref seq[byte]         # May be used to extend lifetime of `buf`
 
   TransportKind* {.pure.} = enum
     Socket,                         # Socket transport
@@ -2449,16 +2450,11 @@ proc write*(transp: StreamTransport, msg: string,
 
   fastWrite(transp, pbytes, rbytes, nbytes)
 
-  let
-    written = nbytes - rbytes # In case fastWrite wrote some
-
-  var localCopy = msg
-  retFuture.addCallback(proc(_: pointer) = reset(localCopy))
-
-  pbytes = cast[ptr byte](addr localCopy[written])
-
+  let gcholder = new(seq[byte])
+  gcholder[] = @(pbytes.makeOpenArray(rbytes))
   var vector = StreamVector(kind: DataBuffer, writer: retFuture,
-                            buf: pbytes, buflen: rbytes, size: nbytes)
+                            buf: baseAddr gcholder[], buflen: rbytes,
+                            size: nbytes, gcholder: gcholder)
   transp.queue.addLast(vector)
   transp.resumeWrite()
   return retFuture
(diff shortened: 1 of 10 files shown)


chronos/transports/stream.nim
```diff
@@ -29,6 +29,7 @@ type
     offset: uint                    # Writer vector offset
     size: int                       # Original size
     writer: Future[int]             # Writer vector completion Future
+    gcholder: ref seq[byte]         # May be used to extend lifetime of `buf`
```
arnetheduck, line 32: the copy is not acceptable, so on cancellation there's a few options:
* return 0 from the write (instead of propagating cancellation)
* just drop the data 
   - we're in undefined-behavior land already, ie a stream where you cancel a few bytes mid-stream and then keep writing doesn't make sense
  - we're not offering "pseudo-messages-in-the-stream" - ie one write != one "message"
  - if someone cancels a write, it's _likely_ the close the stream
* drop the data _and_ close the socket - this is somewhat attractive insofar as it ensures that the other end sees a closed connection instead of a garbled byte stream - what I don't like about this is that it's somewhat the wrong layer to decide - but it's not _entirely_ wrong either

etan-status: Copy is (mostly) pre-existing, `gcholder` just retains it until it's no longer accessed, while the `master` branch also copies but deletes it on cancellation even when it's still being accessed.

- "return 0 from the write" doesn't fix the copy?
- "we're in undefined-behavior land already", this particular one is use-after-free, not UB, it's completely deterministic.
- "cancel a few bytes mid-stream", that's the point, current `master` doesn't allow cancelling mid-stream, it always finishes current chunk of message (except when streaming from file in which case the FD is closed prematurely).
  - problem here is we don't know downstream impact. nim-websock / nim-json-rpc / nim-libp2p, will they gracefully handle this if we just drop data in the middle, if they currently make sure to send each "message" in a single call (do they?)?
  - "we're not offering" -> on `master` we do, though. are we sure noone assumes that behaviour?
  - "likely the close the stream" -> cancellation on high level protocol can also mean that they just stop caring about the response (but in that case the protocol should catch cancellation on a higher level)

given we already copy on `master`, sure we want to integrate changes to cancellation behaviour into this fix pr, given uncertainty about downstream impact?

etan-status: Regarding "drop the data", it's at least better than today's solution of streaming data from a freed pointer, at least 😅 the bugs most likely lie in outer wrappers exposing the raw "write" future rather than a wrapper that handles cancellation... so when outer client cancels because they don't care about response anymore, it triggers the buggy situation.

etan-status: Should we even offer cancellation? or should the user just be required to close the socket themselves?

etan-status: Like, if nim-websock / nim-json-rpc / nim-libp2p etc return raw `write()` futures from chronos, that's likely wrong already (the cancellation should be handled by the higher library, in a wrapper around the write() future). it also depends on context what cancel means. on json-rpc it might just mean to disregard any responses as the req is no longer relevant. like, depending on use case you may still want to send the full message intact, while in other it should be handled with socket close.

not sure if "send full message intact" ever worked though. given that cancellation effectively drops the buffer today, it means that freed memory got streamed instead of the "intact" remainder of the message, so every client building on that assumption is broken already today.

anyway, many options that work technically, just need to decide on one.

etan-status: Best approach as part of the fix seems to just clean up the queue on cancellation, and fix the downstream libraries if they incorrectly make the assumption that cancellation still finishes writing the current chunk (that assumption is broken today, as the data gets replaced with data from the freed pointer). @arnetheduck that would be your option (2), I'd say strictly better than status quo today.

arnetheduck: > Copy is (mostly) pre-existing

the same problem hits the pointer-based pbytes overload - the other overloads semantically are just wrappers over that interface (the reason they're not implemented that way is to avoid the copy if the send is immediate) - this means that anyone using the lifetime of the write future as a "lifetime indicator" for how long they need to keep the pointer alive will hit the same bug.

You're right that we perform a copy in these wrappers in particular if needed, but the problem is general (also in `writeFile` I presume) and affects outside users trying to avoid the copy in a similar way.

> Best approach

yes, I agree - also, let's put it in a separate branch.

> clean up the queue on cancellation

I would probably use an `.finished` check in the writer loop to skip any writes whose future is finished - ie let the writer loop itself do the cleanup so we only have one "consumer" of the queue.

Issue #601, opened 2025-11-21
Move values out of `Future` on `await`


arnetheduck:
Consider the following code:
```nim
proc f(): Future[string] {.async.} =
  repeat('c', 10)

proc g(): Future[string] {.async.} =
  await f()

let x = waitFor g()
```

What practically happens is:

* in the body of `f`, a string is allocated and assigned to the returned `Future`
* `await f` makes a copy of that string and assigns it to the future returned by `g`
* `waitFor g()` makes another copy of the string and assigns it to `x`

In both cases, the value inside the returned `Future` is copied because `await` cannot be sure that it owns the future - that the Future is not used elsewhere.

Lacking knowledge unique ownership at compile time, we have a few options to address this:

* return a `Future[ref string]` instead - causes an additional allocation for the `ref string`
* make `Future.value` return a `var T` instead of `lent T` - this allows `let x = move(waitFor g())`, ie the caller can explicitly choose to move the value out of the future 
  * similarly, we can call the mutable version `mvalue` and have `await` / `waitFor` use the mutable version 
  * ugly because caller has to remember to move, even though 99% of usages don't need it
* make `await`/`waitFor` inject the move automatically
  * If the future has multiple references, this will break existing code that expects the `value` to stay in the future - for example:
  ```nim
  let f = g()
  let a = await f
  let b = await f
  ```
* add `mawait` or similarly named alternative that performs the move, leaving the existing `await` as-is
  * difficult to teach, reasonable only for legacy compat

PR #734, opened 2026-09-07
introduce `mawait`,`mreadFinished` et al

arnetheduck:
When a future is finished and has a single owner, it would be valuable to move the value out of the future and into the caller as outlined in https://github.com/status-im/nim-chronos/issues/601.

Future is a `ref` and single ownership is not tracked meaning that we cannot reliably detect single-ownership at compile time (yet?).

However, by adopting the nim convention of `m` returning a mutable value (such as `mitems` etc), we can give callers a mutable reference / view of the value in the future instance and thus allow them to manually move it out when they deem it safe to do so.

An added bonus is that we can now work with futures of move-only types, using the `move mawait` idiom - manually move the mutable value reference that `mawait` resolves to.

In order for this machinery to be more practical, we also make `readFinished` public - this helper is useful for reading futures that are guaranteed to be finished already wihtout causing an additional `FuturePendingError` exception effect.

Of course, using `readFinished` on a pending future is not valid and carries a runtime risk of panics.


chronos/internal/asyncfutures.nim
```diff
@@ -561,24 +561,84 @@ macro internalRaiseIfError*(fut: InternalRaisesFuture, raises, info: typed) =
...
+  ## This function is suitable for use together with combinators like
+  ## `allFutures` or similar when it is certain that the future is finished.
+  ##
+  ## The mutable view can be used to move the value out of the future.
+  ##
+  ## If the future has multiple await consumers or callbacks, the order in which
+  ## they get to observe the value is not defined - mutating the value is only
+  ## safe for single-owner futures or when only one of the observers accesses
+  ## the value.
+  doAssert fut.finished(), "Calling `readFinished` on unfinished future"
```
moigagoo, line 583: This should probably be mreadFinished.

chronos/internal/asyncfutures.nim
```diff
@@ -561,24 +561,84 @@ macro internalRaiseIfError*(fut: InternalRaisesFuture, raises, info: typed) =
...
+  ##
+  ## If the future failed or was cancelled, the corresponding exception will be
+  ## raised.
+  ##
+  ## If the future is still pending, the behavior is undefined and may result
+  ## in a runtime Defect.
+  ##
+  ## This function is suitable for use together with combinators like
+  ## `allFutures` or similar when it is certain that the future is finished.
+  fut.mreadFinished()
```
moigagoo, line 599: Why do you only do finished assertion in void-flavored procs? Is it because we're OK with re-raising the exception from mreadFinished?

arnetheduck: >  Is it because we're OK with re-raising the exception from mreadFinished?

yes - the only difference is the mutability of the return type (`lent` vs `var`). `void` can have neither `lent` nor `var` so it needs a special case

chronos/internal/asyncfutures.nim
```diff
@@ -561,24 +561,84 @@ macro internalRaiseIfError*(fut: InternalRaisesFuture, raises, info: typed) =
...
   ##
   ## If the future is still pending, `FuturePendingError` will be raised.
+  ##
+  ## The mutable view can be used to move the value out of the future.
+  ##
+  ## If the future has multiple await consumers or callbacks, the order in which
+  ## they get to observe the value is undefined - mutating the value is only
+  ## safe for single-owner futures or when only one of the observers accesses
+  ## the value.
   if not fut.finished():
```
moigagoo, line 629: In other places we do the finished check using assertion, i.e AssertionDefect is raised. Maybe it would make sense to raise AssertionDefect universally (or FuturePendingError universally)?

moigagoo (approved):

chronos/internal/asyncfutures.nim
moigagoo: Thanks!

chronos/internal/asyncfutures.nim
arnetheduck: the idea is to introduce this change without breaking existing things that might depend on being able to catch FuturePendingError

chronos/internal/asyncfutures.nim
```diff
@@ -495,7 +495,7 @@ macro internalRaiseIfError*(fut: FutureBase, info: typed) =
   # This is a macro so we can capture the line info from the original call and
   # report the correct line number on exception effect violation
   let
-    info = info.lineInfoObj()
+    info = fut.lineInfoObj()
```
arnetheduck:
```suggestion
    info = info.lineInfoObj()
```

PR #738, opened 2026-09-17
Make `await` return `var T`

arnetheduck:
When a future is finished and has a single owner, it would be valuable to move the value out of the future and into the caller as outlined in https://github.com/status-im/nim-chronos/issues/601.

Future is a `ref` and single ownership is not tracked meaning that we cannot reliably detect single-ownership reliably at compile time (yet?).

An added bonus is that we can now work with futures of move-only types, using the `move mawait` idiom - manually move the mutable value reference that `mawait` resolves to.

In order for this machinery to be more practical, we also make `readFinished` public - this helper is useful for reading futures that are guaranteed to be finished already wihtout causing an additional `FuturePendingError` exception effect.

Of course, using `readFinished` on a pending future is not valid and carries a runtime risk of panics.

This PR is a less explicit version of https://github.com/status-im/nim-chronos/pull/734 which instead of instroducing a new `mawait` name reuses `await` at the expense of making harder to spot mutable access to the future since any `await` now can perform mutations - in particular, when calling an overloaded function with the result of an await (`f(await g())`), the `var` overload will now be preferred and the value inside the future might accidentally change.


arnetheduck:
https://github.com/status-im/nim-chronos/compare/mawait..varawait

PR #740, opened 2026-09-17
Add ThreadSignalPtr.unregister

Ivansete-status:
- New `ThreadSignalPtr.unregister` proc in `chronos/threadsync.nim`.
- New `hasThreadDispatcher` proc in `chronos/internal/asyncengine.nim`.

## Why

On posix, `wait` registers the signal's fd with the waiting thread's dispatcher and leaves it registered, while `close` only unregisters it from the dispatcher of the thread calling it. A thread that awaits a signal it does not own, then calls `closeThreadDispatcher` (#614), hits:

```
closeDispatcher(): the selector still has descriptors registered - all streams must have been closed before closing [AssertionDefect]
```

This is the case in `nim-ffi`, whose context slots keep their signals across thread pairs, so a worker thread can neither close the signal nor unregister it.

`unregister`:
- removes the signal from the current thread's dispatcher, without closing it.
- returns `ok()` when the thread has no dispatcher, without creating one (hence `hasThreadDispatcher`), or when the signal is not registered.
- refuses while a `wait` on the signal is pending on this thread, as dropping the fd would leave that future never completing.
- does nothing on Windows, where `waitForSingleObject` registers and unregisters per wait.

An explicit call keeps `wait` free of an extra register/unregister pair of syscalls on every wait.

## Tests

New tests in `tests/testthreadsync.nim`: `Defect` without / clean close with `unregister`, no dispatcher created, no-op for a signal never waited on, refusal while a wait is pending.

`testthreadsync` and `testdispatcher` pass under `orc` and `refc` on macOS (kqueue, poll) and Linux (epoll).

## Related
- https://github.com/logos-messaging/nim-ffi/issues/164

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arnetheduck:
cancel the wait?

Ivansete-status:
> cancel the wait?

Tx for the swift comment but, what do you mean?

Apparently, cancelling the wait doesn't unregister the signal's fd. Besides, there's no direct mechanism to cancel the wait from within the `closeThreadDispatcher()`.

PR #741, opened 2026-09-18
Fix `POSIX_SPAWN` flags & avoid inheriting `SIG_IGN` in spawned processes 

nitely:
Changes:

- Fix `POSIX_SPAWN` flags values
- Add regression tests which fails without the fix
- Spawned processes stop inheriting SIG_IGN from their parent

Last change required because [SIGPIPE is ignored](https://github.com/status-im/nim-chronos/blob/e92ebfa30a9acd132158988cf5103eb6217b3b73/chronos/internal/asyncengine.nim#L912) at startup, and also [ioselectors_kqueue ignores each watched signal on macos](https://github.com/status-im/nim-chronos/blob/e92ebfa30a9acd132158988cf5103eb6217b3b73/chronos/ioselects/ioselectors_kqueue.nim#L328) (add/waitSignal). Without it the children won't receive kill signals for example. Ideally only SIGPIPE would need restoring (that fixes linux).

At least libuv seems to do something similar https://github.com/libuv/libuv/blob/84af0b18c5aee743a9d4182fa9c18a18fd9825e3/src/unix/process.c#L328 .


chronos/osdefs.nim
```diff
@@ -1269,21 +1269,27 @@ elif defined(haiku):
...
      defined(haiku):
 
-  const
-    POSIX_SPAWN_RESETIDS* = 0x01
-    POSIX_SPAWN_SETPGROUP* = 0x02
-    POSIX_SPAWN_SETSCHEDPARAM* = 0x04
-    POSIX_SPAWN_SETSCHEDULER* = 0x08
-    POSIX_SPAWN_SETSIGDEF* = 0x10
-    POSIX_SPAWN_SETSIGMASK* = 0x20
+  let
```
arnetheduck: hm, better have them as real constants and certainly so for the platforms where their values have been investigated

